PR #10, opened 2019-10-15
In cluster inspection to detect docker base image. Closes #6

### LINKED ISSUE #6: Pipeline steps base image
Currently Kale uses the KFP default base image when executing pipeline steps.

The alternative could be to use the NotebookServer image.

A user running a Notebook and testing some code, is working on the env defined by the Notebook Server image. If we snapshot and mount the workspace to preserve the user environment, then it is reasonable to run the steps on the same image where the user develops the notebook.

It is not enough to use the default Kale Notebook Server image, as users might extend this image with custom images that add some custom packages.

The approach explained in #3 could be used to retrieve the image of the current container when Kale is executed in the NotebookServer.

StefanoFioravanzo:
- Detect base docker image in case Kale is running in a Kubeflow Notebook server
- Parse `kubectl get pods` output
- Apply current container docker image to every pipeline step

### DIFF
--- a/kale/core.py
+++ b/kale/core.py
@@ -1,9 +1,7 @@
 import os
 import re
-import json
 import pprint
 import tempfile
-import subprocess
 
 import logging
 import logging.handlers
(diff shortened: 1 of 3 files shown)


StefanoFioravanzo:
@elikatsis Do you agree with the approach here to detect the docker image?

PR #48, opened 2020-01-28
Feature nbparser refactoring

StefanoFioravanzo:
- General refactoring and improvement of `nbparser` module
- Unit tests for `nbparser` module


kale/nbparser/parser.py
```diff
@@ -124,62 +116,88 @@ def parse_notebook(notebook):
...
-        for block in nb_graph:
-            block_data = nb_graph.nodes(data=True)[block]
-            nx.set_node_attributes(nb_graph, {block: {
-                'source': global_block + '\n\n' + block_data['source']}})
+            # add node to DAG, adding tags and source code of notebook cell
+            if step_name not in nb_graph.nodes:
+                nb_graph.add_node(step_name, source=[c.source],
+                                  ins=set(), outs=set())
+                for _prev_step_name in tags['prev_steps']:
+                    nb_graph.add_edge(_prev_step_name, step_name)
```
elikatsis: I'd go for `prev_step` instead, because this makes me think of `prev_step_name`

kale/nbparser/parser.py
```diff
@@ -124,62 +116,88 @@ def parse_notebook(notebook):
...
+            else:
+                merge_code(nb_graph, step_name, c.source)
+
+        prev_step_name = step_name
```
elikatsis: This must be inside the `else` statement. If it gets the `None` value then it won't merge the code. Two consecutive untagged blocks should be merged

**We definitely need a test for that!** Could you take a note for that?

```suggestion
            prev_step_name = step_name
```

kale/nbparser/parser.py
```diff
@@ -124,62 +116,88 @@ def parse_notebook(notebook):
...
+
+    # Prepend any `imports` and `functions` cells to every Pipeline step
+    for step in nb_graph:
+        step_source = nb_graph.nodes(data=True)[step]['source']
+        nx.set_node_attributes(
+            nb_graph,
+            {step: {
+                'source': imports_block + functions_block + step_source}})
```
elikatsis: I would go for either
```python
        nx.set_node_attributes(nb_graph, {step: {'source': (imports_block +
                                                            functions_block +
                                                            step_source)}})
```
or
```python
        source = imports_block + functions_block + step_source
        nx.set_node_attributes(nb_graph, step: {'source': source}})
```

kale/nbparser/parser.py
```diff
@@ -124,62 +116,88 @@ def parse_notebook(notebook):
...
+    # merge together pipeline parameters
+    pipeline_parameters = '\n'.join(pipeline_parameters)
+
+    # make the nodes' code a single multiline string
+    # NOTICE: this is temporary, waiting for the artifacts-viz-feature
+    for step in nb_graph:
+        step_source = nb_graph.nodes(data=True)[step]['source']
+        nx.set_node_attributes(
+            nb_graph,
+            {step: {'source': '\n'.join(step_source)}})
```
elikatsis: I'd prefer
```python
        step_dict = {step: {'source': '\n'.join(step_source)}}  # or any other variable name
        nx.set_node_attributes(nb_graph, step_dict)
```
Otherwise, the following is also fine
```python
        nx.set_node_attributes(nb_graph,                                         
                               {step: {'source': '\n'.join(step_source)}})
```

kale/nbparser/parser.py
```diff
@@ -23,63 +23,63 @@ def _copy_tags(tags):
...
+        # `imports` and `functions` are special tags that make a code cell
+        # belong to every Pipeline step Specifically:
+        #  - imports: the code of the corresponding cell(s) will be prepended
+        #             to every Pipeline step
+        #  - functions: same as imports, but the corresponding code is placed
+        #               **after** `imports`
         if t in ['imports', 'functions']:
-            parsed_tags['block_names'] = ['global']
+            parsed_tags['step_names'] = [t]
             return parsed_tags
```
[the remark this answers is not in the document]
StefanoFioravanzo: While refactoring I was conflicted about this too. Eventually I kept the conditions separate because I though it was more readable. But in fact merging them together enforces the fact that they are treated equally.

kale/nbparser/parser.py
```diff
@@ -3,29 +3,14 @@
...
-                 # as if it was empty.
-                 r'^block:([_a-z]([_a-z0-9]*)?)?$',
-                 r'^prev:[_a-z]([_a-z0-9]*)?$']
+_TAGS_LANGUAGE = [r'^imports$',
+                  r'^functions$',
+                  r'^pipeline-parameters$',
+                  r'^skip$',
+                  # Extension may end up with 'block:' as a tag.  We handle
+                  # that as if it was empty.
+                  r'^block:([_a-z]([_a-z0-9]*)?)?$',
```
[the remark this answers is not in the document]
StefanoFioravanzo: This is a **great** idea :)

kale/nbparser/parser.py
```diff
@@ -124,62 +116,88 @@ def parse_notebook(notebook):
...
-                merge_code(nb_graph, current_block, c.source)
+        # if none of the above apply, then we are parsing a code cell with
+        # a block names and (possibly) some dependencies
+
+        # check existence of prev steps. They must already exist in the graph
+        for prev in tags['prev_steps']:
+            if prev not in nb_graph.nodes:
+                raise ValueError("Block %s does not exist. "
+                                 "It was defined as previous block of %s"
+                                 % (prev, tags['step_names']))
```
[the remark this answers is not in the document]
StefanoFioravanzo: Correct. I have added a check in `parse_metadata` to verify that a cell does not contain `prev` tags while not having a `block` tag. See https://github.com/kubeflow-kale/kale/pull/48/commits/7d9682a44ee43eec508c906516b1d2bfb0ef6a5d

kale/nbparser/parser.py
StefanoFioravanzo: That was a great spot, and my bad for introducing this. It must have slipped while refactoring commits. Fixed :)

kale/nbparser/parser.py
```diff
@@ -65,6 +65,10 @@ def parse_metadata(metadata):
         # name(s) of the father Pipeline step(s)
         if tag_name == "prev":
             parsed_tags['prev_steps'].append(value)
+
+    if not parsed_tags['step_names'] and parsed_tags['step_names']:
```
elikatsis: It feels like something is wrong here

kale/nbparser/parser.py
```diff
@@ -57,10 +61,11 @@ def parse_metadata(metadata):
             parsed_tags['step_names'] = [t]
             return parsed_tags
 
-        # now only `block` and `prev` tags remain to be parsed.
+        # now only `block|step` and `prev` tags remain to be parsed.
         tag_name, value = t.split(':')
         # name of the future Pipeline step
-        if tag_name == "block" and value:
+        # TODO: Deprecate `block` in future release
+        if tag_name in ["block", "step"] and value:
```
elikatsis: The deprecation warning I'm referring to in my main comment could be inside this `if` codeblock.
I don't know if some deprecation library could work, but this should be sufficient:
```python
if tag_name == "block":
    log.warning(...)
```

kale/nbparser/parser.py
StefanoFioravanzo: Indeed!

kale/nbparser/parser.py
```diff
@@ -66,14 +67,18 @@ def parse_metadata(metadata):
         # name of the future Pipeline step
         # TODO: Deprecate `block` in future release
         if tag_name in ["block", "step"] and value:
+            if tag_name == "block":
+                warnings.warn("`block` tag will be deprecated in a future "
+                              "version, use `step` tag instead",
+                              DeprecationWarning)
```
StefanoFioravanzo: @elikatsis  I am phrasing this generically as `future version` and not specifically like `v.0.5` because we don't really have a proper release process or feature roadmap. I think it's fine to leave it like this, wdyt?

elikatsis: Yes, I totally agree.

nit (based on styling guidelines):
```python
warnings.warn("`block` tag will be deprecated in a future"
              " version, use `step` tag instead",
              DeprecationWarning)
```
We tend to put the space character to the start of the following line.

PR #55, opened 2020-02-03
Notebook metadata validation

StefanoFioravanzo:
- Move metadata validation from `core` to `utils` module.
- Move some volumes metadata validation from `codegen` to `utils`.
- Add base default metadata dict to standardise fields and values
- Add tests for metadata

### DIFF
--- a/kale/tests/unit_tests/test_metadata_utils.py
+++ b/kale/tests/unit_tests/test_metadata_utils.py
@@ -40,7 +40,7 @@ from kale.utils import metadata_utils
 ])
 def test_validate_volumes_metadata(volumes, target):
     """Tests volumes validation method."""
-    assert target == metadata_utils._validate_volumes_metadata(volumes)
+    assert target == metadata_utils._parse_volumes_metadata(volumes)
 
 
 @pytest.mark.parametrize("volumes,match", [
@@ -69,7 +69,7 @@ def test_validate_volumes_metadata(volumes, target):
 def test_convert_volume_annotations_exc(volumes, match):
     """Tests that volume spec errors are caught."""
     with pytest.raises(ValueError, match=match):
-        metadata_utils._validate_volumes_metadata(volumes)
+        metadata_utils._parse_volumes_metadata(volumes)
 
 
 @pytest.mark.parametrize("metadata,target", [
@@ -86,13 +86,13 @@ def test_validate_metadata(random_string, metadata, target):
     # metadata dict
     target.update({'pipeline_name': metadata['pipeline_name'] + '-rnd'})
     target.update({'experiment_name': metadata['experiment_name']})
-    assert target == metadata_utils.validate_metadata(metadata)
+    assert target == metadata_utils.parse_metadata(metadata)
 
 
 def test_validate_metadata_missing_required():
     """Tests that required metadata keys are checked for."""
     with pytest.raises(ValueError, match=r"Key experiment_name not found.*"):
-        metadata_utils.validate_metadata({})
+        metadata_utils.parse_metadata({})
 
     with pytest.raises(ValueError, match=r"Key pipeline_name not found.*"):
-        metadata_utils.validate_metadata({'experiment_name': 'test'})
+        metadata_utils.parse_metadata({'experiment_name': 'test'})
--- a/kale/tests/unit_tests/test_metadata_utils.py
+++ b/kale/tests/unit_tests/test_metadata_utils.py
@@ -1,3 +1,4 @@
+import copy
 import pytest
 
 from unittest import mock
@@ -73,13 +74,9 @@ def test_convert_volume_annotations_exc(volumes, match):
 
 @pytest.mark.parametrize("metadata,target", [
     ({'pipeline_name': 'test',
-      'experiment_name': 'test'},
-     metadata_utils.DEFAULT_METADATA),
-    # ---
-    ({'pipeline_name': 'test',
-      'experiment_name': 'test'},
-     metadata_utils.DEFAULT_METADATA),
-
+      'experiment_name': 'test',
+      'volumes': []},
+     copy.deepcopy(metadata_utils.DEFAULT_METADATA)),
 ])
 @mock.patch('kale.utils.metadata_utils.random_string')
 def test_validate_metadata(random_string, metadata, target):
--- a/kale/tests/unit_tests/test_metadata_utils.py
+++ b/kale/tests/unit_tests/test_metadata_utils.py
@@ -9,14 +9,14 @@ from kale.utils import metadata_utils
     ([], []),
     # ---
     ([{
-        'name': 'v1',
+        'name': 'v',
         'annotations': [{'key': 'a1', 'value': 'v1'}],
         'size': 5,
         'type': 'pv',
         'mount_point': '/'
     }],
      [{
-         'name': 'v1',
+         'name': 'v',
          'annotations': {'a1': 'v1'},
          'size': '5',
          'type': 'pv',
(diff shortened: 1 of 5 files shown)


kale/utils/metadata_utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+METADATA_REQUIRED_KEYS = [
+    'experiment_name',
+    'pipeline_name',
+]
+kale_step_name_regex = r'^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
+kale_name_msg = ("must consist of lower case alphanumeric characters "
+                 "or '-', and must start and end with an alphanumeric"
+                 " character.")
+k8s_valid_name_regex = r'^[\.\-a-z0-9]+$'
```
elikatsis: I think the correct regex is: `r'^[a-z][a-z0-9-.]*$'`
Plus, escaping `-` and `.` isn't required.

kale/utils/metadata_utils.py
```diff
@@ -0,0 +1,103 @@
...
+        if required not in metadata:
+            raise ValueError("Key %s not found. Add this field either on "
+                             "the notebook metadata or as an override" %
+                             required)
+
+    if not re.match(kale_step_name_regex, metadata['pipeline_name']):
+        raise ValueError("Pipeline name  %s" % kale_name_msg)
+
+    volumes = metadata.get('volumes', [])
+    if volumes or isinstance(volumes, list):
```
elikatsis: It should be `and` here, instead of `or`.

kale/utils/metadata_utils.py
```diff
@@ -49,21 +49,16 @@ def validate_metadata(notebook_metadata):
                              "the notebook metadata or as an override"
                              .format(required))
 
-    metadata = copy.deepcopy(DEFAULT_METADATA)
+    metadata = DEFAULT_METADATA.copy()
```
elikatsis: This really has to be a deepcopy, since `DEFAULT_METADATA` dictionary contains a list.

kale/utils/metadata_utils.py
```diff
@@ -49,21 +49,16 @@ def validate_metadata(notebook_metadata):
...
     metadata.update(notebook_metadata)
-    # check for required fields
```
elikatsis: How come you got this removed?

kale/tests/unit_tests/test_metadata_utils.py
```diff
@@ -37,23 +38,64 @@
...
+        'annotations': {'key': 'a1', 'value': 'v1'},
+        'size': 5,
+        'type': 'pv',
+        'mount_point': '/'
+    }], "Volume spec: annotations must be a list"),
 ])
-def test_convert_volume_annotations_exc(volumes):
-    """Tests that volumes are correctly converted from list into dict
+def test_convert_volume_annotations_exc(volumes, match):
+    """Tests that volumes are correctly converted from list into dict."""
```
elikatsis: Fix docstring:
```
Tests that volume spec errors are caught
```

kale/tests/unit_tests/test_metadata_utils.py
```diff
@@ -37,23 +38,64 @@
...
+    with pytest.raises(ValueError, match=match):
+        metadata_utils._validate_volumes_metadata(volumes)
+
+
+@pytest.mark.parametrize("metadata,target", [
+    ({'pipeline_name': 'test',
+      'experiment_name': 'test'},
+     metadata_utils.DEFAULT_METADATA),
```
elikatsis: I think we should `copy.deepcopy()` even that, because it should affect the next tests, depending on memory behavior.

Also, what's the purpose of two identical test parameters?

kale/utils/metadata_utils.py
```diff
@@ -53,10 +53,11 @@ def validate_metadata(notebook_metadata):
     metadata.update(notebook_metadata)
 
     if not re.match(kale_step_name_regex, metadata['pipeline_name']):
-        raise ValueError("Pipeline name  %s" % kale_name_msg)
+        raise ValueError("Pipeline name  {}".format(kale_name_msg))
```
elikatsis: This was Python3.5 compatible already :P 
But we can keep it for uniformity. (No need to comment on every occurrence)

kale/tests/unit_tests/test_metadata_utils.py
```diff
@@ -37,23 +38,64 @@
      }])
 ])
 def test_validate_volumes_metadata(volumes, target):
-    """Tests that volumes are correctly converted from list into dict
+    """Tests that volumes are correctly converted from list into dict."""
```
elikatsis: In this test, the conversion of volume **annotations** from list to dict is tested.
But the general purpose is to test the `_validate_volumes_metadata()` function. So I'd propose the doc string to be
```
Tests volumes validation method
```
or similar.

kale/utils/metadata_utils.py
elikatsis: Scratch that, I didn't see :p

kale/utils/metadata_utils.py
StefanoFioravanzo: Actually we should remove `volumes` and have just:
```python
if isinstance(volume, list):
```
Otherwise the condition would never pass with an empty list

kale/tests/unit_tests/test_metadata_utils.py
StefanoFioravanzo: Thanks for the `deepcopy` catch. As for the two identical test, this is an oversight of mine. This tests should have many more parameters, to test many more test cases. My intention was to set up the basic test and address the others in the future.

kale/utils/metadata_utils.py
StefanoFioravanzo: @elikatsis So we should also change the `kale_name_msg` in `must consist of lower case alphanumeric characters,  '-' or '.', and must start and end with a character`
Do you also confirm that the beginning of the regex is `[a-z]` and not `[a-z0-9]`?

elikatsis: Discussed live and came into this conclusion: `r'^[a-z][a-z0-9-.]*[a-z]$'`

kale/utils/metadata_utils.py
```diff
@@ -0,0 +1,103 @@
...
+]
+kale_step_name_regex = r'^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
+kale_name_msg = ("must consist of lower case alphanumeric characters "
+                 "or '-', and must start and end with an alphanumeric"
+                 " character.")
+k8s_valid_name_regex = r'^[\.\-a-z0-9]+$'
+k8s_name_msg = ("must consist of lower case alphanumeric characters, "
+                "'-' or '.'")
+volume_types = ['pv', 'pvc', 'new_pvc', 'clone']
+volume_required_fields = ['name', 'annotations', 'size', 'type', 'mount_point']
```
klolos: Module constants are typically named in all caps (https://www.python.org/dev/peps/pep-0008/#constants). Also, you may want to define these as tuples, which are immutable.

kale/utils/metadata_utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def validate_metadata(metadata):
```
klolos: Should this maybe be renamed to `parse_metadata`, since it modifies its argument? Validation functions are usually expected to perform only validation.

Also, you may want to avoid modifying the argument altogether by deepcopying the provided metadata object before performing any change. Performance is probably not an issue here, and its a bit safer since you can avoid aliasing bugs.

PR #56, opened 2020-02-10
Add e2e test from Titanic notebook to Python DSL

StefanoFioravanzo:
The test takes the existing Titanic notebook from the public `examples` repo. I chose to take the public one instead to make a copy into the test folder so that we always test the public version.

The test produces Python DSL in the same way we would from a Notebook RPC and then compares the output character-wise with a static file.


StefanoFioravanzo:
You should be able to run it with `pytest kale/tests/e2e`

PR #62, opened 2020-02-14
Run code inside an IPython kernel and capture artifacts

StefanoFioravanzo:
This PR introduces the following changes:

1. Step's code is handles as a list of code blocks, to keep information about the original notebook code cells
2. Utils to create an ipython kernel on the fly and run some code in it, while monitoring for stdout, stderr and error and redirecting the to the main thread.
3. Utils to parse the rich outputs produced by (2) and parse them to create a single html artifact for each execution
4. Extension to the code generation module to generate a pipeline that takes advantage of (1) (2) and (3)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,8 @@ setup(
         'jinja2 >=2.10, <3.0',
         'graphviz >=0.13, <1.0',
         'pyflakes >=2.1.1',
-        'dill >=0.3, <0.4'
+        'dill >=0.3, <0.4',
+        'IPython >= 7.6.0'
     ],
     extras_require={
         'dev': [
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
               'kale.rpc'
               ],
     install_requires=[
-        'kfp >=0.1.31, <0.2',
+        'kfp >= 0.2',
         'autopep8 >=1.4, <1.5',
         'nbformat >=4.4, <5.0',
         'networkx >=2.3, <3.0',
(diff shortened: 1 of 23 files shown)


StefanoFioravanzo:
@elikatsis I updated the PR with a better handling of the watcher thread. The relevant changes as

- The watcher thread is created with `daemon=True`, so that it will make the process hand when the main thread exists
- The main thread kills the ipython kernel when the code execution is finished, so that resources are disposed of gracefully (sockets etc..)

In this way we can call `kc.iopub_channel.get_msg` without a timeout and not get into trouble with long running codes.

setup.py
```diff
@@ -33,7 +33,11 @@
         'jinja2 >=2.10, <3.0',
         'graphviz >=0.13, <1.0',
         'pyflakes >=2.1.1',
-        'dill >=0.3, <0.4'
+        'dill >=0.3, <0.4',
+        'IPython >= 7.6.0',
+        'jupyter-client >= 5.3.4',
```
elikatsis: + jupyter-core >= 4.6.0
+ kfp-server-api == 0.1.18.3

elikatsis: We can also go for
+ kfp >= 0.2

elikatsis: + `packaging` is a pytest requirement, which we only install on `dev` installation, but you're using it on `jupyter_utils`, so you should also add an entry for that here.

elikatsis:
I also got this on final auto snapshot step:
```
This step is in Error state with this message: failed to save outputs: path /final_auto_snapshot.html does not exist (or /final_auto_snapshot.html is empty) in archive /argo/outputs/artifacts/final_auto_snapshot.tgz
```

This means that you are adding a `/final_auto_snapshot.html` output artifact, but you are not generating it.
So you should add an if statement to not add the following for the `final_auto_snapshot` step (in `pipeline_template.jinja2` file):
```
mlpipeline_ui_metadata = {'{{ name }}': '/{{ name }}.html'}
```

These will be simplified in the future, when we declare component outputs via function outputs (`->` notation).

PR #63, opened 2020-02-14
Add PyTest workflow

StefanoFioravanzo:
The workflow will run at every commit (to any branch or PR). We could restrict this to run only during PR or commits to mater, but since it is inexpensive on our part and the task is relatively quick we can leave like this.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,0 +1,23 @@
+name: PyTest
+
+on: [push]
+
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v1
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,23 @@
+name: PyTest
+
+on: [push]
+
+jobs:
+  build:
```
elikatsis: nit: Rename to `test` ?

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,23 @@
...
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.6
```
elikatsis: Should we do the same for all python versions? So that we get to know fast any incompatibilities

PR #69, opened 2020-02-27
Dockerfiles: Force-update unsatisfied kale requirements

elikatsis:
Use `pip install -U` in Dockerfiles when installing kale.
This force-updates kale requirements that are not satisfied.

Signed-off-by: Ilias Katsakioris <elikatsis@arrikto.com>

### DIFF
--- a/docker/jupyterlab/Dockerfile
+++ b/docker/jupyterlab/Dockerfile
@@ -21,7 +21,7 @@ RUN echo "set background=dark" >> /etc/vim/vimrc.local
 RUN pip3 install --upgrade pip && \
     pip3 install --upgrade "jupyterlab<2.0.0" && \
     pip3 install https://storage.googleapis.com/ml-pipeline/release/latest/kfp.tar.gz --upgrade && \
-    pip3 install kubeflow-kale && \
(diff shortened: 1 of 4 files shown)


StefanoFioravanzo:
Have you tried this and does it work for you? I did some investigation and found this: https://stackoverflow.com/questions/55804822/pip-install-latest-dependency-versions

Pip also has the `--upgrade-strategy` option, and it looks like that in our case we should use `--upgrade --upgrade-strategy eager`

elikatsis:
It did work for me. As I can see in the [documentation](https://pip.pypa.io/en/stable/reference/pip_install/#argument-handling)

```
--upgrade-strategy <upgrade_strategy>:
Determines how dependency upgrading should be handled [default: only-if-needed]. “eager” - dependencies are upgraded regardless of whether the currently installed version satisfies the requirements of the upgraded package(s). “only-if-needed” - are upgraded only when they do not satisfy the requirements of the upgraded package(s).
```

If I understand correctly, the default behavior of `--upgrade/-U` (aka `--upgrade-strategy only-if-needed`) is to install different versions of the dependences, when these are not satisfied.

PR #70, opened 2020-03-03
Simplify marshal code in templates

StefanoFioravanzo:
This PR introduces a refactoring of the code that marshals data in and out of pipeline steps.

Instead of producing a lot code using templating, the marshalling code is mode to `marshal/utils.py`, thus making the generated code much simpler.

### DIFF
--- a/kale/marshal/utils.py
+++ b/kale/marshal/utils.py
@@ -1,0 +1,66 @@
+import os
+
+from kale.marshal import resource_save
+from kale.marshal import resource_load
+
+KALE_DATA_DIRECTORY = ""
+KALE_DIRECTORY_FILE_NAMES = []
+
+
+def set_kale_data_directory(path):
+    """Set the Kale data directory path."""
+    global KALE_DATA_DIRECTORY
+    KALE_DATA_DIRECTORY = path
+    # create dir if not exists
+    if not os.path.isdir(KALE_DATA_DIRECTORY):
+        os.makedirs(KALE_DATA_DIRECTORY, exist_ok=True)
+
+
+def set_kale_directory_file_names():
+    """Set the list of filenames present in the Kale data directory path."""
+    global KALE_DIRECTORY_FILE_NAMES
+    KALE_DIRECTORY_FILE_NAMES = [
+        os.path.splitext(f)[0]
+        for f in os.listdir(KALE_DATA_DIRECTORY)
+        if os.path.isfile(os.path.join(KALE_DATA_DIRECTORY, f))
+    ]
+
+
+def save(obj, obj_name):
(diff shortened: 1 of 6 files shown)


kale/marshal/utils.py
```diff
@@ -0,0 +1,84 @@
...
+    """Set the list of filenames present in the Kale data directory path."""
+    global KALE_DIRECTORY_FILE_NAMES
+    KALE_DIRECTORY_FILE_NAMES = [
+        os.path.splitext(f)[0]
+        for f in os.listdir(KALE_DATA_DIRECTORY)
+        if os.path.isfile(os.path.join(KALE_DATA_DIRECTORY, f))
+    ]
+
+
+def _get_ipykernel_locals():
```
StefanoFioravanzo: @elikatsis @klolos I am using this approach to make it seamless retrieving the `locals` (that is the same as `globals`) from the ipython context, that is calling either `save()` or `load()` and that, in turn, call `_get_ipykernel_locals`. I am not 100% sure this is the best solution, or that it could lead to issues. WDYT?

kale/marshal/utils.py
```diff
@@ -0,0 +1,84 @@
...
+    """Get the locals from the ipykernel context."""
+    global _IPYKERNEL_LOCALS
+    if not _IPYKERNEL_LOCALS:
+        frame = inspect.currentframe()
+        try:
+            _IPYKERNEL_LOCALS = frame.f_back.f_back.f_locals
+        finally:
+            # https://docs.python.org/3/library/inspect.html#the-interpreter-stack
+            del frame
+    return _IPYKERNEL_LOCALS
```
StefanoFioravanzo: @elikatsis @klolos I am using this approach to make it seamless retrieving the `locals` (that is the same as `globals`) from the ipython context, that is calling either `save()` or `load()` and that, in turn, call `_get_ipykernel_locals`. I am not 100% sure this is the best solution, or that it could lead to issues. WDYT?

klolos: The obvious issue is that this function would only work if it is called by a function exactly one frame below the kernel. If for example you have another helper calling `load()`, you would end up adding the variable to that function's context. I'm not sure its worth the trouble to do that. In particular:

During `save()` I don't understand why you need the locals in the first place. If the variable is not in `locals()`, won't `_kale_marshal_utils.save({{ out_var }}, "{{ out_var }}")` fail anyway, since the first argument is undefined? Also, isn't it a Kale bug to attempt to marshal an undefined variable?

During `load()` I think you could just do `{{ in_var }} = _kale_marshal_utils.load("{{ in_var }}")`, so you don't even need to use any `locals()` magic at all, right? Please tell me if I'm missing something.

PR #74, opened 2020-03-04
Pass appropriate pipeline parameters to pipeline steps

StefanoFioravanzo:
This PR restricts the number of pipeline parameters that are passed to the lightweight functions to those that are actually needed by each pipeline step.

### DIFF
--- a/kale/codegen/generate_code.py
+++ b/kale/codegen/generate_code.py
@@ -99,9 +99,10 @@ def get_args(parameters):
         - 'parameters_types': parameters types as list
         - 'parameters_values': parameters default values as list
     """
-    parameters_names = sorted(parameters.keys())
-    parameters_types = [arg[0] for arg in parameters.values()]
-    parameters_values = [arg[1] for arg in parameters.values()]
+    sorted_parameters = dict(sorted(parameters.items()))
+    parameters_names = list(sorted_parameters.keys())
+    parameters_types = [arg[0] for arg in sorted_parameters.values()]
+    parameters_values = [arg[1] for arg in sorted_parameters.values()]
 
     return {
         'parameters_names': parameters_names,
@@ -139,12 +140,14 @@ def generate_lightweight_component(template, step_name, step_data, nb_path,
     step_source = [re.sub(r"'''", "\\'\\'\\'", s) for s in step_source]
     step_marshal_in = step_data.get('ins', [])
     step_marshal_out = step_data.get('outs', [])
+    step_parameters = get_args(step_data.get('parameters', {}))
 
     fn_code = template.render(
         step_name=step_name,
         function_body=step_source,
         in_variables=step_marshal_in,
         out_variables=step_marshal_out,
+        parameters=step_parameters,
         nb_path=nb_path,
         # step_parameters overrides the parameters fields of metadata
         **{**metadata, **step_parameters}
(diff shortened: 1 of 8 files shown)


kale/static_analysis/dependencies.py
```diff
@@ -61,25 +61,37 @@ def pyflakes_report(code):
...
-        if ignore_symbols:
-            ins.difference_update(set(ignore_symbols))
-        nx.set_node_attributes(nb_graph, {block: {'ins': sorted(ins)}})
-
-
-def detect_out_dependencies(nb_graph: nx.DiGraph, ignore_symbols: set = None):
+        # Pipeline parameters will be part of the names that are missing,
+        # but of course we don't want to marshal them in as they will be
+        # present as parameters
+        relevant_parameters = []
```
elikatsis: This variable, `relevant_parameters`, is assigned with the result of the operation `ins.intersection(params)`, which is a `set`.

So, I suggest:
```suggestion
        relevant_parameters = set()
```

kale/static_analysis/dependencies.py
```diff
@@ -61,25 +61,37 @@ def pyflakes_report(code):
...
+        if pipeline_parameters:
+            # Not all pipeline parameters are needed in every pipeline step,
+            # these are the parameters that are actually needed by this step.
+            params = set(pipeline_parameters.keys())
+            relevant_parameters = ins.intersection(set(params))
```
elikatsis: `intersection()` method accepts any iterable, i.e., even lists.
So this could be simplified as:
```suggestion
            relevant_parameters = ins.intersection(pipeline_parameter.keys())
```

kale/codegen/generate_code.py
```diff
@@ -160,12 +164,18 @@ def generate_pipeline(template, nb_graph, step_names, lightweight_components,
     leaf_steps = [x for x in nb_graph.nodes()
                   if nb_graph.out_degree(x) == 0]
 
+    # create a dict with step names and their parameters
+    steps_parameters = {step: sorted(nb_graph.nodes(data=True)[step]
```
elikatsis: Rename suggestion: `all_step_parameters`, or something else if you have a preference.

I'm not fond of `step_parameters` and `steps_parameters`, because the difference between the names is pretty small and it's not clear what the difference is. Thus, it's a bug trap :sweat_smile:

kale/codegen/generate_code.py
```diff
@@ -99,9 +100,10 @@ def get_args(parameters):
         - 'parameters_types': parameters types as list
         - 'parameters_values': parameters default values as list
     """
-    parameters_names = sorted(parameters.keys())
-    parameters_types = [arg[0] for arg in parameters.values()]
-    parameters_values = [arg[1] for arg in parameters.values()]
+    sorted_parameters = collections.OrderedDict(sorted(parameters.items()))
+    parameters_names = sorted(sorted_parameters.keys())
+    parameters_types = [arg[0] for arg in sorted_parameters.values()]
+    parameters_values = [arg[1] for arg in sorted_parameters.values()]
```
elikatsis: nits:
```suggestion
    sorted_parameters = dict(sorted(parameters.items()))
    parameters_names = list(sorted_parameters.keys())
    parameters_types = [arg[0] for arg in sorted_parameters.values()]
    parameters_values = [arg[1] for arg in sorted_parameters.values()]
```

explanation:
1. Ordered dict isn't required, since it's for maintaining the insertion order (but we don't insert anything new)
   >Python OrderedDict is a dict subclass that maintains the items insertion order. When we iterate over an OrderedDict, items are returned in the order they were inserted. A regular dictionary doesn't track the insertion order.
2. keys are already sorted (because of `sorted(parameters.items())`)
3. In some other case, using `sorted()` when getting the keys and not when getting the values would result in errors. That is, because you wouldn't know if the order of keys would change when you sort them.
   Here it is not an issue, since `sorted_parameters` is already sorted by keys

After that change you can also remove the `collections` import statement.

PR #75, opened 2020-03-04
Escape characters with unicode_escape encode

StefanoFioravanzo:
We need to escape characters like `\n`, `\r` etc from the source code, otherwise they are parsed as escape characters by ipython and they screw up the executed source code.

This solution is just a proposal, open for discussion.


StefanoFioravanzo:
Changed the escaping to operate **line wise** instead of **block wise**, so that the source code maintains its indentation in the generated python script

StefanoFioravanzo:
@elikatsis I updated the comments to reflect the new changes. I think that this is a good enough solution for now for escaping the code, as it maintains the formatting.

PR #124, opened 2020-05-08
Integrate Kale with ML Metadata

elikatsis:


backend/kale/utils/mlmd_utils.py
```diff
@@ -499,6 +499,7 @@ def _create_rok_artifact_from_task(task):
 
     custom_properties = dict()
     for i in range(members):
+        # TODO: Change member_i to contain a JSON string
```
StefanoFioravanzo, line 502: Is this TODO something that you want to do in the future because of lack of time or some other reason?

PR #132, opened 2020-05-20
Katib progress: Show current optimal trial

elikatsis:
* Add dialog-body CSS class for notebook dialogs
* Introduce InfoBadge
* Return and show `currentOptimalTrial` field of Katib experiment's status

### DIFF
--- a/labextension/src/widgets/deploys-progress/DeployProgress.tsx
+++ b/labextension/src/widgets/deploys-progress/DeployProgress.tsx
@@ -94,30 +94,30 @@ interface DeployProgress extends DeployProgressState {
 
 export const DeployProgress: React.FunctionComponent<DeployProgress> = props => {
   const getWarningBadge = (title: string, content: any) => {
-    return content ? (
-      <a
-        onClick={_ => {
-          NotebookUtils.showMessage(title, content);
-        }}
-      >
-        <WarningIcon style={{ color: color.alert, height: 18, width: 18 }} />
-      </a>
-    ) : (
-      ''
+    return (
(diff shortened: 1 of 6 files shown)


labextension/src/widgets/deploys-progress/DeployProgress.tsx
```diff
@@ -105,6 +106,20 @@ export const DeployProgress: React.FunctionComponent<DeployProgress> = props =>
     );
   };
 
+  const getInfoBadge = (title: string, content: any) => {
+    return content ? (
```
StefanoFioravanzo, line 111: Let's convert this syntax to `<bool> && element`, so it's cleaner.
Take a look here https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator
I know that we basically never do this in the codebase, but I'd like to start introducing a cleaner syntax whenever possible :)

labextension/src/widgets/deploys-progress/DeployProgress.tsx
```diff
@@ -575,7 +588,13 @@ export const DeployProgress: React.FunctionComponent<DeployProgress> = props =>
             <div className="deploy-progress-label">
               Running Katib experiment...
             </div>
-            <div className="deploy-progress-value">{katibTpl}</div>
+            <div className="deploy-progress-value">
+              {katibTpl}
+              {getInfoBadge(
+                'Katib best result (so far)',
```
StefanoFioravanzo, line 594: What about "Katib current best result" ?

PR #170, opened 2020-07-28
Improve UX on progress of Katib experiment

elikatsis:


labextension/src/widgets/deploys-progress/KatibProgress.tsx
```diff
@@ -77,7 +109,11 @@ export const KatibProgress: React.FunctionComponent<IKatibProgressProps> = props
   const getText = (experiment: IKatibExperiment) => {
     switch (experiment.status) {
       case KatibExperimentStatus.FAILED:
-        return 'Failed';
+        if (isKatibGoalNotReached(experiment)) {
```
StefanoFioravanzo, line 112: Let's use the compact syntax here: `<condition> ? true : false`

labextension/src/widgets/deploys-progress/KatibProgress.tsx
```diff
@@ -86,32 +122,89 @@ export const KatibProgress: React.FunctionComponent<IKatibProgressProps> = props
   };
 
   const getComponent = (experiment: IKatibExperiment) => {
-    let IconComponent: any = UnknownIcon;
-    let iconColor = '#5f6368';
+    let tooltipSet = false;
+    let styles = { color: '#5f6368', height: 18, width: 18 };
+    let IconComponent = <UnknownIcon style={styles} />;
 
     switch (experiment.status) {
```
StefanoFioravanzo, line 129: It looks to me like the two cases `FAILED` and `SUCCEEDED` implement the same logic. Maybe we could move it a separate function?

Issue #189, opened 2020-08-28
Support Python 3.7 and 3.8 and kfp 1.0.0


ksonbol:
Hello,

I was only able to install kale with python 3.6.3 and it installed kfp 0.3.0 as a requirement which is quite outdated now.

With Python 3.7.7, there is this dependency conflict:
`kfp 0.3.0 requires kfp-server-api<0.4.0,>=0.2.5, but you'll have kfp-server-api 0.1.18.3 which is incompatible`.

With Python 3.8, it is not possible to install it because of the `ml-metadata` dependency which isn't upgraded to support python 3.8 yet.

StefanoFioravanzo:
Hu @ksonbol can you try installing from `master`? We have removed the limiting kfp dependency but haven't release a new version yet

DavidSpek:
ml-metadata v0.24.0 was released 2 days ago with support for python 3.8. 
https://github.com/google/ml-metadata/releases/tag/v0.24.0

DavidSpek:
@ksonbol Python 3.8 is now compatible, along with KFP 1.0.0. Can you confirm this issue is resolved for you?

ksonbol:
Thank you @DavidSpek . I can confirm that building from master works fine. Looking forward to the coming release :)

DavidSpek:
I believe this issue can be closed in that case.

Issue #157, opened 2020-07-10
Incompatible kfp-server-api version and python version requirements incorrect


DavidSpek:
While trying to incorporate kubeflow-kale into my notebook image I consistently get the error:
`ERROR: kfp 0.3.0 has requirement kfp-server-api<0.4.0,>=0.2.5, but you'll have kfp-server-api 0.1.18.3 which is incompatible.`
It seems there there is a discrepancy in the setup.py requirements. 

Also, the image I am using has python 3.8 installed by default, which should be supported as `python_requires='>=3.6.0'` has been specified. However, the requirement `'ml_metadata > 0.21, < 0.22'` creates an error with python 3.8 as ml_metadata is not available for python 3.8.

elikatsis:
Hello,

Regarding the package `kfp-server-api`, we have pinned version `0.1.18.3` because that way we workaround these issues:
* https://github.com/kubeflow/pipelines/issues/2764
* https://github.com/kubeflow/pipelines/issues/3686
We don't make use of any newer client feature, so using that version is ok for us (the `ERROR` is more like a `WARNING`).
In the near future, Kale will target more recent KFP deployment versions and we will be fixing this.

For the python version, indeed it seems like `ml-metadata` isn't available for python 3.8 (https://pypi.org/project/ml-metadata/). We should check all packages and fix the `setup.py` accordingly.

DavidSpek:
ml-metadata 0.24.0 was released 2 days ago and adds support for python 3.8 and fixes a grpcio bug that seems relevant. 
https://github.com/google/ml-metadata/releases/tag/v0.24.0

PR #211, opened 2020-09-18
Change to ml_metadata 0.24.0

DavidSpek:
Fixes issue #189 and #157


StefanoFioravanzo:
@DavidSpek can you remove the `gcrpcio` dependency as well, since the newest version of `ml-metadata` comes with it? Can you do that in the same commit? Also, can you remove the merge commit and just rebase the one single commit on top of master? Thanks!

DavidSpek:
@StefanoFioravanzo This should be good now.

StefanoFioravanzo (approved):

Issue #210, opened 2020-09-18
Kale Pipeline list experiments raised an unhandled exception "Namespace is empty"


davidspek:
After applying a fix to allow Kale to compile, upload and run pipelines in a multi-user environment according to the steps I posted in https://github.com/kubeflow-kale/kale/issues/204#issuecomment-694771168, I receive the following error that Kale can't properly list experiments as the namespace field is empty. Despite this error, Kale is able to upload the pipeline. However manually running that pipeline results in a HTTP error 403 during the `Getting workflow` step. 

kale.log showing the function 'list_experiments' raised an unhandled exception
```
2020-09-18 12:16:20 run:124 [[ERROR]] [TID=e9ow93ylkm] [/home/jovyan/mzquality.ipynb] RPC function 'list_experiments' raised an unhandled exception
Traceback (most recent call last):
  File "/opt/conda/lib/python3.8/site-packages/kale/rpc/run.py", line 116, in run
    result = func(request, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/kale/rpc/kfp.py", line 37, in list_experiments
    for e in c.list_experiments().experiments or []]
  File "/opt/conda/lib/python3.8/site-packages/kfp/_client.py", line 330, in list_experiments
    response = self._experiment_api.list_experiment(
  File "/opt/conda/lib/python3.8/site-packages/kfp_server_api/api/experiment_service_api.py", line 581, in list_experiment
    return self.list_experiment_with_http_info(**kwargs)  # noqa: E501
  File "/opt/conda/lib/python3.8/site-packages/kfp_server_api/api/experiment_service_api.py", line 682, in list_experiment_with_http_info
    return self.api_client.call_api(
  File "/opt/conda/lib/python3.8/site-packages/kfp_server_api/api_client.py", line 378, in call_api
    return self.__call_api(resource_path, method,
  File "/opt/conda/lib/python3.8/site-packages/kfp_server_api/api_client.py", line 202, in __call_api
    raise e
  File "/opt/conda/lib/python3.8/site-packages/kfp_server_api/api_client.py", line 195, in __call_api
    response_data = self.request(
  File "/opt/conda/lib/python3.8/site-packages/kfp_server_api/api_client.py", line 403, in request
    return self.rest_client.GET(url,
  File "/opt/conda/lib/python3.8/site-packages/kfp_server_api/rest.py", line 244, in GET
    return self.request("GET", url,
[...]

PR #213, opened 2020-09-18
Initial support for Kale in multi-user environment

DavidSpek:
Sets the k8s namespace that the Jupyter Notebook Server is running in for KFP to allow Kale to properly list existing experiments. Related to issue #210


DavidSpek:
@StefanoFioravanzo Just calling `get_user_namespace` when executing `kfp.list_experiments` doesn't seem to be enough to set the namespace for the notebook server. Do you have an idea for a better approach for this?

PR #217, opened 2020-09-26
[WIP] Add support for Rook CephFS snapshots

DavidSpek:
This pull request aims to add support for taking snapshots of notebook persistent volumes using Rook CephFS as Storage Class, similar to the functionality Kale already has with Rok. Due to the fact that the standard Kubernetes `VolumeSnapshot` API is used, this pull request could be extended to all CSI drivers that support snapshot functionality. Do note that Kale relies on `ReadWriteMany` support to be able to execute pipeline steps in parallel. There is a [method](https://github.com/kubeflow-kale/kale/blob/4cf2a1988bb19eb4904a1fa8908514ab7fbb539f/FAQ.md#pod-has-unbound-immediate-persistentvolumeclaim) to use `ReadWriteOnce` volmues; however, this requires manual editing of the created pipeline before uploading. 

Useful information:
[The API Objects `VolumeSnapshot`, `VolumeSnapshotContent` and `VolumeSnapshotClass` are CRDs and not part of the Core API. ](https://kubernetes.io/docs/concepts/storage/volume-snapshots/). As such the CustomObjectsAPI is used to create CephFS snapshots: https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#delete_namespaced_custom_object
To create a PVC from a CephFS snapshot the Core API is used. 

To allow a notebook to create Rook CephFS snapshot a ClusterRole and RoleBinding for the default-editor service account in the given namespace must be applied. Below are examples for use with the namespace "admin". Any tips, comments or recommendations to improve the security of the ClusterRole and RoleBinding and how it can best be deployed are greatly appreciated. 

**!!!WARNING!!!**
**Might not be secure, only use for testing!**

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: snapshot-access
rules:
  - apiGroups: ["snapshot.storage.k8s.io"]
    resources: ["volumesnapshots"]
    verbs: ["create", "get", "list", "watch", "patch", "delete"]
  - apiGroups: ["snapshot.storage.k8s.io"]
    resources: ["volumesnapshotcontents"]
    verbs: ["create", "get", "list", "watch", "update", "delete"]
  - apiGroups: ["snapshot.storage.k8s.io"]
    resources: ["volumesnapshotclasses"]
    verbs: ["get", "list", "watch"]
  - apiGroups: ["snapshot.storage.k8s.io"]
    resources: ["volumesnapshotcontents/status"]
    verbs: ["update"]
  - apiGroups: ["snapshot.storage.k8s.io"]
    resources: ["volumesnapshots/status"]
    verbs: ["update"]
```

```
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: allow-snapshot-nb-admin
  namespace: admin
subjects:
- kind: ServiceAccount
  name: default-editor
  namespace: admin
roleRef:
  kind: ClusterRole
  name: snapshot-access
  apiGroup: rbac.authorization.k8s.io
```


DavidSpek:
@StefanoFioravanzo @elikatsis Is it an idea to merge this pull request into a new branch so others can help with further development?

DavidSpek:
Update:

Add ability to snapshot all the PVCs atached to the pod. If using `snapshot_notebook` snapshot all PVCs connected to the notebook and add some notebook metadata as labels (pod name, container name, if the volume was the workspace directory) and annotations (mount path and docker image of the container) to the snapshot. A `list_pvc_snapshots` function is added that can be used to list snapshots filter using a label selector. For example: `list_pvc_snapshots(label_selector="pod=dev-0, is_workspace_dir=True")`. Eventually, the plan is to be able to start a new notebook server from a (set of) snapshot(s) using the metadata, similar to how Rok functions.

backend/kale/common/cephfsutils.py
```diff
@@ -0,0 +1,258 @@
...
+    group="snapshot.storage.k8s.io",
+    version="v1beta1",
+    namespace=namespace,
+    plural="volumesnapshots",
+    label_selector=label_selector)
+    return pvc_snapshots
+
+
+def hydrate_pvc_from_snapshot(new_pvc_name, source_snapshot_name):
+    """Create a new PVC out of a Rok snapshot."""
```
RDarrylR, line 195: Change Rok to Ceph

backend/kale/common/podutils.py
```diff
@@ -171,19 +174,19 @@ def _list_volumes(client, namespace, pod_name, container_name):
         #  result in an incomplete notebook snapshot.
         pvc = client.read_namespaced_persistent_volume_claim(pvc.claim_name,
                                                              namespace)
-        if pvc.spec.storage_class_name != ROK_CSI_STORAGE_CLASS:
-            msg = ("Found PVC with storage class '%s'. Only storage class '%s'"
-                   " is supported."
-                   % (pvc.spec.storage_class_name, ROK_CSI_STORAGE_CLASS))
+        if pvc.spec.storage_class_name != ROK_CSI_STORAGE_CLASS and pvc.spec.storage_class_name != ROOK_CEPHFS_CSI_STORAGE_CLASS:
```
RDarrylR, line 177: Likely worth making a list of supported storage classes rather than just adding a new case for each type. Hopefully we will have many types. With a list it can just check if the current storage class is in the list.

DavidSpek: Thanks for the reviews and tips. Removing this hardcoding is something I have been thinking about on the side for a bit already. A list is indeed an option but it is still hardcoded. A more elegant solution would be to check if the storage provider supports snapshots or to check if a snapshot class exists on the cluster. The issue this creates is knowing if RWO or RWX volumes are supported.

backend/kale/common/cephfsutils.py
DavidSpek: Thanks for catching that. I changed `Rok` to `volume`, to keep it more general. What does need some closer attention in this part is the `access_mode`, which I chose to used `ReadWriteMany` while the original Rok code uses `ReadWriteOnce`. Rok needs ReadWriteMany for certain things so I chose to use that here, though I am not sure if it is required. This also limits the supported CSI storage providers to those that support ReadWriteMany. I think the best solution is to create a function and check if the CSI supports ReadWriteMany (this could then also be used when generating the pipeline so people do not need to edit the python as stated [here](https://github.com/kubeflow-kale/kale/blob/master/FAQ.md#pod-has-unbound-immediate-persistentvolumeclaim).

DavidSpek:
Update:

I have just added the ability to restore a Notebook server from a "Notebook snapshot". In reality this is an individual snapshot of each PVC mounted to a Notebook server with some metadata written to each snapshot. You can call the function `restore_notebook` and pass it the snapshot name of one of the PVCs that was mounted to the notebook. The function will then discover other PVCs that were mounted, create new PVCs from the snapshots of those PVCs and finally start a new Notebook server with the new PVCs and the metadata from the snapshots (name, image, namespace).

backend/kale/common/podutils.py
DavidSpek: With the latest commit I have added functions that check if the provisioner of the PVC has an accompanying VolumeSnapshotClass rather than hardcoding what storage classes are supported. This also checks if a VolumeSnapshotClass actually exists, as you could be using CephFS or CephRDB without having added the VolumeSnapshotClass in which case it also isn't possible to take volume snapshots.

DavidSpek:
Update:
With the latest commit I have added functions that check if the provisioner of the PVC has an accompanying VolumeSnapshotClass rather than hardcoding what storage classes are supported. This also checks if a VolumeSnapshotClass actually exists, as you could be using CephFS, CephRDB or another storage provisioner without having added the VolumeSnapshotClass in which case it also isn't possible to take volume snapshots.

This is a first step to allow Kale to be storage provider agnostic when it comes to taking snapshots, as the other functions I wrote should universal.

For this functionality to work the following ClusterRole and ClusterRoleBinding are needed to be able to detect the VolumeSnapshotClasses and Storage Classes. As before, this might not be secure in every environment so proceed with caution. 

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: view-snapshot_storage_classes
rules:
  - apiGroups: ["snapshot.storage.k8s.io"]
    resources: ["volumesnapshotclasses"]
    verbs: ["get", "list", "watch"]
  - apiGroups: ["storage.k8s.io"]
    resources: ["storageclasses"]
    verbs: ["get", "list"]
```
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: view-snapshotclasses-nb-admin
  namespace: admin
subjects:
- kind: ServiceAccount
  name: default-editor
  namespace: admin
roleRef:
  kind: ClusterRole
  name: view-snapshot_storage_classes
  apiGroup: rbac.authorization.k8s.io
```

mr-yaky:
@DavidSpek do I understand correctly that by default Kale can stores snapshots only in Rok storage from Arrikto project? 
I'm now fighting with following:
```
Message: compile_notebook() got an unexpected keyword argument 'auto_snapshot'
```
I guess I must have this Rok storage as additional component. If MiniKf is free to check Kale locally then maybe should I switch it at the begging ?

And how is going with testing your code from this PR? I just see here the WIP status, so I think it's still in developing, right ?

mr-yaky:
I have separated Ceph cluster with CephFS and it already connected with K8s with using CSI. So, maybe I can check it with RWX volume from CephFS storage class.

DavidSpek:
@mr-yaky Kale at the present time only supports Rok. MiniKF includes Rok so you can use that to test with all features enabled. 

As you can see I haven't looked at this code in a month now, as I was busy with a KFP PR to automatically set the namespace in notebooks (which has now been merged) along with other work. 
The current state of this PR is that all the functions work when running them manually. It also correctly checks what snapshot providers you have and if the PVC belongs to a storage class that has a snapshot class associated with it. At that point you will also be able to use the buttons in the UI, and you can have it create a snapshot when you enable `use this notebook's volume`. However, when choosing `use this notebook's volume` it will error on trying to figure out if the snapshot was successfully created. So at this point, it is not yet completely usable. 

The next step, once the above error is figured out (and probably a few more), is too look at how to properly implement the metadata. At the moment the snapshot name is hardcoded, thus creating a second snapshot from a notebook will cause an error. This could be solved by adding a UUID, however I was thinking a meaningful UUID such as the snapshot content might be more useful for backtracking a snapshots source later on. Similarly, while it is already possible to restore a notebook from a snapshot of a pvc that was mounted to the notebook (even if it had multiple PVCs) by creating new PVC(s) from the snapshot, there should be a closer look at what all the relevant metadata is that should be used for this. 

I've been in contact with @StefanoFioravanzo about this for a while, and I hope to have some feedback on the code and the implementation soon to be able to look at this further. If you have the time and energy to test out this PR and provide feedback and suggestions it is very welcome as this is my first ever attempt at programming. If you have deployed the `external-snapshotter` to your cluster and have a snapshot class for CephFS setup you should be able to test this with RWX volumes (but seeing as it uses the generic CSI driver it should work for all storage backends that have snapshot support).

mr-yaky:
Got it, then I'll try to test first with MiniKF and I'll be able to test this PR on next integration step for Kubeflow.
Thank you for the good explanation about the status of PR!

backend/kale/common/snapshotutils.py
```diff
@@ -0,0 +1,356 @@
...
+            image=podutils.get_docker_base_image(),
+            path=i[0],
+            pod=pod_name,
+            default_container=podutils.get_container_name(),
+            is_workspace_dir=str(podutils.is_workspace_dir(i[0])))
+        snapshot_names.append("nb-snapshot-" + i[1])
+    return snapshot_names
+
+
+def check_snapshot_status(snapshot_name):
```
DavidSpek, line 145: This is what is causing erros in Kale after the snapshot has been created.

PR #219, opened 2020-09-28
Add setting the namespace for kfp in multi-user environments

DavidSpek:
Fix for pullrequest #213 to clean up the commits update to current master.
Sets the k8s namespace that the Jupyter Notebook Server is running in for KFP to allow Kale to properly list existing experiments. Related to issue #210.

### DIFF
--- a/backend/kale/rpc/kfp.py
+++ b/backend/kale/rpc/kfp.py
@@ -12,10 +12,12 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import os
 import kfp
 from kfp_server_api.rest import ApiException
 
 from kale.common import kfputils
+from kale.common import podutils
 
 _client = None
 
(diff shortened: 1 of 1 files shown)


elikatsis:
Hi @DavidSpek,

I believe this issue must be fixed on the KFP Client side. Kale should just be setting the namespace [if needed] and not inherit and tackle other software's bad handling.
For example, what if the config path changes? That's code in the KFP repository which shouldn't affect the code in this repository.

DavidSpek:
@elikatsis I agree that the config path changing or something similar would indeed be troublesome for Kale. However, it does clearly state that `set_user_namespace` is needed for Kubeflow Pipelines in the multi-user mode. The alternative is to always provide the additional namespace argument when using the KFP client, rather than setting a default which I have done with this pull request. Would adding the namespace to calls using the KFP client be the preferred way?

elikatsis:
@DavidSpek, yes `set_user_namespace()` is required. No argument against that.

However, what you do in this PR is fix something which should be fixed on the KFP SDK side. I strongly believe you should PR this on the KFP side.

DavidSpek:
@elikatsis  I'll take adding this to the KFP SDK once I finish up some more work on the cephfs integration backend.

DavidSpek:
@elikatsis I have created a similar PR for the KFP SDK. https://github.com/kubeflow/pipelines/pull/4638

DavidSpek:
After a long journey this has finally been implemented and merged into KFP in https://github.com/kubeflow/pipelines/commit/0df9473bbaecc062284a92e012394cdc95a70f1f, so I will be closing this PR.

PR #287, opened 2021-01-25
Update to Jupyterlab 3

davidspek:
Quick upgrade to jupyterlab 3 by running the script. Not yet manually tested. A build workflow is added to build the extension as a python package.


konsloiz:
Hi @DavidSpek! Is this solution working for you? We are looking for ways to upgrade it as well! Thanks.

PR #293, opened 2021-02-01
(Backend): Check if snapshots are available for notebook pvcs

DavidSpek:

This code adds the functionality to check if a snapshot provisioner exists in the cluster, and if the storage provider of the PVCs for the notebook have a snapshot class associated with them. Additionally, the check that restricted using snapshots to `ROK_CSI_STORAGE_CLASS` has been expanded for the previously mentioned checks.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-    "python.linting.flake8Enabled": true,
-    "python.linting.enabled": true
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "python.linting.flake8Enabled": true,
+    "python.linting.enabled": true
+}
--- a/backend/kale/common/podutils.py
+++ b/backend/kale/common/podutils.py
@@ -317,13 +317,13 @@ def get_snapshotclasses(label_selector=""):
         version="v1beta1",
         plural="volumesnapshotclasses",
         label_selector=label_selector)
-    return snapshotclasses
+    return snapshotclasses.get("items")
 
 
 def list_snapshotclass_storage_provisioners(label_selector=""):
     """List the storage provisioners of the snapshotclasses."""
-    snapshotclasses = get_snapshotclasses(label_selector)["items"]
-    return [snap_prov["dirver"] for snap_prov in snapshotclasses]
+    return [snap_prov["driver"] for
+            snap_prov in get_snapshotclasses(label_selector)]
 
 
 def check_snapshot_availability():
--- a/backend/kale/common/podutils.py
+++ b/backend/kale/common/podutils.py
@@ -145,6 +145,8 @@ def _get_mount_path(container, volume):
 def _list_volumes(client, namespace, pod_name, container_name):
     pod = client.read_namespaced_pod(pod_name, namespace)
     container = _get_pod_container(pod, container_name)
+    snapshotclass_provisioners = list_snapshotclass_storage_provisioners()
+    provisioner_names = get_snapshotclass_provisioners_names()
 
     rok_volumes = []
     for volume in pod.spec.volumes:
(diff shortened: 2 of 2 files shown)


backend/kale/common/podutils.py
```diff
@@ -300,3 +307,63 @@ def compute_component_id(pod):
...
+    return snapshotclasses
+
+
+def list_snapshotclass_storage_provisioners(label_selector=""):
+    """List the storage provisioners of the snapshotclasses."""
+    snapshotclasses = get_snapshotclasses(label_selector)["items"]
+    snapshotclass_provisioners = []
+    for i in snapshotclasses:
+        snapshotclass_provisioners.append(i["driver"])
+    return snapshotclass_provisioners
```
StefanoFioravanzo, line 330: Use a list comprehension here: `return [class["dirver"] for class in snapshotclasses]`

backend/kale/common/podutils.py
```diff
@@ -300,3 +307,63 @@ def compute_component_id(pod):
     component_id = component_name + "@sha256=" + component_spec_digest
     log.info("Computed component ID: %s", component_id)
     return component_id
+
+
+def get_snapshotclasses(label_selector=""):
```
StefanoFioravanzo, line 312: If you don't use the `label_selector` in future PRs, then let's remove it. Same thing for `list_snapshotclass_storage_provisioners`

backend/kale/common/podutils.py
```diff
@@ -145,6 +145,8 @@ def _get_mount_path(container, volume):
 def _list_volumes(client, namespace, pod_name, container_name):
     pod = client.read_namespaced_pod(pod_name, namespace)
     container = _get_pod_container(pod, container_name)
+    snapshotclass_provisioners = list_snapshotclass_storage_provisioners()
+    provisioner_names = get_snapshotclass_provisioners_names()
```
StefanoFioravanzo, line 149: Can you move these calls directly where they are used? Since both `snapshotclass_provisioners` and `provisioner_names` are used only once, I would avoid doing this assignment and just call the functions later on

backend/kale/common/podutils.py
```diff
@@ -300,3 +307,63 @@ def compute_component_id(pod):
...
+    client = k8sutils.get_storage_client()
+    classes = client.list_storage_class().items
+    snapshotclass_provisioners = list_snapshotclass_storage_provisioners()
+    storage_class_names = []
+    for i in classes:
+        name = i.metadata.name
+        provisioner = i.provisioner
+        if provisioner in snapshotclass_provisioners:
+            storage_class_names.append(name)
+    return storage_class_names
```
StefanoFioravanzo, line 369: same thing here, doing a list comprehension is more Pythonic:

```python
return [class.metadata.name for class in classes 
        if class.provisioner in snapshotclass_provisioners]
```

backend/kale/common/podutils.py
```diff
@@ -300,3 +307,63 @@ def compute_component_id(pod):
...
+    namespace = get_namespace()
+    pod_name = get_pod_name()
+    pod = client.read_namespaced_pod(pod_name, namespace)
+    snapshotclass_provisioners = list_snapshotclass_storage_provisioners()
+
+    for volume in pod.spec.volumes:
+        pvc = volume.persistent_volume_claim
+        if not pvc:
+            continue
+        pvc = client.read_namespaced_persistent_volume_claim(pvc.claim_name,
```
StefanoFioravanzo, line 345: Don't re-use the same variable name. Either name the first `pvc` differently, or this one

backend/kale/common/podutils.py
```diff
@@ -300,3 +307,63 @@ def compute_component_id(pod):
...
+                                                             namespace)
+        ann = pvc.metadata.annotations
+        provisioner = ann.get("volume.beta.kubernetes.io/storage-provisioner",
+                              None)
```
StefanoFioravanzo, line 349: No need to `ann.get("..", None)`. `get` already returns `None` by default, so you can just `ann.get("...")`

backend/kale/common/podutils.py
DavidSpek: I didn't change this from how it already was in Kale. https://github.com/kubeflow-kale/kale/blob/07ccdc5bdc4c39a3c04dad44d4c07447378efb98/backend/kale/common/podutils.py#L151

backend/kale/common/podutils.py
DavidSpek: Same here, this is how it is done in Kale currently. 
https://github.com/kubeflow-kale/kale/blob/07ccdc5bdc4c39a3c04dad44d4c07447378efb98/backend/kale/common/podutils.py#L168

backend/kale/common/podutils.py
DavidSpek: Label_selector is used in later logic.

backend/kale/common/podutils.py
DavidSpek: I also found this a strange way of doing things, but didn't think the fix should be included in the addition of these features.

DavidSpek:
@StefanoFioravanzo Seems like the PR is not picking up my new commit strangely enough.

backend/kale/common/podutils.py
StefanoFioravanzo: Yes you are referring to the `_list_volumes` function. Indeed, I will fix this in the original function as well

backend/kale/common/podutils.py
StefanoFioravanzo: So we did it wrong the first time :)

backend/kale/common/podutils.py
```diff
@@ -300,3 +306,55 @@ def compute_component_id(pod):
...
+def get_snapshotclasses(label_selector=""):
+    """List snapshotclasses."""
+    co_client = k8sutils.get_co_client()
+
+    snapshotclasses = co_client.list_cluster_custom_object(
+        group="snapshot.storage.k8s.io",
+        version="v1beta1",
+        plural="volumesnapshotclasses",
+        label_selector=label_selector)
+    return snapshotclasses
```
StefanoFioravanzo, line 320: Maybe return `snapshotclasses.get("items", [])` here? So you don't have to index the result in `list_snapshotclass_storage_provisioners`

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
StefanoFioravanzo, line 1: I guess you committed this file by mistake :)

DavidSpek: Yep, but it might be handy to have it, or add .vscode to the gitignore.

DavidSpek:
@StefanoFioravanzo I think it is good for merging now.

StefanoFioravanzo (approved):

PR #294, opened 2021-02-01
(Backend): Add ability to create a PVC snapsthot

davidspek:

This code adds the function that allows for creating a PVC snapshot. To ensure compatibility with all CSI storage drivers, a function is provided that gets the correct Volume Snapshot Class Name for a given PVC.

### DIFF
--- a/backend/kale/common/snapshotutils.py
+++ b/backend/kale/common/snapshotutils.py
@@ -1,0 +1,77 @@
+# Copyright 2020 The Kale Authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import logging
+
+from kale.common import podutils, k8sutils
+
+NOTEBOOK_SNAPSHOT_COMMIT_MESSAGE = """\
+This is a snapshot of notebook {} in namespace {}.
+This snapshot was created by Kale in order to clone the volumes of the notebook
+and use them to spawn a Kubeflow pipeline.\
+"""
+
+log = logging.getLogger(__name__)
+
+
+def snapshot_pvc(snapshot_name, pvc_name, labels, annotations):
+    """Perform a snapshot over a PVC."""
+    if annotations == {}:
+        annotations = {"access_mode": get_pvc_access_mode(pvc_name)}
+    snapshot_resource = {
+        "apiVersion": "snapshot.storage.k8s.io/v1beta1",
+        "kind": "VolumeSnapshot",
+        "metadata": {
+            "name": snapshot_name,
+            "annotations": annotations,
+            "labels": labels
+        },
+        "spec": {
+            "volumeSnapshotClassName": get_snapshotclass_name(pvc_name),
+            "source": {"persistentVolumeClaimName": pvc_name}
+        }
+    }
+    co_client = k8sutils.get_co_client()
+    namespace = podutils.get_namespace()
+    log.info("Taking a snapshot of PVC %s in namespace %s ...",
+             (pvc_name, namespace))
+    task_info = co_client.create_namespaced_custom_object(
+        group="snapshot.storage.k8s.io",
+        version="v1beta1",
+        namespace=namespace,
+        plural="volumesnapshots",
+        body=snapshot_resource)
+
+    return task_info
+
+
+def get_snapshotclass_name(pvc_name, label_selector=""):
+    """Get the Volume Snapshot Class Name for a PVC."""
+    client = k8sutils.get_v1_client()
+    namespace = podutils.get_namespace()
+    pvc = client.read_namespaced_persistent_volume_claim(pvc_name, namespace)
+    ann = pvc.metadata.annotations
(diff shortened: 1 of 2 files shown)


backend/kale/common/snapshotutils.py
```diff
@@ -0,0 +1,119 @@
+#  Licensed under the Apache License, Version 2.0 (the "License");
```
StefanoFioravanzo, line 1: The License header is missing the first line: `# Copyright 2020 The Kale Authors`

StefanoFioravanzo: And of course you can file a PR to be added to the `AUTHORS` file

backend/kale/common/snapshotutils.py
```diff
@@ -0,0 +1,119 @@
...
+        },
+        "spec": {
+            "volumeSnapshotClassName": get_snapshotclass_name(pvc_name),
+            "source": {"persistentVolumeClaimName": pvc_name}
+        }
+    }
+    co_client = k8sutils.get_co_client()
+    namespace = podutils.get_namespace()
+    log.info("Taking a snapshot of PVC %s in namespace %s ..."
+             % (pvc_name, namespace))
```
StefanoFioravanzo, line 96: You can use commas instead of `%` when logging. So this becomes

```python
log.info("Taking a snapshot of PVC %s in namespace %s ...", (pvc_name, namespace))
```

backend/kale/common/snapshotutils.py
```diff
@@ -0,0 +1,119 @@
...
+NOTEBOOK_SNAPSHOT_COMMIT_MESSAGE = """\
+This is a snapshot of notebook {} in namespace {}.
+This snapshot was created by Kale in order to clone the volumes of the notebook
+and use them to spawn a Kubeflow pipeline.\
+"""
+
+log = logging.getLogger(__name__)
+
+
+def snapshot_pvc(snapshot_name, pvc_name, image="", path="", **kwargs):
```
StefanoFioravanzo, line 75: Why not pass a general `annotations` argument here, instead of restricting the possible annotations to `image` and `path`?

backend/kale/common/snapshotutils.py
```diff
@@ -0,0 +1,119 @@
...
+        group="snapshot.storage.k8s.io",
+        version="v1beta1",
+        namespace=namespace,
+        plural="volumesnapshots",
+        body=snapshot_resource)
+
+    return task_info
+
+
+def get_snapshotclass_name(pvc_name, label_selector=""):
```
StefanoFioravanzo, line 107: Same argument about the `label_selector` here, if it's not needed le't remove it. Otherwise I'd like to know how it is useful

backend/kale/common/snapshotutils.py
```diff
@@ -0,0 +1,119 @@
...
+#    name: allow-snapshot-nb-admin
+#    namespace: admin
+#  subjects:
+#  - kind: ServiceAccount
+#    name: default-editor
+#    namespace: admin
+#  roleRef:
+#    kind: ClusterRole
+#    name: snapshot-access
+#    apiGroup: rbac.authorization.k8s.io
```
StefanoFioravanzo, line 60: In general, if you want to provide documentation for a module, use triple quotes. Look up "modules docstrings" to see some examples. Also

StefanoFioravanzo: Also, this doesn't seem the right place to document these things. If you feel like these are important let's have something under an `examples` folder or similar

backend/kale/common/snapshotutils.py
DavidSpek: That is handled by the `**kwargs`.

DavidSpek: I specified image and path as those are the most important ones for general functionality.

backend/kale/common/snapshotutils.py
DavidSpek: Agreed, it was more a development comment for others. Should probably create a README for this with some other info instead.

backend/kale/common/snapshotutils.py
DavidSpek: It is used to be able to determine if snapshots belong together during a restoration. 
https://github.com/kubeflow-kale/kale/pull/302/files#r569494524

backend/kale/common/snapshotutils.py
```diff
@@ -0,0 +1,72 @@
...
+def get_snapshotclass_name(pvc_name, label_selector=""):
+    """Get the Volume Snapshot Class Name for a PVC."""
+    client = k8sutils.get_v1_client()
+    namespace = podutils.get_namespace()
+    pvc = client.read_namespaced_persistent_volume_claim(pvc_name, namespace)
+    ann = pvc.metadata.annotations
+    provisioner = ann.get("volume.beta.kubernetes.io/storage-provisioner",
+                          None)
+    snapshotclasses = podutils.get_snapshotclasses(label_selector)["items"]
+    for i in snapshotclasses:
```
DavidSpek, line 69: Should use string compression here as well.

DavidSpek:
@StefanoFioravanzo I believe this PR is also ready for merging, I will then refactor the downstream ones so they include the latest changes of this PR.

StefanoFioravanzo:
@DavidSpek I had added a couple of additional review comments but forgot to submit the review...sorry. You should see them now

backend/kale/common/snapshotutils.py
DavidSpek: Makes sense. This will also make the expansion of the metadata that is written to the snapshots easier. I'll get to implementing this change ASAP.

DavidSpek:
@StefanoFioravanzo This PR should be good to go now

DavidSpek:
@StefanoFioravanzo maybe we can pick these PRs up again.

PR #295, opened 2021-02-01
(Backend): Add ability to create a snapshot of a pod

davidspek:

This PR adds the ability to snapshot all the volumes connected to a pod, along with writing the container images of that pod to the snapshot metadata. 
Note, at the moment the snapshot name is `pod-snapshot-` + `pvc_name`. This means snapshotting a pod twice will result in an error as the snapshot with that name already exists. This will need to be solved with a UUID of some sorts, I was hoping to find one that is useful/meaningful for other methods of identification as well, so I am open to suggestions.

### DIFF
--- a/backend/kale/common/snapshotutils.py
+++ b/backend/kale/common/snapshotutils.py
@@ -1,0 +1,77 @@
+# Copyright 2020 The Kale Authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import logging
+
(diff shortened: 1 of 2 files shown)


backend/kale/common/snapshotutils.py
```diff
@@ -117,3 +117,22 @@ def get_snapshotclass_name(pvc_name, label_selector=""):
...
+def snapshot_pod():
+    """Take snapshots of the current Pod's PVCs."""
+    volumes = [(path, volume.name, size)
+               for path, volume, size in podutils.list_volumes()]
+    namespace = podutils.get_namespace()
+    pod_name = podutils.get_pod_name()
+    log.info("Taking a snapshot of pod %s in namespace %s ..."
+             % (pod_name, namespace))
+    snapshot_names = []
+    for i in volumes:
```
StefanoFioravanzo, line 131: Can you change `i` to something more idiomatic, e.g. `vol`? And maybe assign `vol[1]` to `name` so that it's clear what you are doing

DavidSpek: Yeah, when I was splitting up the PRs I noticed this as well. Will change it ASAP.

DavidSpek:
@StefanoFioravanzo I believe this PR is ready for merging

DavidSpek:
@StefanoFioravanzo I have added a small function to generate 8 character long UUID's to ensure snapshot names are unique and to be able to have multiple "versions" of snapshots for a pod.

PR #296, opened 2021-02-01
(Backend): Add ability to create a snapshot of a notebook

DavidSpek:

This PR adds the ability to snapshot all the volumes connected to a notebook, along with writing the container images of that notebook and if a given volume is the workspace directory to the snapshot metadata.
Note, at the moment the snapshot name is `nb-snapshot-` + `pvc_name`. This means snapshotting a notebook twice will result in an error as the snapshot with that name already exists. This will need to be solved with a UUID of some sorts, I was hoping to find one that is useful/meaningful for other methods of identification as well, so I am open to suggestions.


DavidSpek:
@StefanoFioravanzo I believe this PR is also ready for merging. It is essentially the same as the snapshot_pod, but with some extra metadata.

PR #301, opened 2021-02-01
(Backend): Get all snapshots that originate from the same notebook

DavidSpek:

The function in this PR will search through the existing snapshots and find the snapshots originating from the same notebook. This function is used for restoring notebooks and is necessary to be able to restore all the volumes that were attached to the originating notebook.


DavidSpek:
This function will likely need to be reworked for a more specific selector.

PR #302, opened 2021-02-01
(Backend): Function to restore all PVCs originating from the same notebook snapshot

DavidSpek:

This PR adds the function that will restore all the the snapshots of each volume that was mounted to the notebook when the snapshot was taken. To do this it uses the `get_nb_pvcs_from_snapshot()` function from the previous commit that finds all the snapshots belonging to the same notebook snapshot.


backend/kale/common/snapshotutils.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+def get_nb_image_from_snapshot(snapshot_name):
+    """Get the image of the notebook that the snapshot was taken from."""
+    snapshot = get_pvc_snapshot(snapshot_name=snapshot_name)
+    image = snapshot["metadata"]["annotations"]["container_image"]
+    return image
+
+
+def get_nb_pvcs_from_snapshot(snapshot_name):
```
DavidSpek, line 303: @StefanoFioravanzo This is where the label_selector is used.

PR #336, opened 2021-04-15
Support both v1alpha3 and v1beta1 Katib versions

elikatsis:
Also make the trial image configurable via an environment variable


StefanoFioravanzo (approved):

Bowen0729:
> Also make the trial image configurable via an environment variable

What should I do?
I want to change the Trial image, and where should I configuration the environment variable

PR #447, opened 2025-07-17
feat: update backend to support KFP V2

Amrit27k:
This is the first PR of a full refactoring of the Kale codebase to support Kubeflow Pipelines V2 and JupyterLab 4.x

This PR includes all the backend changes to support thw new KFP V2 semantics, including logging artifacts natively with the KFP DSL, and not relying on local storage and volumes for data marshalling.

There is still a lot of refactoring required, including removing deprecated codepaths, updating tests, and refactoring some old logic that needs a refresh.

This PR provides a fully functional CLI-based notebook2KFP conversion path. It has been tested with:

- python 3.10
- Kubeflow Pipelines 2.4.0

---

Note:

- All previous tests are still passing but one: the unit tests for the `Compiler` class have been removed due to some changes in its logic. Some refactoring and new tests are due.


backend/kale/processors/nbprocessor.py
```diff
@@ -664,62 +682,156 @@ def dependencies_detection(self, imports_and_functions: str = ""):
                 marshal_candidates = astutils.get_marshal_candidates(
                     anc_source)
                 outs = ins_left.intersection(marshal_candidates)
-                # Remove the ins that have already been assigned to an ancestor
-                ins_left.difference_update(outs)
-                # Include free variables
-                to_remove = set()
+                print(f"Anc step {anc} marshal candidates inside dependency detect: {marshal_candidates}")
+                print(f"Anc step {anc} outs inside dependency detect: {outs}")
+                # # Remove the ins that have already been assigned to an ancestor
```
StefanoFioravanzo: @Amrit27k could you explain a bit the changes that you applies here? It seems like you introduced some new logic.

It seems like you wanted to keep the original code (what you left as commented code) but below modified it. I think that if you remove the commented code block, then the diffi should become easier to read and easier to compare with the previous version.

all-in-all, it would be great if you could describe how the new logic differs from the previous one

backend/kale/step.py
```diff
@@ -69,6 +78,24 @@ def __call__(self, *args, **kwargs):
         """Handler for when the @step decorated function is called."""
         return execution_handler(self, *args, **kwargs)
 
+    # helper method to manage artifacts
+    def add_artifact(self, artifact_name: str, artifact_type: str, is_input: bool):
+        # artifact_type will be like 'Dataset', 'Model', etc.
```
StefanoFioravanzo: @Amrit27k Can you add a proper docstring here? You can follow the same pattern as the other fns/methods you see around here

backend/kale/templates/new_nb_function_template.jinja2
```diff
@@ -0,0 +1,61 @@
...
+    from kale import marshal as _kale_marshal
+    _kale_marshal.set_data_dir("/marshal")
+{%- for output_art in step_outputs %}
+    # Save {{ output_art.name }} to output artifact
+    _kale_marshal.save({{ output_art.name }}, "{{ output_art.name }}", path={{ output_art.name }}.path)
+{%- endfor %}
+    # -----------------------DATA SAVING END-----------------------------------
+    '''
+
+    # run the code blocks inside a jupyter kernel
```
StefanoFioravanzo, line 45: @Amrit27k why is the data saving part before the user's code blocks?

backend/kale/cli.py
```diff
@@ -43,8 +46,180 @@
...
+#         stream_handler.setLevel(logging.DEBUG)
+
+#     if not os.getenv("NB_PREFIX"):
+#         logger.error("You may run this command only inside a Notebook Server.")
+#         sys.exit(1)
+
+#     args.func(args, logger)
+# Bypass Kale's installation check
+os.environ['KALE_DEV_MODE'] = '1'
+KALE_AVAILABLE = True
```
StefanoFioravanzo: @Amrit27k what are these two lines for?

backend/kale/cli.py
```diff
@@ -43,8 +46,180 @@
...
+#             )
+
+
+# KALE_VOLUMES_DESCRIPTION = """
+# Call kale-volumes to get information about Rok volumes currently mounted on
+# your Notebook Server.
+# """
+
+
+# def kale_volumes():
```
StefanoFioravanzo: @Amrit27k I think it's ok to run remove this function

backend/kale/cli.py
```diff
@@ -43,8 +46,180 @@
 """
 
 
+# def main():
```
StefanoFioravanzo: @Amrit27k I think it's ok to just remove this commented function, I see you are using it below with a few changes.

backend/kale/step.py
Amrit27k: Sure.

backend/kale/templates/new_nb_function_template.jinja2
Amrit27k: I did this to run the data saving portion as a code block at the end of user's code block

backend/kale/cli.py
Amrit27k: I can remove this, just made this during debugging mode.

backend/kale/cli.py
Amrit27k: Yes.

backend/kale/cli.py
Amrit27k: Ok

backend/kale/processors/nbprocessor.py
Amrit27k: Done.

PR #465, opened 2025-09-16
Readme updates

hmtosi:
As we begin installing and developing `kale v2.0`, we are finding a lot of small places that the `README.md` can be clarified. I am opening this PR to provide a place to make these updates.


hmtosi:
@jesuino this is ready for review - please make sure that the steps for installing the extension in dev mode are correct.

@StefanoFioravanzo we were wondering if you could take a quick look as well. Specifically, are the steps for a regular user correct? Should they be expected to use a `venv` or should we take that part out? And please let us know if anything else should be changed.

README.md
```diff
@@ -158,7 +158,8 @@ In order to see the Kale controls, you must open one of the curated example note
 <img alt="Kale JupyterLab Extension" src="docs/imgs/Extension.png"/>
 
 With this Kale Control Panel fully operational, you should be able to start a pipeline run from JupyterLab and see it appear in the `kfp` UI.
-Make sure to `Select experiment` and choose a `Pipeline Name`, and then hit `Compile and Run`. You will see progress bars appear.
+Make sure to `Select experiment` (ie. `Kale-Pipeline-Experiment`) and choose a `Pipeline Name` (ie. `candies-sharing`), and then hit `Compile and Run`. You will see progress bars appear. (NOTE: If you do not specify `Select Experiment` and instead let it use the `Default`, you may see the error message `An RPC Error has occurred`.)
```
jesuino, line 161: Is there an issue for this error? I would link the issue here so users can track the fix.

README.md
```diff
@@ -238,41 +240,20 @@ conda activate my_project_env
 # OR
 source .venv/bin/activate
 
-# make sure you have the correct version of jupyterlab
```
jesuino, line 241: because we are removing this we should either link to the `labextension/README.md`, or delete that file and bring the section [`Development Install`](https://github.com/kubeflow-kale/kale/tree/v2.0-frontend-dev/labextension#development-install) to here because the steps above are missing two important information about the lab extension in development mode:

* If we do only `pip install -e ".[test]"` on `labextension` then user will have to run this command every time they make a change to re-install the python package. Ideally developers should also create the symlink to the extension using the command: `jupyter labextension develop . --overwrite`
* I am also missing the `jlpm watch` command, which can be used for hot replacement of files and auto build when the source code is modified

README.md
```diff
@@ -187,23 +188,33 @@ limitations imposed by the Kale data marshalling model.
...
+
 # install the extension in dev mode
 pip install -e .[dev]
+
+# checkout to the labextension/ directory
+cd ..
+cd labextension
+
+#install the Jupyter extension
+pip install -e ".[test]"
```
jesuino, line 217: I really like having both build steps (Kale and Lab Extension) on this section. We saw that initial contributors may forget to activate the Kale backend API when running the lab extension locally and having both build steps together may help to avoid confusion.
I would only add the steps I mentioned in my other comment.

hmtosi:
closing due to complications from merging `main` <- `v2.0-frontend-dev`, work will be merged in #483

PR #470, opened 2025-09-20
Starting an extension from scratch with tests and backend

jesuino:
In this PR the build for the frontend is fixed. It also includes a few outdated code removal and tests and a backend for the extension is added. Notice that this is not the backend code from the root directory, it is the backend for the extension and I will elaborate more why it is here.

### Motivation

* The current front-end architecture for the extension does not align with common architecture from the community.
* To run in dev mode, the user must manually copy and paste some files, rerun the build script, and then restart JupyterLab.
* This process is flaky and leaves a lot of room for user error. Moreover, the `static` files that are copy-and-pasted are not meant to be pushed to github.
* This PR changes the kale architecture to bring it in line with other extensions and resolve flaky behavior.

### How to build the extension after these changes

The steps to build the extension `kubeflow-kale-labextension` are:

* go into `./labextension`
* Follow the steps from section `Development install` from `labextension/README.md`

This is the frontend installation, doing this is not enough because the extension is very tight to the backend execution code. In fact, if you only install the extension and start jupyter lab, you will see the following in browser console:

image

To install the backend you must follow the instructions from the root README.md for the backend.

The things I added for the extension I do believe that will be useful in a near future (tests, binder and so on)

Please try my branch and let me know if it works.


labextension/src/widgets/cell-metadata/InlineCellMetadata.tsx
```diff
@@ -247,7 +245,7 @@ export class InlineCellsMetadata extends React.Component<IProps, IState> {
           prevBlockNames: []
         };
       }
-      let previousBlockName = '';
+      let previousBlockName: any = '';
```
hmtosi: should we choose a more specific type than `any`?

labextension/src/components/Select.tsx
```diff
@@ -204,7 +204,7 @@ export const Select: React.FC<SelectProps> = props => {
       value={value || ''}
       variant={variant}
       helperText={helperText}
-      onChange={evt =>
+      onChange={(evt: any) =>
```
hmtosi: should we choose a more specific type for this than `any`?

labextension/README.md
```diff
@@ -1,9 +1,15 @@
...
+[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/kubeflow-kale-labextension/kale/main?urlpath=lab)
+
 
 Convert Notebooks to Kubeflow pipelines with Kale
 
+This extension is composed of a Python package named `kubeflow-kale-labextension`
+for the server extension and a NPM package named `kubeflow-kale-labextension`
+for the frontend extension.
+
 ## Requirements
```
hmtosi, line 12: Does this need to be done in a venv?

jesuino: The UI build is not required to be done in a venv, it can be done in a separated terminal.

jesuino:
Hello @hmtosi thanks for your review. 

>> labextension/ui-tests/yarn.lock

It is empty because we didn't build the UI, but it should be fine to keep it empty for now;

hmtosi (approved):
Thank you for the changes Will, it is all set to merge in my view

ederign (changes requested):
Great work @jesuino . Can you also fix our CI/CD? https://github.com/kubeflow-kale/kale/issues/472

jesuino:
the Ci/CD failures were because of my merge commit, which is reverted now!

ederign:
I talked with @jesuino and @hmtosi in DM, and we both agree that, as we are working on a feature branch, we can merge this and unblock @hmtosi contributions and @jesuino will work to fix the CI. We need to fix the CI as soon as possible to rebase this to main! /cc @StefanoFioravanzo

PR #477, opened 2025-09-26
Renaming extension backend to avoid problems with shared-data folder copy

### LINKED ISSUE #476: [bug] labextension pip install not installing static files
### Environment

labextension from `v2.0-frontend-dev` branch

*  How did you deploy Kale?

N/A

*  Kale version: 

Unreleased

### Steps to reproduce

* Run `pip install -e .` from the `kale/labextension` dir
* The package will be installed but jupyter lab will not list it as an extension and it will not work

### Expected result

* The extension should be usable after installing it via pip

### Materials and reference

We narrowed the issue down to the configuration in `pyproject.toml` not supporting directories, which is very weird. When doing a pip install -e . it should copy all the static content to the package shared data, but it is not copying it, only install.json is copied but it is mapped directly. The folder `~/.local/share/jupyter/labextensions/kubeflow-kale-labextension/` should have all static content:
```
$ ls ~/.local/share/jupyter/labextensions/kubeflow-kale-labextension/
install.json 
```
In my analysis the issue is with the shared-data not working for directories. This is mapped in `tool.hatch.build.targets.wheel.shared-data` in `pyproject.toml` I did try other folders and it is not working with folders at all, but when I map any file, it works.
```
[tool.hatch.build.targets.wheel.shared-data]
"kubeflow-kale-labextension/labextension" = "share/jupyter/labextensions/kubeflow-kale-labextension"
"install.json" = "share/jupyter/labextensions/kubeflow-kale-labextension/install.json"
"jupyter-config/server-config" = "etc/jupyter/jupyter_server_config.d"
```
I did manually copy the content of kubeflow-kale-labextension/labextension/ to ~/.local/share/jupyter/labextensions/kubeflow-kale-labextension/ and it works, so I am sure that the root cause is because this directory is not being copied. We can workaround this by using the symlink for local developments instead doing a pip install -e . (jupyter labextension develop --overwrite .), but bear in mind that the package is completely broken due this. FOr example, the `whl` file lacks the static content as well:

```
$ unzip kubeflow_kale_labextension-0.7.1-py3-none-any.whl 
Archive:  kubeflow_kale_labextension-0.7.1-py3-none-any.whl
  inflating: kubeflow-kale-labextension/_version.py  
  inflating: kubeflow_kale_labextension-0.7.1.data/data/share/jupyter/labextensions/kubeflow-kale-labextension/install.json  
  inflating: kubeflow_kale_labextension-0.7.1.dist-info/METADATA  
  inflating: kubeflow_kale_labextension-0.7.1.dist-info/WHEEL  
  inflating: kubeflow_kale_labextension-0.7.1.dist-info/licenses/LICENSE  
  inflating: kubeflow_kale_labextension-0.7.1.dist-info/RECORD
```
### Labels

/area frontend

---

jesuino:
This PR should close #476 and #475

When the backend base dir has hypen in the name, it breaks the exclude patterns which prevent the extension bundle to be copied into the installed package share directory.

This PR should fix this.


hmtosi (commented):
These changes look good to me. Following @jesuino's instructions for installing the extension, I was able to get it running, enable the `kale` dashboard, and successfully trigger a `kfp` run from the JupyterLab environment.

I see that the failing `flake8` action is also failing on `v2.0-frontend-dev`, and the community has agreed that can be fixed in a different PR, so I this can be merged as is in my opinion. 

It would be nice to update the `README.md` to include the new install instructions, but I'm not sure if that should happen here or separately.

hmtosi (approved):

ederign (approved):

PR #480, opened 2025-10-01
ISSUE-452: Fix Kale icon in LeftPanel

ada333:
The Kale icon in LeftPanel of JupyterLab was in a square. 
I think it was because the Jupyter Lab creates a default icon when you add new LeftPanel button so the Kale icon was on top of the default Jupyter Lab icon. 
I deleted the CSS class defining the Icon since it wasnt working. 
I added new LabIcon object (JupyterLab recommended way) with the same attributes as the CSS class and then added it when creating the widget: `widget.title.icon = kaleIcon`.
I tested it by running Jupyter Lab:
Before:
Screenshot From 2025-10-01 14-26-08

After:
Screenshot From 2025-10-01 14-24-07


labextension/src/widget.tsx
```diff
@@ -49,6 +50,15 @@ export interface IKubeflowKale {
 }
 
 const id = 'kubeflow-kale-labextension:deploymentPanel';
+
+const KALE_LOGO_SVG = `
```
jesuino: Would you please move this to a new file?

* create the svg file with this variable content in directory `style/icons`, for example see [trash.svg](https://github.com/opendatahub-io/odh-ide-extensions/tree/main/odh-jupyter-trash-cleanup/style/icons)
* Import the svg into this file,see the [trash example ](https://github.com/opendatahub-io/odh-ide-extensions/blob/main/odh-jupyter-trash-cleanup/src/TrashIcon.tsx#L2): 
```
import kaleIconSvg from '../style/icons/kale.svg';
// ...
const kaleIcon = new LabIcon({ name: 'kale:logo', svgstr: kaleIconSvg });
``` 

* We also need to make sure that Typescript recognizes SVG files, so creates a file named `svg.d.ts` in `src`:

```
declare module '*.svg' {
  const value: string;
  export default value;
}
```

jesuino (changes requested):
Hello @ada333 your changes look great to me, thanks! I would just suggest moving the SVG to a new file, this way we avoid noise in the code

jesuino:
Hello @ada333 

Thanks for the changes,

would you please run prettier?

I think `jlpm prettier` on `labextension` directory should do it.

```
 Checking formatting...
[warn] src/svg.d.ts
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```

Thanks!

jesuino (approved):
LGTM!

ederign:
Thanks @ada333 for your first contribution!

PR #491, opened 2025-10-14
feat(backend): support dynamic versioning for dev workflow

### LINKED ISSUE #449: Improve version management and dev workflow
## What happened  
Kale’s version is currently **hardcoded in the codebase** and needs to be manually updated in multiple places when publishing a new release.  
This creates a recursive dependency:  
- The version must be updated before publishing, or  
- It must be updated right after publishing, requiring another commit/PR.  

See for example: [PR #448](https://github.com/kubeflow-kale/kale/pull/448), where version bumps are performed manually across files.  

This makes it harder to test Kale with Kubeflow Pipelines (KFP), which installs Kale at runtime from PyPI and depends on the correct version being published.  

## What did you expect to happen  
- Kale should detect its version dynamically at runtime instead of relying on hardcoded strings.  
- KFP should always receive the correct version when it installs Kale.  
- Development and testing workflows should not require manual version bumps or unnecessary commits.  

## Anything else we need to know  
There are a few related improvements that would streamline development and testing:  

1. **Dynamic runtime version detection**  
   - Remove hardcoded version strings, so the version is always consistent with the package metadata.  

2. **Streamlined TestPyPI usage**  
   - Automate the dev versioning scheme (e.g., `x.y.z.devN`) so that dev builds can be published and tested without manual version bumps.  

3. **Local development index**  
   - Provide a way to bypass TestPyPI entirely by exposing a temporary local PyPI-compatible index that KFP can consume for installs.  

## Potential directions (not proposing final solutions yet)  
- Use runtime version detection from packaging metadata.  
- Adopt an automated dev versioning scheme.  
- Explore local PyPI-compatible indexes for faster iteration.

StefanoFioravanzo:
**Local development and PyPI workflow**

I decided to rely on `devpi` as the local PyPI-compatible index for Kale’s development builds.
This allows contributors to iterate quickly and test KFP integration without uploading every build to TestPyPI.

The dev/prod selection is now handled entirely through environment variables, keeping the build logic simple and isolated:

- `KALE_DEV_MODE=1` → enables dev mode and points Kale to a local devpi instance
- `KALE_DEVPI_SIMPLE_URL` → overrides the default http://127.0.0.1:3141/root/dev/+simple/
- `KALE_PIP_INDEX_URLS` → optional, comma-separated list of custom indexes (takes precedence over dev mode)

The CLI automatically sets these when the `--dev` flag is used, so users can switch between environments without changing code. The compiler now reads these variables to decide which index URLs to bake into generated components.

**build and version management**

I changed Kale’s packaging to use a `pyproject.toml` build system with setuptools_scm for versioning.
This removes the need to hardcode and manually bump the version in multiple places (as noted in the original issue).
The version is now derived automatically from Git tags:
	•	Tagged commit → stable version
	•	Commits after a tag → development build (<VERSION>+gHASH)

This makes the release and devpi workflows cleaner and ensures KFP always installs the exact version the build was created with.

Also added a CONTRIBUTOR.md guide that explains the use of the helper scripts and the environment variables.

Closes #449

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -7,7 +7,7 @@ name = "kubeflow-kale"
 # setuptools_scm handles the version automatically
 dynamic = ["version"]
 description = "Convert JupyterNotebooks to Kubeflow Pipelines deployments"
-readme = "../README.md"
+readme = "README.md"
 requires-python = ">=3.10.0"
 license = { text = "Apache License Version 2.0" }
 
--- a/backend/README.md
+++ b/backend/README.md
@@ -1,0 +1,5 @@
+# Kubeflow Kale Backend
+
+This package contains the Kubeflow Kale Python backend. For general
+project documentation and usage instructions, please refer to the
+repository root [README](../README.md).


StefanoFioravanzo:
One known issues I am left with is that when running the e2e tests we are matching the generated KFP DSL against a static reference, which cannot match the correct value of the Kale git version being used (that's why the e2e tests are failing in the CI right now)

One solution that comes to mind is to have a particular test flag that is passed to the compiler, so that we do not render the Kale version when running these tests. Not sure if there are other cleaner solutions..

StefanoFioravanzo:
@Amrit27k let me know if you want to take a look at this. Also @adrielparedes maybe someone from the team could help review this

cc @ederign

adrielparedes:
@jesuino @hmtosi can you take a look on this please!

Amrit27k:
Will take a look. Thanks!

CONTRIBUTE.md
```diff
@@ -0,0 +1,126 @@
...
+- Python 3.10 (managed via `pyenv`, `conda`, or your system package manager).
+- Recent `pip` and `virtualenv`/`venv`.
+- `node`/`yarn` are only required when working on the labextension (not covered here).
+- Optional but recommended: `devpi-server` and `devpi-client` for local package
+  publishing (see “Local package index”).
+
+## 2. Create a virtual environment
+
+```bash
+python3.10 -m venv .venv
```
hmtosi: On my system, creating the venv with `python3 -m venv .venv` led to a `jupyter` version conflict, but `python -m venv .venv` worked. Not a blocker, but it would be good to understand how this works across different systems

CONTRIBUTE.md
```diff
@@ -0,0 +1,126 @@
...
+- The labextension has independent tooling (`jlpm lint`, `jlpm prettier`). Run
+  those only if you modify the frontend package.
+
+## 8. Typical development checklist
+
+1. Activate your virtual environment.
+2. `pip install -e .[dev]`.
+3. Optionally start `devpi` via `source scripts/devpi-main.sh`.
+4. Implement and re-publish using `scripts/devpi-publish.sh` when you need a new
+   dev build.
```
hmtosi, line 119: Would this be whenever you make a change to the backend code and want to test it?

CONTRIBUTE.md
```diff
@@ -0,0 +1,126 @@
...
+5. Update fixtures (`backend/kale/tests/assets/kfp_dsl/*.py`) whenever template
+   output changes.
```
hmtosi, line 121: What is the template output/when would it change?

backend/scripts/devpi-main.sh
```diff
@@ -0,0 +1,11 @@
...
+./scripts/devpi-up.sh
+export KALE_DEV_MODE=1
+export KALE_DEVPI_SIMPLE_URL=http://127.0.0.1:3141/root/dev/+simple/
+
+# 2) Build and publish to local index (auto removes same dev version)
+./scripts/devpi-publish.sh
+
+# 3) Run your KFP pipeline generation/execution
+# The template now uses the local index while KALE_DEV_MODE=1 is set.
+# kale compile ...
```
hmtosi, line 11: Is this the same as starting the local `kubernetes` install? Perhaps we should include some examples to help the user understand (ie. `kubectl port-forward -n kubeflow svc/ml-pipeline-ui 8080:80`)

CONTRIBUTE.md
StefanoFioravanzo: Hmm that's interesting. Indeed in general it might be good to use whatever default version of `python` is available in the system/active venv. I'll change this to just `python`

CONTRIBUTE.md
StefanoFioravanzo: Well no. You'd run (2) just one to activate the interactive installation, then you also initiate devpi only once. It's step (4) that needs to be run every time you make changes, because you are publishing the new version to your local temporary pypi repo. I will clarify this

CONTRIBUTE.md
StefanoFioravanzo: The template output is the KFP DSL output that is generated using the jinja templates under the `backend/templates` folder. The `Compiler` uses those templates to generate the output script that is then used to build and upload the pipeline.

Our E2E tests work by doing a 1:1 comparison between a notebook conversion and its reference DSL script

backend/scripts/devpi-main.sh
StefanoFioravanzo: Well I feel like the instructions to setup KFP inside whatever type of K8s environment should be separate from this. This part indeed assumes that you do have a KFP env running so that Kale can compile and upload a pipeline. The point is that you can also compile a Kale notebook even if you don't have a KFP env running, but just want to see the KFP DSL output from the Kale conversion

CONTRIBUTE.md
Amrit27k: I believe, some changes are needed in the README file under 'CONTRIBUTE' section, maybe just a reference to the CONTRIBUTE.md with some explanations as devpi is really something which we were trying to look as local pypi few months ago, but we ended up using testpypi that time.

CONTRIBUTE.md
Amrit27k, line 1: wouldn't this be a better idea to put this file under /backend directory as this just explains the backend contribution details?

Amrit27k:
> One known issues I am left with is that when running the e2e tests we are matching the generated KFP DSL against a static reference, which cannot match the correct value of the Kale git version being used (that's why the e2e tests are failing in the CI right now)
> 
> One solution that comes to mind is to have a particular test flag that is passed to the compiler, so that we do not render the Kale version when running these tests. Not sure if there are other cleaner solutions..

Will this be a better idea? Like, we can use `kale.__version__` approach to get the installed dev build version name of kale in the e2e test script, print it, and modify the generated dsl script by just changing 'kubeflow-kale-<build-version>' if exists, to 'kubeflow-kale', then do an assert with golden file? We can also do the test-flag approach, as I'm just thinking to log the build-version for better readability to the dev. test-flag approach will need changes to backend, but the above mentioned approach might just need to modify e2e script. Just thinking out loud 🤔

backend/pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+requires = ["setuptools>=64", "setuptools_scm>=8", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "kubeflow-kale"
+# setuptools_scm handles the version automatically
+dynamic = ["version"]
+description = "Convert JupyterNotebooks to Kubeflow Pipelines deployments"
+readme = "README.md"
+requires-python = ">=3.10.0"
```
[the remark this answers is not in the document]
StefanoFioravanzo: Fixed

CONTRIBUTE.md
StefanoFioravanzo: @Amrit27k are you suggesting that using `devpi` didn't work for you?

StefanoFioravanzo: As far as the main README file is concerned: for sure I agree it needs updating. I would like to first merge this PR and then revisit the main READ to both validate the installation instructions AND to properly link it / update it wrt the new CONTRIBUTING documentation

CONTRIBUTE.md
StefanoFioravanzo: Ideally I would like this CONTRIBUTING file to also host guidelines for frontend development

CONTRIBUTE.md
hmtosi: @StefanoFioravanzo thank you for considering. On my side it's still showing `python3.10 -m venv .venv` on line 21. Is that what you intended?

CONTRIBUTE.md
hmtosi: @StefanoFioravanzo thank you for clarifying the steps. 

I agree, I think this section is fine in this PR, but it would be helpful to update the main `README.md` after this merges to point folks in the right direction for installing.

CONTRIBUTE.md
hmtosi: This is very helpful, thank you for explaining! It might be good to add a sentence about that in `Step 5` to help users understand when they should update the fixtures - I'm still not sure I would know when to do so.

If I was a user, it would help me to see something like...
5. Update fixtures (`backend/kale/tests/assets/kfp_dsl/*.py`) whenever template output changes. (In other words, whenever you run `xyz` script, which will update the KFP DSL template output.)

backend/scripts/devpi-main.sh
hmtosi: I see, I was mistaking `kale compile` for setting up the initial `kfp` environment. Is there a specific command for running the KFP pipeline generation?

CONTRIBUTE.md
Amrit27k: > @Amrit27k are you suggesting that using `devpi` didn't work for you?

No, I was not aware of 'devpi' before, this is something new for me. During the backend development, we went for 'testpypi', but if I had clue about this, then probably, we could have used 'devpi' during development itself. 😃

backend/scripts/devpi-main.sh
Amrit27k: Not sure what you mean by kfp pipeline generation, but this is the overall command which process the notebook, generates DSL, compiles the pipeline and runs it -
python ./backend/kale/cli.py --nb ./examples/base/candies_sharing.ipynb --kfp_host http://127.0.0.1:8080  --run_pipeline

There's no individual command yet which just runs the KFP pipeline. Does Line 11 needs to be replaced by the above command probably?

StefanoFioravanzo:
@hmtosi @Amrit27k tests are passing now.

Two things I changed:

1. here: https://github.com/kubeflow-kale/kale/pull/491/commits/ef0f2773c7d48c96fe2080b575a3bc3214c30c99 added a mock for the runtime Kale version to force it to resolve to just `kubeflow-kale` at test time.
2. https://github.com/kubeflow-kale/kale/pull/491/commits/75db621a08c3bc64c412d1a18bb15abfefbc0b44 added a backend readme otherwise `python build` cannot resolve the proejct root README during build time in the GH workflow (outside of the backend directory)

hmtosi (approved):

hmtosi:
@StefanoFioravanzo is this ready for @ederign to approve and merge it?

StefanoFioravanzo:
@hmtosi I feel like this is in a good state, we can iterate and improve in follow up PRs if we find out that the dev flow is still not ideal. @ederign please approve if are ok with the current state of this PR :)

ederign (approved):

ederign (changes requested):
@StefanoFioravanzo, probably we need to update: https://github.com/kubeflow-kale/kale/blob/dba8e05bcf611cce53fc2b886593f6475257989c/.github/workflows/build_backend.yml#L20 to 3.10.19? Check the CI failure.

ederign:
Should be fixed on: https://github.com/kubeflow-kale/kale/pull/505

PR #493, opened 2025-10-15
ISSUE-490: Fix error in webconsole when running pipeline

ada333:
Somewhere in Kubeflow the status was set to PENDING (all uppercase) but switch case accepted only Pending.
The easiest solution I could think of was normalize the status string before switch case to start with upper case and other letters lower case.

When running pipeline the error did not appear and pipeline passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ python -m venv .venv
 source .venv/bin/activate
 ```
 
-Install JupyterLab into your virtual environment. Then install the Kale backend from PyPI and the JupyterLab extension.
+Install the Kale backend from PyPI, and install the JupyterLab extension, `kubeflow-kale-labextension`.
 
 ```bash
 # activate your virtual environment if you haven't already
@@ -159,6 +159,7 @@ In order to see the Kale controls, you must open one of the curated example note
 
 With this Kale Control Panel fully operational, you should be able to start a pipeline run from JupyterLab and see it appear in the `kfp` UI.
 Make sure to `Select experiment` and choose a `Pipeline Name`, and then hit `Compile and Run`. You will see progress bars appear.
+
 Navigate to the `kfp` UI, refresh the page, and check that your new pipeline run has appeared here.
 
 ## FAQ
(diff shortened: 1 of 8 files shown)


labextension/src/widgets/deploys-progress/DeployProgress.tsx
```diff
@@ -89,7 +89,11 @@ export const DeployProgress: React.FunctionComponent<
   const getRunComponent = (pipeline: any) => {
     let IconComponent: any = UnknownIcon;
     let iconColor = '#5f6368';
-    switch (pipeline.status) {
+    // normalize pipeline status to match PipelineStatus enum
```
jesuino: Why not modify the [PipelineStatus enum](https://github.com/kubeflow-kale/kale/blob/main/labextension/src/widgets/deploys-progress/DeployProgress.tsx#L33) with the expected `pipeline.status` value?

ada333: Thanks for review!
you mean modify enum like this? :

     ERROR = 'Error',
     FAILED = 'Failed',
     PENDING = 'PENDING',
     RUNNING = 'Running',
     SKIPPED = 'Skipped',
     SUCCEEDED = 'Succeeded',
     TERMINATING = 'Terminating',
     TERMINATED = 'Terminated',
     UNKNOWN = 'Unknown'

only PENDING in uppercase or all of the statuses? because only Pending gives me the error.

hmtosi: Based on group discussion, `enum` should be modified like:
 ERROR = "ERROR"
 FAILED = "FAILED"
 PENDING = "PENDING",
 RUNNING = "Running",
 SKIPPED = "Skipped",
 SUCCEEDED = "Succeeded",
 TERMINATING = "Terminating",
 TERMINATED = "Terminated",
 UNKNOWN = "Unknown"

It doesn't look like `tsx` files differentiate between `'` and `"` but to match it to the [runtime state](https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/model/run.go#L29) I might put `"`

jesuino: Yes, I mean modify the enum to match the constants  for the v2 of pipelines API -> https://github.com/kubeflow/pipelines/blob/master/backend/src/apiserver/model/run.go#L28-L36

The enum is basically broken for v2, we don't see any more errors because this method is not called for other enum states and this is something that we likely should investigate as well.

hmtosi:
@ederign or @StefanoFioravanzo could you please merge this if it looks ok to you?

ederign (approved):

Issue #451, opened 2025-09-04
Error displayed opening a python notebook


ederign:
1. Open candies_sharing on v2.0-frontend-dev branch
2. Click on Kale extension
3. Click on Enable

<img width="1682" height="1222" alt="Image" src="https://github.com/user-attachments/assets/654f61a3-1151-4339-a53e-e427367a70bb" />

hmtosi:
@ederign I will work on this issue, will you please assign it to me?

jesuino:
I did change on my installation the method [`scrollToCell` in `NotebookUtils`](https://github.com/kubeflow-kale/kale/blob/v2.0-frontend-dev/labextension/src/lib/NotebookUtils.tsx#L555) to use the object cell instead the cell index and it seems to have fixed this error:

```
notebook.content.scrollToCell(cell.cell);
```

jesuino:
This should be fixed by this PR: https://github.com/kubeflow-kale/kale/pull/470

hmtosi:
I ran JupyterLab with the changes from #470. Opening a notebook does still throw a couple of errors, but not the weakmap warning.

<img width="2547" height="1234" alt="Image" src="https://github.com/user-attachments/assets/d87f99e8-1bcf-4dce-9af1-71c9bc497444" />
<img width="2547" height="1234" alt="Image" src="https://github.com/user-attachments/assets/83f1efad-2038-42be-aaa1-24f4235171b7" />

Once the notebook is opened and I have dismissed those two errors, I am able to "Enable" Kale with no further errors:

<img width="2547" height="1234" alt="Image" src="https://github.com/user-attachments/assets/529c59e1-2b0f-4bde-b781-ea0fec9dd367" />


Shall we close this ticket and open a new one for the other issues, or should we address those here as well?

StefanoFioravanzo:
@hmtosi I think we should keep this issue until we have a mechanism to nicely display any error that occurs when opening a notebook, removing these cryptic errors that are popping up now.
Do you have an idea whey we are still having them?

jesuino:
@StefanoFioravanzo me and @hmtosi  made a local research and the reason is that Kale is registering to listen to all errors that happens on the client and it includes even errors unrelated to Kale

[...]

PR #496, opened 2025-10-21
Issue 451 - fixing errors that occur when opening and closing notebooks

hmtosi:
**Overview**
As described in #451, errors with other installed labextensions can cause the `kale` UI to throw uncaught errors when `.ipynb` files are opened or closed. This was confusing, because many of the errors do not originate with `kale` at all.

**Current Changes**
This PR makes the following changes:
1. The errors are now logged in the console so a developer can find their actual root cause
2. The blocking modal has been removed, so a user can continue in `kale` without having to stop and dismiss the uncaught error
3. The error message now appears in a small, non-blocking toast instead
4. The toast parses the error message to identify the root cause of the error and posts that in the message, so that users have an easier time debugging these uncaught errors

**To Test/Reproduce**
1. Make sure you have a labextension that is installed but not properly functioning (ie. `Enabled` but not `OK`). For me, running `pip uninstall lckr-jupyterlab-variableinspector` achieves this. You can double check by running `jupyter labextension list` and looking for an extension with a red X next to it.
2. Start Jupyter lab
3. Open a notebook, and then close it
4. Likely, either opening or closing the notebook will trigger the error

**Remaining Questions**
1. Currently, the toast pop-up disappears after 7 seconds, and it has no buttons. Should I add a button at the bottom of the toast to dismiss it? Should I adjust the amount of time it displays for?
5. I only tested this with the `variable inspector` extension, which shows up in the error message as `@lckr/jupyterlab_variableinspector`. The string handling used to identify the error location is thus based off of this syntax. Other extensions may have different syntax and will not be handled as nicely, but for those I am printing the entire stack trace. Please let me know if I should adjust this at all.

### DIFF
--- a/labextension/src/widgets/deploys-progress/DeployProgress.tsx
+++ b/labextension/src/widgets/deploys-progress/DeployProgress.tsx
@@ -31,15 +31,15 @@ import DeployUtils from './DeployUtils';
 
 // From kubeflow/pipelines repo
 enum PipelineStatus {
-  ERROR = 'Error',
-  FAILED = 'Failed',
-  PENDING = 'Pending',
-  RUNNING = 'Running',
-  SKIPPED = 'Skipped',
-  SUCCEEDED = 'Succeeded',
-  TERMINATING = 'Terminating',
-  TERMINATED = 'Terminated',
-  UNKNOWN = 'Unknown'
+  ERROR = 'ERROR',
+  FAILED = 'FAILED',
+  PENDING = 'PENDING',
+  RUNNING = 'RUNNING',
+  SKIPPED = 'SKIPPED',
+  SUCCEEDED = 'SUCCEEDED',
+  TERMINATING = 'TERMINATING',
+  TERMINATED = 'TERMINATED',
+  UNKNOWN = 'UNKNOWN'
 }
 
 interface IDeployProgressProps extends DeployProgressState {


labextension/src/lib/RPCUtils.tsx
```diff
@@ -19,20 +19,41 @@ import { NotebookPanel } from '@jupyterlab/notebook';
 import { Kernel } from '@jupyterlab/services';
 import NotebookUtils from './NotebookUtils';
 import { isError, IError, IOutput } from '@jupyterlab/nbformat';
+import { ToastUtils } from './ToastUtils';
```
jesuino: Why not using the built in Notification API from Juypyter Lab?

See [examples](https://github.com/jupyterlab/extension-examples/tree/main/notifications)

An error Notification will require the following code:

```
import { Notification } from '@jupyterlab/apputils';

//...


// Create an error notification with an action button
Notification.error('Watch out something went wrong.', {
  actions: [
    { label: 'Help', callback: () => alert('This was a fake error.') }
  ],
  autoClose: 3000
});
```

labextension/src/lib/RPCUtils.tsx
```diff
@@ -19,20 +19,41 @@ import { NotebookPanel } from '@jupyterlab/notebook';
...
-      true,
-    ).then();
+    // pull the stacktrace for the unhandled error
+    const errorStack = event.reason.stack;
+    // isolate the url to find the root cause of the error
+    const stackLines = errorStack.split('\n');
+    const urlSplit = stackLines.slice(1,2).toString()
+
+    // call the toast pop up
+    if (urlSplit.includes('http://localhost:') && urlSplit.includes('lab/extensions/')){
```
jesuino: Will this line work for when JupyterLab is running on a server? Or: Will the stack trace always have `http://localhost` in it?

labextension/src/lib/RPCUtils.tsx
```diff
@@ -19,20 +19,41 @@ import { NotebookPanel } from '@jupyterlab/notebook';
...
+      const extensionName = extensionParts.join('/')
+      // print the jupyterlab extension name
+      ToastUtils.showErrorToast(
+        'Unhandled Labextension Error - please see the console',
+        `${event.reason.name}: ${event.reason.message}`,
+        extensionName
+      ).then();
+    } else {
+      // otherwise, print the stacktrace to the error location
+        ToastUtils.showErrorToast(
```
jesuino: As said before, I do recomment using the built in Jupyter notification API. It will handle the queue of Notifications so they won't override each other and we can save a lot of code change by using a few lines of the code already provided by Jupyter Lab.

jesuino (changes requested):
Thanks for your great work! I am wondering why not use the built-in Jupyter Lab notification library?

labextension/src/lib/RPCUtils.tsx
hmtosi: Thank you, I have addressed this. The code now uses the built-in functionality.

labextension/src/lib/RPCUtils.tsx
hmtosi: You are right, this was just an easy test to check that it was working. Thank you for flagging it. I have updated the string handling, but I don't know what the urls will be on the server. Please check again to see my changes and let me know if it needs further work.

labextension/src/lib/RPCUtils.tsx
```diff
@@ -19,20 +19,44 @@ import { NotebookPanel } from '@jupyterlab/notebook';
...
+          { label: 'Details', callback: () => alert(`${alert_string}${extensionName}`) }
+        ],
+        autoClose: 3000
+      });
+    } else {
+      // otherwise, print the stacktrace to the error location
+      const alert_string = 'Unhandled Error - please see console for more details. '
+      Notification.error(`${event.reason.name}: ${event.reason.message}`, {
+        actions: [
+          { label: 'Details', callback: () => alert(`${alert_string}${stackLines.slice(1,2)}`) }
```
jesuino: The alert error is not making any sense to me and I think that in production it will not make sense as well because usually we minify the JS code, which will make the stack trace look weird. Could we make a better error message than `stackLines.slice(1,2)`. Perhaps something like this:

```
      const errorDetails = ['Error Trace:'];
      stackLines.splice(0, 5)
        .filter((line: string) => line.trim().length > 0)
        .forEach((line: string) =>
          errorDetails.push(line)
        );
      errorDetails.push('...');
      errorDetails.push('Check console for full trace.');
Notification.error(`${event.reason.name}: ${event.reason.message}`, {
        actions: [
          { label: 'Details', callback: () => NotebookUtils.showMessage('Unhandled Error', errorDetails) }
        ],
        autoClose: 3000
      });
 ```

image

File: labextension/src/lib/RPCUtils.tsx [hunk 5]
```diff
@@ -19,20 +19,44 @@ import { NotebookPanel } from '@jupyterlab/notebook';
...
+      // if the error is caused by a jupyterlab extension, isolate the extension name
+      const urlSplit = stackLines.slice(1,2).toString()
+      const extensionSplit = urlSplit.split('@').slice(1,2).toString()
+      const extensionParts = extensionSplit.split('/')
+      extensionParts.pop();
+      const extensionName = extensionParts.join('/')
+      const alert_string = 'Unhandled Error - please see console for more details. Error originating from '
+      Notification.error(`${event.reason.name}: ${event.reason.message}`, {
+        actions: [
+          { label: 'Details', callback: () => alert(`${alert_string}${extensionName}`) }
```
jesuino: I would avoid using alert:

* It does not have a fixed style for all browsers
* We have a pop-up API from Jupyter Lab and you can use `NotebookUtils,showMessage` as a shortcut to use it:
Just be aware that `showMessage` receives a String for the PopUp title and a list of strings for the pop-up body:

```
      Notification.error(`${event.reason.name}: ${event.reason.message}`, {
        actions: [
          { label: 'Details', callback: () => NotebookUtils.showMessage(alert_string, [extensionName]) }
        ],
        autoClose: 3000
      });
```

Also please notice that if you decide to use the `showMessage` pop-up then I would make sure that the title is smaller than the text  `alert_string`.

jesuino (changes requested):
Hello @hmtosi 

It is look good and it works locally. I have a suggest change to avoid the use of `alert`. Fortunately Jupyter Lab and Kale have a few methods to show a dialog. Please let me know your thoughts about using it.

labextension/src/lib/RPCUtils.tsx
hmtosi: @jesuino does the string handling to find `lab/extension` look ok to you, or do you think I should change it? I don't think it's broad enough to cover all the issues we would see, but I thought we could start there as most of the errors seemed related to jupyter extension installs.

jesuino: The string handling code looks good, but I am afraid of extensionName being undefined or if the code throws an error, we must add tests later to harden this method. For now I would extract these lines to a method. So instead:

```
      const urlSplit = stackLines.slice(1,2).toString()
      const extensionSplit = urlSplit.split('@').slice(1,2).toString()
      const extensionParts = extensionSplit.split('/')
      extensionParts.pop();
      const extensionName = extensionParts.join('/')
 ```
 
 I would create a new function in module RPCUtils:
 
 ```

const extensionName = getExtensionName(stackLines);

//...

function getExtensionName(stackLines: any) {
  const urlSplit = stackLines.slice(1, 2).toString();
  const extensionSplit = urlSplit.split('@').slice(1, 2).toString();
  const extensionParts = extensionSplit.split('/');
  extensionParts.pop();
  const extensionName = extensionParts.join('/');
  return extensionName || 'unknown extension';
}
```
Because later we can unit test this function to harden it.

hmtosi: Thank you for the suggestion, I have made this change

labextension/src/lib/RPCUtils.tsx
hmtosi: thank you for the suggestion, I have made these changes.

labextension/src/lib/RPCUtils.tsx
hmtosi: for the time being, I have stopped printing any part of the stack trace, and changed the error message to direct the user to check the console. Perhaps as we explore further we will see more patterns that can be handled. Please let me know if that is ok with you

hmtosi:
@jesuino thank you for your review. I have made all suggested changes, except for the stack trace printing for non-extension errors. For simplicity at this time, we may just leave that out, and direct users to look at the console. Please let me know what you think.

jesuino (approved):
It is looking good to me, thanks!

labextension/src/lib/RPCUtils.tsx
```diff
@@ -19,23 +19,55 @@ import { NotebookPanel } from '@jupyterlab/notebook';
...
   }
 };
 
+function getExtensionName(stackLines: Array<string>){
+  const urlSplit = stackLines.slice(1,2).toString();
+  const extensionSplit = urlSplit.split('@').slice(1,2).toString();
+  const extensionParts = extensionSplit.split('/');
+  extensionParts.pop();
+  const extensionName = extensionParts.join('/');
+  return extensionName || 'unkonwn Jupyterlab extension';
```
jesuino: typo on `unkonwn`

hmtosi: fixed!

ada333:
I tried this in my local environment and it works as expected. The code looks good to me. Definitely a big upgrade from the previous error message!

PR #498, opened 2025-10-22
Add GitHub workflows for automated tag-based releases

StefanoFioravanzo:
> [!WARNING]
> This PR is built on top of https://github.com/kubeflow-kale/kale/pull/491. Disregard all commits but the last two ones. The only changes relevant to this PR are in the `.github/workflows` folder 
> This PR is dependent on #491  because it introduces a new pyproject.toml for the backend that changes the way the backend builds.

**Summary**

- Creates GitHub releases automatically for `v*` tags with generated notes so tagging alone publishes the release page.
- Publishes backend builds to PyPI and the JupyterLab extension to npm straight from the tagged revision once the release is published.
- Adds a manual dispatch mode to rehearse the release jobs against a supplied ref, either hitting the real registries or a safe “test” path using TestPyPI and npm dry runs.

**some details**

- `.github/workflows/create-release.yml:1` tag-driven release job that calls `softprops/action-gh-release@v2` with generated notes.
- `.github/workflows/publish-release.yml:5` listens to both real releases and `workflow_dispatch`, wiring inputs for the target ref and release/test mode while still skipping prereleases.
- `.github/workflows/publish-release.yml:41` builds the backend once and conditionally uploads to PyPI or TestPyPI (`TEST_PYPI_API_TOKEN`) depending on the mode.
- `.github/workflows/publish-release.yml:78` builds the labextension in the same run and either publishes to npm with the prod token or performs a dry-run plus `npm pack` using the staging token.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -3,19 +3,36 @@ name: Publish release artifacts
 on:
   release:
     types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Git ref (tag or branch) to publish or test'
+        required: true
+        type: string
+      mode:
+        description: 'Where to publish artifacts'
+        required: true
+        default: release
+        type: choice
+        options:
+          - release
+          - test
 
 permissions:
   contents: read
 
 jobs:
   publish-backend:
-    if: ${{ !github.event.release.prerelease }}
+    if: ${{ github.event_name != 'release' || !github.event.release.prerelease }}
     runs-on: ubuntu-latest
+    env:
+      RELEASE_MODE: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.mode || 'release' }}
+      RELEASE_TAG: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.tag || github.event.release.tag_name }}
     steps:
       - name: Check out release tag
         uses: actions/checkout@v4
         with:
-          ref: ${{ github.event.release.tag_name }}
+          ref: ${{ env.RELEASE_TAG }}
           fetch-depth: 0
 
       - name: Set up Python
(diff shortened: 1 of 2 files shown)


StefanoFioravanzo:
@ederign @adrielparedes please see this PR as an initial proposal for an automated release workflow

CONTRIBUTE.md
```diff
@@ -0,0 +1,126 @@
...
+If you alter template output, update the generated files accordingly.
+
+## 7. Formatting and linting
+
+- Python formatting is enforced via the existing code style (PEP 8). Many generated
+  files run through `autopep8`.
+- The labextension has independent tooling (`jlpm lint`, `jlpm prettier`). Run
+  those only if you modify the frontend package.
+
+## 8. Typical development checklist
```
jesuino, line 113: I tried to follow these steps quite a few times and I noticed that the pipeline generated by Kale will still refer to localhost(127.0.0.1) devpi. But it fails due the fact that KFP is adding the pip index-url automatically to trusted host (see this [issue](https://github.com/kubeflow/pipelines/issues/11155)). 

In any case, if we fix that, we would still have to figure out how to fix the problem with an accessible devpi for KFP - in this case I think that the best solution would be running devpi on a pod in kubeflow namespace. (great thing that you made INDEX_URL a parameter!)

Does this make sense?

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,120 @@
+name: Publish release artifacts
```
jesuino, line 1: Could we also have a daily publish for Kale backend at least? This way we could use the latest published dev version without having to build the backend locally for developing for the frontend!

jesuino (changes requested):
Hello @StefanoFioravanzo Thank you so much for working on this.

I tried to run in dev mode, but I faced problems that I clarified in my comment.

Another thing I noticed is that `.venv/' is not in the gitignore. I was using conda and didn't notice it.

CONTRIBUTE.md
StefanoFioravanzo: I feel like there are two options:

1. for super simple local dev setups using a minimal k8s setup for KFP (e.g. via `kind`), exposing somehow the local devpi to the local k8s
2. indeed running devpi inside the cluster for testing scenario involving clusters running remotely or other use cases more difficult to manage. 

We should probably discuss this more in a separate issue, since this is more about the local dev experience rather than publishing a new release. The initial PR was clearly incomplete and cannot cover all use cases.

.github/workflows/publish-release.yml
StefanoFioravanzo: In principle I agree that having a nightly version would simplify fronted dev. Not sure I have seen projects doing this directly to PyPI. After a quick look online it seems to me that PyPI might cap due to size limits at some point, so nightly might not be recommended. Some folks do this using test.pipy.org (https://discuss.python.org/t/publishing-nightly-builds-on-test-pypi-org-with-a-time-based-retention-policy/3152)

We could think of publishing a nightly there, although there doesn't seems to be an automated garbage collection mechanism as well, not sure if at some point we will need to manually remove old versions.

Worth looking into what other projects are doing!

CONTRIBUTE.md
jesuino: It makes sense to me! Let's discuss somewhere else

hmtosi:
@StefanoFioravanzo is this PR ready to move from draft to review?

StefanoFioravanzo:
Hey @hmtosi there we a few things that @jesuino pointed out that I wanted to look at. Also, before moving forward with this I wanted to make sure that our workflow for testing a local version against a remote k8s cluster is flawless. I had started working on a proposal early last week but I didn't find the time to complete it :/

ederign:
Fixed here: https://github.com/kubeflow/kale/pull/673

PR #499, opened 2025-10-27
ISSUE-495: Avoid any type in typescript code

ada333:
I tried rewrite any types with the correct type for the variables. I made comments in some places (not all places where any is left), where I think it is not possible to use other type than any.

### DIFF
--- a/labextension/src/lib/NotebookUtils.tsx
+++ b/labextension/src/lib/NotebookUtils.tsx
@@ -1,18 +1,5 @@
-/*
- * Copyright 2019-2020 The Kale Authors
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2019–2025 The Kale Contributors.
 
 import { JSONObject, PartialJSONValue } from '@lumino/coreutils';
 import { Dialog, showDialog } from '@jupyterlab/apputils';
--- a/labextension/src/lib/TagsUtils.ts
+++ b/labextension/src/lib/TagsUtils.ts
@@ -1,18 +1,5 @@
-/*
- * Copyright 2020 The Kale Authors
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2019–2025 The Kale Contributors.
 
 import { Notebook, NotebookPanel } from '@jupyterlab/notebook';
 import CellUtils from './CellUtils';
--- a/labextension/src/widgets/deploys-progress/DeployUtils.tsx
+++ b/labextension/src/widgets/deploys-progress/DeployUtils.tsx
@@ -1,18 +1,6 @@
-/*
- * Copyright 2020 The Kale Authors
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2019–2025 The Kale Contributors.
+
 import * as React from 'react';
 import WarningIcon from '@mui/icons-material/Warning';
 import InfoIcon from '@mui/icons-material/Info';
--- a/.github/workflows/build_backend.yml
+++ b/.github/workflows/build_backend.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Python 3.10
         uses: actions/setup-python@v1
         with:
-          python-version: 3.10.18
+          python-version: 3.10.19
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
@@ -34,7 +34,7 @@ jobs:
       - name: Set up Python 3.10
         uses: actions/setup-python@v1
         with:
-          python-version: 3.10.18
+          python-version: 3.10.19
       - name: Install flake8
         run: |
           pip install flake8 flake8-docstrings
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -1,0 +1,126 @@
+# Contributing to the Kale Backend
+
+This guide walks through setting up a local development environment for the Kale
+backend, explains how the package version flows through the system, and shows how
+to iterate safely using a temporary Python package index.
+
+> The backend lives under `backend/`. All commands below assume you run them from
+> the repository root unless noted otherwise.
+
+## 1. Prerequisites
+
+- Python 3.10 (managed via `pyenv`, `conda`, or your system package manager).
+- Recent `pip` and `virtualenv`/`venv`.
+- `node`/`yarn` are only required when working on the labextension (not covered here).
+- Optional but recommended: `devpi-server` and `devpi-client` for local package
+  publishing (see “Local package index”).
+
+## 2. Create a virtual environment
+
+```bash
+python -m venv .venv
+source .venv/bin/activate
+python -m pip install --upgrade pip
+```
+
+## 3. Install the backend in editable mode
+
+```bash
+cd backend
+pip install -e .[dev]
+```
+
+The `pip install -e .[dev]` step relies on `backend/pyproject.toml`:
+
+- Dependencies are declared in `[project] dependencies`.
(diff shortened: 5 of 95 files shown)


labextension/src/lib/CellUtils.ts
```diff
@@ -15,11 +17,13 @@ export default class CellUtilities {
...
    * @param notebook The notebook to get the cell from
    * @param index The index of the cell to read
-   * @returns any - A string value of the cell output from the specified
+   * @returns MultilineString | PartialJSONObject | null | undefined - A string value of the cell output from the specified
    * notebook and cell index, or null if there is no output.
    * @throws An error message if there are issues in getting the output
    */
-  public static readOutput(notebook: Notebook, index: number): any {
+  // here I think any type could be left because the return types are so messy 
+  // and I had to change line 202
```
hmtosi: I agree, it seems like this would be a place where `any` is acceptable

labextension/src/lib/CellUtils.ts
```diff
@@ -56,6 +60,7 @@ export default class CellUtilities {
    * @param key The key of the value.
    * @returns any - The value of the metadata. Returns null if the key doesn't exist.
    */
+  // I dont know what type to use here in this function
```
hmtosi, line 59: I believe the metadata object is json, if and when it exists, so perhaps PartialJSONObject could work here?

labextension/src/lib/NotebookUtils.tsx
```diff
@@ -338,7 +339,7 @@ export default class NotebookUtilities {
       ) {
         while (i < notebook.model.cells.length) {
           const currentCellModel = notebook.model.cells.get(i);
-          if (!isCodeCellModel(currentCellModel as any)) {
+          if (!currentCellModel || !isCodeCellModel(currentCellModel as CodeCellModel)) {
```
hmtosi, line 329: similar comment to line 330

labextension/src/lib/NotebookUtils.tsx
```diff
@@ -326,7 +327,7 @@ export default class NotebookUtilities {
 
     for (let i = 0; i < notebook.model.cells.length; i++) {
       const cellModel = notebook.model.cells.get(i);
-      if (!isCodeCellModel(cellModel as any)) {
+      if (!cellModel ||!isCodeCellModel(cellModel)) {
```
hmtosi: I'm not familiar with what this function is trying to do, but should it be `if (!cellModel && !isCodeCellModel` ? Because otherwise I think these options are mutually exclusive, so if one is true the other is not true, which then aligns with the if statement every time (ie. this if will be true every time)... or am I confused?

labextension/src/lib/CellUtils.ts
```diff
@@ -258,7 +265,9 @@ export default class CellUtilities {
 
     // Create a new cell - use different approaches based on available APIs
     let cell: ICellModel;
-    const model = notebook.model as any;
+    const model = notebook.model as any;
```
hmtosi: for this and 294, does the type have to be defined, or could we leave it as `const model = notebook.model` ?

labextension/src/lib/TagsUtils.ts
```diff
@@ -240,6 +241,7 @@ export default class TagsUtils {
     if (!(removedCell instanceof CodeCellModel)) {
       return;
     }
+    // no appropriate type to use here because either there is .get(tags) or .tags
```
hmtosi: In general for this situation, is it better to define it as `any` or just leave it with no defined type?

labextension/src/widgets/deploys-progress/DeployUtils.tsx
```diff
@@ -51,7 +51,7 @@ export default class DeployUtils {
     canceled: '#ff992a'
   };
 
-  public static getInfoBadge(title: string, content: any) {
+  public static getInfoBadge(title: string, content?: string[]) {
```
hmtosi: for both this and line 68, I believe it should  be `content: string[]` (remove the `?`)

both lines call the `NotebookUtils.showMessage` function which does not allow `content` to be null, so it should not  be an optional property - it should be required

labextension/src/lib/CellUtils.ts
```diff
@@ -15,11 +17,13 @@ export default class CellUtilities {
    * @description Reads the output at a cell within the specified notebook and returns it as a string
    * @param notebook The notebook to get the cell from
    * @param index The index of the cell to read
-   * @returns any - A string value of the cell output from the specified
+   * @returns MultilineString | PartialJSONObject | null | undefined - A string value of the cell output from the specified
    * notebook and cell index, or null if there is no output.
    * @throws An error message if there are issues in getting the output
    */
-  public static readOutput(notebook: Notebook, index: number): any {
+  // here I think any type could be left because the return types are so messy
```
jesuino: If you check the only place where this method is used the output is actually cast to `string`:

```
const output = CellUtilities.readOutput(notebook, index);
notebook.activeCellIndex = oldIndex;
return output as string;
```

Here's what I suggest:

* modify the method to return `string | undefined`
* on `return null` change to `return undefined`
* cast the returned values to string:

```
    if (out && isExecuteResult(out)) {
      return out.data['text/plain'] as string;
    }
    if (out && isStream(out)) {
      return out.text as string;
    }
```
This is not ideal, but will decrease the complexity, later we can revisit this code to make sure that it is up-to-date with later Jupyter Lab APIs.

labextension/src/lib/CellUtils.ts
```diff
@@ -141,8 +147,9 @@ export default class CellUtilities {
     let cell: ICellModel;
     for (let idx = 0; idx < cells.length; idx += 1) {
       cell = cells.get(idx);
-      const metadata = cell.metadata as any;
-
+      const metadata = cell.metadata as any; 
+      //here again any type is needed because key can be reached either by
+      // has function or by direct access
       // Safe metadata checking
       const hasKey = (metadata && typeof metadata.has === 'function')
```
jesuino, line 148: I do believe these changes were made before types, so possibly this could be simplified to:

```
const hasKey = cell.metadata[key] !== undefined;
```

The reason is because `cell.metadata` is of type `PartialJSONObject`, which allow us to address object using brackets.

Please notice that I didn't test this, if we decide to make this change, we must make sure that this does not break anything!

labextension/src/lib/CellUtils.ts
```diff
@@ -196,7 +203,7 @@ export default class CellUtilities {
       // await command.execute("notebook:run-cell");
       const output = CellUtilities.readOutput(notebook, index);
       notebook.activeCellIndex = oldIndex;
-      return output;
+      return output as string;
```
jesuino, line 200: you can remove this if you modify `readOutput` as I described above

labextension/src/lib/CellUtils.ts
```diff
@@ -282,7 +291,7 @@ export default class CellUtilities {
     if (oldIndex >= index) {
       oldIndex += 1;
     }
-    const cells = notebook.model.cells as any;
+    const cells = notebook.model.cells as any; // here I dont know what type to use
```
jesuino: I would remove the `as any` and then try to use the correct method. I explain:

```
      if (typeof cells.insert === 'function') {
        cells.insert(0, cell);
      } else if (typeof cells.insertAll === 'function') {
        cells.insertAll(0, [cell]);
      } else {
        // Fallback
        cells.unshift(cell);
      }
```

I do believe that `typeof cells.insertAll === 'function'` is trying to call the right method, but checking the object [CellList](https://jupyterlab.readthedocs.io/en/4.1.x/api/classes/notebook.CellList.html), we don't have methods `insert` or `insertAll`, so my guess is that this is dead code, but we need to further check to confirm this.

jesuino: @ada333 I found that this method is not even called anymore. The top level method that calls this is deprecated (see _runAndDelete_ on line 459) , so it is safe to remove it and all references. I did this locally and it seems to work. (see the [diff](https://gist.github.com/jesuino/973dd68cbaa2d6aaf39165d321c2f9e6))


In this case you can ignore this `any` and we can remove this later. I created an issue to address a future code cleanup: https://github.com/kubeflow-kale/kale/issues/502

labextension/src/lib/CellUtils.ts
ada333: It still complains because the return type can be `undefined`.

labextension/src/lib/CellUtils.ts
ada333: unfortunately no: `Property 'contentFactory' does not exist on type 'INotebookModel'`

labextension/src/lib/NotebookUtils.tsx
ada333: I added the `!cellModel` for safety, just to make sure that `isCodeCellModel()` is not called with null or undefined

labextension/src/lib/TagsUtils.ts
ada333: it would be better to leave it with no defined type, but it is not possible because of this code:
`if (metadata && typeof metadata.get === 'function')
 {
      tagsValue = metadata.get('tags');
    } else if (metadata && metadata.tags) {
      tagsValue = metadata.tags;
    } else {
      return; // No tags found
    }`
    I think some versions of JupyterLab have the metadata.get('tags') and some have metadata.tags and because of that there is no type that has both of these so it has to be `any` type.

labextension/src/lib/CellUtils.ts
ada333: I think that this is used because in different jupyterlab versions the cell.metadata has a different type, here is what I found:
Newer JupyterLab (v3.x and later): Cell metadata (cell.metadata) is an IObservableJSON object. This object acts like a Map and has specific methods for interaction, such as .has(key), .get(key), and .set(key).

Older JupyterLab (v1.x, v2.x): Cell metadata was often treated as a plain JavaScript object ({}). To check if a key existed, you would use standard object property access, like metadata[key] !== undefined.

hmtosi: so we should keep it as is, to support older versions of Jupyterlab, right?

jesuino (approved):
LGTM!

So far so good, we can revisit this topic again  once we clean the code and refactor to hooks!

labextension/src/lib/CellUtils.ts
ada333: I would be kind of scared to delete that code since it could change the functionality, and we would need to test it properly. But theoretically it should be okay to not support older versions since in ReadMe we require Jupyterlab version >= 4.0.0

hmtosi: That sounds good. We discussed as a team that we will check how to modernize this code in the future.

hmtosi:
@ederign we have agreed that our discussion about the changes is satisfied. If it looks good to you, could you please merge?

ederign:
@ada333 thank you so much for your contribution!!!!

PR #504, opened 2025-10-30
Issue 494 - fix RPC error and allow kfp to run with Default experiment

### LINKED ISSUE #494: [frontend] Running a pipeline with the `Default experiment` throws an RPC error
When a user has opened an `.ipynb` file and enabled the `kale` panel, they are at the point of hitting `Compile and Run`. However, if they do not  `Select Experiment` and instead let `kfp` use the `Default`, they may see the error message `An RPC Error has occurred`. 

We have two options for a bug fix, and we need to agree as a community:
1. Allow user to use the `Default` experiment (we will need to find the root cause of the error message and fix it)
2. Require user to select an experiment (we will need to change the error message to be more descriptive)

hmtosi:
Fixes #494 

When a user selected to `Compile and Run` a pipeline without first selecting an `Experiment`, the pipeline would not run, and an RPC error was thrown saying that `Experiment Name` was blank. We found this was because the code was retrieving the experiment names from the server, but it was not properly assigning them to the current notebook when it was compiled.

This PR fixes that, and also cleans up the code to automatically assign the first experiment's name, instead of resetting to a blank experiment, if no experiment is assigned.

### DIFF
--- a/labextension/src/widgets/LeftPanel.tsx
+++ b/labextension/src/widgets/LeftPanel.tsx
@@ -1,18 +1,5 @@
-/*
- * Copyright 2019-2020 The Kale Authors
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2019–2025 The Kale Contributors.
 
 import * as React from 'react';
 import { INotebookTracker, NotebookPanel } from '@jupyterlab/notebook';
--- a/.github/workflows/build_backend.yml
+++ b/.github/workflows/build_backend.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Python 3.10
         uses: actions/setup-python@v1
         with:
-          python-version: 3.10.18
+          python-version: 3.10.19
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
@@ -34,7 +34,7 @@ jobs:
       - name: Set up Python 3.10
         uses: actions/setup-python@v1
         with:
-          python-version: 3.10.18
+          python-version: 3.10.19
       - name: Install flake8
         run: |
           pip install flake8 flake8-docstrings
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -1,0 +1,126 @@
+# Contributing to the Kale Backend
+
+This guide walks through setting up a local development environment for the Kale
+backend, explains how the package version flows through the system, and shows how
+to iterate safely using a temporary Python package index.
+
+> The backend lives under `backend/`. All commands below assume you run them from
+> the repository root unless noted otherwise.
+
+## 1. Prerequisites
+
+- Python 3.10 (managed via `pyenv`, `conda`, or your system package manager).
+- Recent `pip` and `virtualenv`/`venv`.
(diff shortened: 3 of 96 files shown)


labextension/src/widgets/LeftPanel.tsx
```diff
@@ -274,7 +289,23 @@ export class KubeflowKaleLeftPanel extends React.Component<IProps, IState> {
...
           experiment_name = notebookMetadata['experiment_name'];
         } else {
-          this.resetState();
+          // If no experiment data exists, use the first experiment from the list if available
+          if (this.state.experiments.length > 0) {
+            experiment = this.state.experiments[0];
+            experiment_name = this.state.experiments[0].name;
+          } else {
+            // Keep existing experiment values if they exist, otherwise use defaults
+            if (
```
jesuino, line 302: I wonder if it wouldn't be easier to setup the experiment default value on line 260/257:

```
      // if the key exists in the notebook's metadata
      if (notebookMetadata) {
        let experiment: IExperiment = this.state.metadata.experiment;
        let experiment_name: string = this.state.metadata.experiment_name;
        if (notebookMetadata['experiment']) {
          experiment = {
            id: notebookMetadata['experiment']['id'] || this.state.metadata.experiment.id,
            name: notebookMetadata['experiment']['name'] || this.state.metadata.experiment.name
          };
          experiment_name = experiment.name;
```

With this I would say we should be able to avoid the code boilerplate.

labextension/src/widgets/LeftPanel.tsx
```diff
@@ -260,6 +262,19 @@ export class KubeflowKaleLeftPanel extends React.Component<IProps, IState> {
             name: notebookMetadata['experiment']['name'] || ''
           };
           experiment_name = notebookMetadata['experiment']['name'];
+          // If the experiment has empty values, use the first experiment from the list if available
+          const experimentsToUse =
```
jesuino, line 270: I didn't why keep the `fetchedExperiments` on a separated variable if you can get it from the state itself: `this.state.experiments`

jesuino (changes requested):
Great work! I see that it works well for when a notebook has metadata, but if fails for new notebooks or notebooks that has no metadata. The reason I think is because on the the [else clause for the notebook metadata check](https://github.com/kubeflow-kale/kale/blob/main/labextension/src/widgets/LeftPanel.tsx#L295) there's a state reset, making it lose the states that came from the server in the first place.

I quickly test the following and it seems to work for me:

```
        this.setState((prevState) => ({
          metadata: {
            ...DefaultState.metadata,
            experiment: prevState.metadata.experiment,
            experiment_name: prevState.metadata.experiment_name,
          }
```

I also added a few comments to the code changes, let me know if you have any question!

labextension/src/widgets/LeftPanel.tsx
hmtosi: Because `fetchedExperiments` is the `experiments` that were just newly fetched from the server with `commands.getExperiments(...)`, which are pulled as long as `this.props.backend` is true.  In the example you flagged, `this.state.experiments` refers to the `experiments` that come from the metadata, which is not necessarily up to date with the `experiments` that come from the server (I think). so using `fetchedExperiments` makes sure we're using the newest possible experiment data, but if it is empty, then we will revert to what is already in this.state.

labextension/src/widgets/LeftPanel.tsx
hmtosi: This is a good suggestion, thank you. I have made the change, but I haven't refactored any of the later code based on this. We can discuss refactoring once we have the extension running more smoothly I think

hmtosi:
@jesuino thank you for your review. I have implemented the metadata upgrade to stop the error from occurring on a new notebook. I also implemented the set up for the default experiment value. Please re-review, and let me know if they are the way you expect.

jesuino (approved):
It is working very well, thanks! It feels so good running pipelines now from Kale without having to select an experiment every time!


https://github.com/user-attachments/assets/23466b93-b2f3-4dc7-8e55-1062aa6de68f

hmtosi:
@ederign @StefanoFioravanzo this is ready for your review. Thanks!

ada333:
I tried this and it works great! The changes in code look good to me.

PR #516, opened 2025-11-13
Issue 515: fix warning symbol logic, clean up progress messages

### LINKED ISSUE #515: [frontend] Warning icons appear even if there is no warning
Currently when a pipeline is triggered from kale, all four steps have a Warning symbol next to them. Hovering over them shows a "Warning" message, but doesn't explain what the actual warning is. 

Image

I believe the code is showing these warnings whether or not one actually exists based on [this line](https://github.com/kubeflow-kale/kale/blob/7ef576cf07f09d33b9ee44b778169d16af889942/labextension/src/widgets/deploys-progress/DeployProgress.tsx#L332) (and similar ones for validating, compiling, and uploading). We should update this so the "Run Warnings" message only appears when there are warnings, and also display the error messages more clearly when they exist.

hmtosi:
Closes #515 

This PR removes the logic that was always displaying the warning symbol. The logic already exists for displaying an error if it occurs, so that was preserved in the PR. Now, an error will only be displayed when it has occurred.

There are some other small edits to further tidy the progress bars and their code.

### DIFF
--- a/backend/kale/common/kfputils.py
+++ b/backend/kale/common/kfputils.py
@@ -55,7 +55,7 @@ def get_pipeline_id(pipeline_name: str, host: str = None) -> str:
         f = next(
             filter(
                 lambda x: x.display_name == pipeline_name,
-                pipelines.pipelines
+                pipelines.pipelines or []
             ),
             None
         )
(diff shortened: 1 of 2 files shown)


jesuino (approved):
LGTM!

jesuino:
I could not reproduce a specific Kale warning, but forcing warnings worked very well to me:

image

Issue #500, opened 2025-10-27
[frontend] JupyterLab Dark theme hides some kale features


hmtosi:
If a user has JupyterLab's `Dark` or `Dark High Contrast` theme, the `X` button to exit the pencil icon menu is not visible (but it is still clickable). 

I have not done a full exploration to find what else may be affected.

<img width="2556" height="1394" alt="Image" src="https://github.com/user-attachments/assets/253582d5-52fa-4ad4-abf9-abcbad343d36" />

<img width="2556" height="1394" alt="Image" src="https://github.com/user-attachments/assets/e29249bc-d270-4e0a-9d87-e84d48cf0410" />

cordeirops:
I would like to work on this issue.

ederign:
@cordeirops great! Let's use slack to discuss any questions that you have on this one!

PR #517, opened 2025-11-13
fix issue #500 - JupyterLab Dark theme hides some kale features

cordeirops:
Closes https://github.com/kubeflow-kale/kale/issues/500 

This commit fixes some UI elements that were not properly visible when the JupyterLab Dark theme was enabled.

Since frontend development is not my strongest area, the current solution might not be the most optimal one, but it was the best approach I found to preserve functionality and visual consistency.

I’m open to suggestions or improvements if someone has a cleaner way to handle this.

The changes ensure that styles such as hover, focus, and other interactive states keep their intended behavior and original colors.

**Before & after:**

Untitled design

### DIFF
--- a/backend/kale/common/kfputils.py
+++ b/backend/kale/common/kfputils.py
@@ -55,7 +55,7 @@ def get_pipeline_id(pipeline_name: str, host: str = None) -> str:
         f = next(
             filter(
                 lambda x: x.display_name == pipeline_name,
-                pipelines.pipelines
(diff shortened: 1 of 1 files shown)


ederign:
@hmtosi can you test this one for us?

hmtosi (approved):
It works great, thank you for your contribution!

/lgtm

ederign (approved):

PR #520, opened 2025-11-18
Issue 489: Fix the links

ada333:
Right now it is a draft because the 8080 is hardcoded and I wasnt able to get the value any other way.

This PR should fix the links when running a notebook in Kale. 
Right now the localhost:8080 is hardcoded because calling .host or .uihost hasnt worked for me.
Also I added a better error log when opening a compiled path:
image

### DIFF
--- a/labextension/src/widgets/deploys-progress/DeployProgress.tsx
+++ b/labextension/src/widgets/deploys-progress/DeployProgress.tsx
@@ -31,13 +31,12 @@ enum PipelineStatus {
   UNKNOWN = 'UNKNOWN'
 }
 
-
 const logLinksHint = (kfpUiHost: string) => {
   console.info(
     `default for upload and run links is ${kfpUiHost} ` +
       'if your kpf ui is running somewhere else, set the KF_PIPELINES_UI_ENDPOINT environment variable.'
-    );
-  };
+  );
+};
 
 interface IDeployProgressProps extends DeployProgressState {
   onRemove?: () => void;
--- a/labextension/src/widgets/LeftPanel.tsx
+++ b/labextension/src/widgets/LeftPanel.tsx
@@ -213,8 +213,8 @@ export class KubeflowKaleLeftPanel extends React.Component<IProps, IState> {
       // wait for the session to be ready before reading metadata
       await notebook.sessionContext.ready;
 
-      const kfpUiHost =(await commands.getKfpUiHost()) || DEFAULT_UI_URL;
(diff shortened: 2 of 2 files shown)


labextension/src/lib/Commands.ts
```diff
@@ -132,6 +132,19 @@ export default class Commands {
...
 
+  getKfpUiHost = async (): Promise<string> => {
+    try {
+      return await _legacy_executeRpc(
+        this._notebook,
+        this._kernel,
+        'kfp.get_ui_host',
+      );
+    } catch (error) {
+      console.error('Failed to retrieve KFP UI host', error);
```
jesuino, line 143: Would you please give a little more detail on how user can set up this value?

labextension/src/widgets/deploys-progress/DeployProgress.tsx
```diff
@@ -38,24 +39,22 @@ export const DeployProgress: React.FunctionComponent<
...
   };
 
   const getRunLink = (run: RunPipeline) => {
-    // link: /_/pipeline/#/runs/details/<id>
-    // id = runPipeline.id
     if (!run.id) {
       return '#';
     }
-    const link = `${window.location.origin}/_/pipeline/#/runs/details/${run.id}`;
+    const base = props.kfpUiHost || 'http://localhost:8080';
```
jesuino: I would make the default URL a constant so it will be easier to modify in future.

const DEFAULT_UI_URL = 'http://localhost:8080'
...
const base = props.kfpUiHost || DEFAULT_UI_URL;
```

labextension/src/lib/Commands.ts
ada333: When the value is not set the backend function returns null and this error is not triggered. I am trying to think of the best solution for this - how to let the user know that he needs to set up the env var KF_PIPELINES_UI_ENDPOINT and that default is localhost:8080

labextension/src/widgets/deploys-progress/DeployProgress.tsx
hmtosi: Adam has made the change and the constant is easy to modify

labextension/src/lib/Commands.ts
hmtosi: I agree, we need to determine/document how to set up the env variable KF_PIPELINES_UI_ENDPOINT. I was not able to determine this in the code. However, I was able to reassign the default to `localhost:8081` and have that work successfully.

ederign (approved):

Issue #486, opened 2025-10-09
[frontend] Labels are Misplaced when scrolling the notebook


jesuino:
### Environment

Kale

*  How did you deploy Kale?
Dev mode from main

### Steps to reproduce
* Open an example that contains Kale metadata

<img width="1489" height="885" alt="Image" src="https://github.com/user-attachments/assets/5e39551b-4b44-4293-bad8-19893be23264" />

* Scroll to the bottom of the notebook and then scroll back to the top
* See that the labels are now misplaced - they should be before the cell, but they are after the code cell

<img width="1489" height="885" alt="Image" src="https://github.com/user-attachments/assets/58532f30-f8a4-4339-a20a-60cc0f1ff6d1" />

### Expected result

Labels should not change its location.

### Materials and Reference

n/a

jesuino:
It seems that it is not reproducible anymore with Jupyter 4.5.0

https://jupyterlab.readthedocs.io/en/latest/getting_started/changelog.html

PR #528, opened 2025-11-27
Making sure that the editor and the tags come before the cell

jesuino:
To solve issue [486 ](https://github.com/kubeflow-kale/kale/issues/486) I did make the notebook cell container a flex box and made sure that the editor will show first on it

TO test this fix you must open and scroll the notebook - before this fix you would notice that sometimes the editor was placed after the Notebook cell, now it show always come first.

### DIFF
--- a/labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
+++ b/labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
@@ -159,7 +159,7 @@ export class CellMetadataEditor extends React.Component<IProps, IState> {
 
     if (editor && inlineElement && !isEditorAlreadInPlace) {
       metadataWrapper.style.display = 'flex';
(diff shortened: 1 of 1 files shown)


ada333:
When I open dog-breed example I still can see the labels misplaced, Im not sure what the issue might be, I also checked if I have the right version of frontend (the flex box in the source code on line 161):
Screenshot From 2025-12-02 12-22-07

jesuino:
@ada333 thanks for your review. I am starting thinking that this is because of Jupyter Lab version or browser version. What versions do you have right now?

Thanks!

jesuino:
We are not able to reproduce this anymore after [Jupyter Lab 4.5.0](https://jupyterlab.readthedocs.io/en/latest/getting_started/changelog.html), so I am going ahead and closing this!

Issue #461, opened 2025-09-10
chore(backend): Create e2e test that runs with KFP


StefanoFioravanzo:
We need an E2E test in our GitHub CI that compiles a notebooks using the Kale CLI and runs it in a temporary K8s cluster with KFP, verifying the pipelines execution is successful

ada333:
I would like to work on this.

hmtosi:
@ederign can you please close this as the PR has been merged?

ederign:
Done!

PR #532, opened 2025-12-04
chore(gha): Create E2E test that runs with KFP

ada333:
This PR adds a GHA that installs kfp on kind cluster and runs the base Kale example (candies) and checks if run was successful.

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -2,7 +2,13 @@ name: E2E Test
 description: "Installs Kale and runs example notebook on KFP to verify end-to-end workflow"
 
 on:
-  workflow_dispatch
+  workflow_dispatch:
+    inputs:
+      pipeline_version:
+        description: 'KFP version to install'
+        required: false
+        default: '2.15.0'
+        type: string
 
 jobs:
   test:
@@ -19,7 +25,7 @@ jobs:
 
       - name: Install KFP
         run: |
-          export PIPELINE_VERSION=2.15.0
+          export PIPELINE_VERSION=${{ inputs.pipeline_version }}
           kubectl apply -k "github.com/kubeflow/pipelines/manifests/kustomize/cluster-scoped-resources?ref=$PIPELINE_VERSION"
           kubectl wait --for condition=established --timeout=60s crd/applications.app.k8s.io
           kubectl apply -k "github.com/kubeflow/pipelines/manifests/kustomize/env/platform-agnostic?ref=$PIPELINE_VERSION"


backend/kale/compiler.py
```diff
@@ -263,7 +263,7 @@ def _get_package_list_from_imports(self):
         """
         package_names = set()
         if KALE_VERSION != "0+unknown":
-            package_names.add(f"kubeflow-kale=={KALE_VERSION}")
+            package_names.add("dev-kubeflow-kale")
```
jesuino: As we talked, you could do this replace after the KFP YAML is generated and before submitting it to KFP;

Isn't it possible to let kind access the host machine and access the devpi server which kale dev package is published? Perhaps this kind configuration could help:

https://kind.sigs.k8s.io/docs/user/configuration/#extra-port-mappings

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,117 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      use-local-kale:
+        required: false
+        description: "Use the local kale version"
+        default: "true"
+      kale-installation-string:
+        required: false
+        description: "Installation string for kale, only used if use-local-kale is false (e.g. dev-kubeflow-kale or kubeflow-kale==0.7.0)"
```
hmtosi: For clarity, I would change this to `Installation string for kale (e.g. dev-kubeflow-kale or kubeflow-kale==0.7.0). Only used if use-local-kale is False.`

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,117 @@
...
+          python-version: '3.10'
+      - name: Install Libraries
+        run: |                  
+          python -m pip install --upgrade pip
+          if [ "$USE_LOCAL_KALE" = "true" ]; then
+            cd backend && pip install -e .[dev]
+          else
+            pip install $KALE_INSTALLATION_STRING
+          fi
+          echo "KALE_PIP_INDEX_URLS=https://test.pypi.org/simple" >> $GITHUB_ENV
```
hmtosi: This is the url for test PyPI, which is where `dev-kubeflow-kale` is published, but would it also need `https://pypi.org/simple` for `kubeflow-kale`?

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,118 @@
...
+          
+      - name: Run example notebook
+        run: |  
+          kale --nb examples/base/candies_sharing.ipynb \
+               --pipeline_name e2e-test \
+               --pipeline_descr "E2E Test" 
+
+          sed -i "s/'kubeflow-kale[^']*'/'dev-kubeflow-kale'/g" .kale/e2e-test.kale.py
```
hmtosi: Doesn't this cause it to automatically run the `dev-kubeflow-kale` version no matter what is set in the installation string?

.github/workflows/e2e-test.yml
ada333: I don´t think so.

jesuino: @hmtosi the production pypi is added anyway:

https://github.com/kubeflow-kale/kale/blob/main/backend/kale/common/utils.py#L242

.github/workflows/e2e-test.yml
jesuino: This must be changed, You could reuse the existing kale version parameter or have a new `kale_runtime_package_version` that would specify only the runtime kale version to be used when running the pipeline on KFP.

ederign (changes requested):
@ada333 I'll need to take a deep look on this to see what will be our strategy now that we are migrating to uv. But I'll review this as soon as I figure out a plan for how uv (and our artifacts should be used).

ederign:
@ada333 is this ready for review?

ada333:
@ederign @jesuino @hmtosi this PR is ready for re-review!

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,91 @@
...
+
+      - name: Kubernetes KinD Cluster
+        uses: helm/kind-action@v1.8.0
+        with:
+          version: v0.20.0
+          cluster_name: kind-kale-test
+
+      - name: Install KFP
+        run: |
+          export PIPELINE_VERSION=2.15.0
```
jesuino: I would make this a parameter because later we could test Kale with other KFP versions!

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Set KFP host address for Linux
+        run: |
+          HOST_IP=$(ip -4 addr show | grep -oP '(?<=inet\s)\d+(\.\d+){3}' | grep -v '127.0.0.1' | head -n1)
+          echo "KFP_HOST_ADDR=${HOST_IP}" >> $GITHUB_ENV
+      - name: Build and serve Kale wheel for KFP
+        run: |
+          make kfp-serve &
+          sleep 5
+
+      - name: Run example notebook
```
jesuino, line 62: Would it be possible to include a way to check the pipeline artifacts so we can make sure that they were correctly generated? At least one artifact check should be fine. Artifacts are important because the way that Kale transmits data between the notebook cells is by using artifacts.

ada333: @jesuino good idea, thanks. I am not sure how to check the artifact generation in code but I will try to figure it out.

ederign:
@ada333 can you please sign on the commit (for DCo)

ada333:
I was not sure how to add the artifact generation check so I just created an issue for it: https://github.com/kubeflow/kale/issues/565

hmtosi:
Closes #461

Issue #537, opened 2025-12-12
[backend] Allow code blocks to have the same name for input and output objects


hmtosi:
It is a common data science practice is to keep the same data frame table (without renaming) throughout the script, and perform manipulations directly on it. For example, the v1 `titanic-ml` example uses the `train_df` and `test_df` at every step.

However, when kale compiles the pipeline, the kfp dsl throws an error because there are steps with `train_df_artifact` or `test_df_artifact` as both an input and output. 

<img width="1400" height="626" alt="Image" src="https://github.com/user-attachments/assets/d5ae97e8-fb7e-4b2d-805f-6ebc53c0996d" />

This error is intended to prevent users from accidentally referencing the wrong object. However, to get the pipeline to run, the data frames had to be renamed at every step (as seen in #533, for v2 `titanic-ml`). This interrupts the data science workflow, and likely will confuse users. Thus, we should explore whether it is possible for the kfp dsl to accept the same name for an input or output.  

I think a good solution would be adjusting the compiler or the jinja templates to allow a user to input/output the same names, but have them be renamed based on the step in the backend.

ada333:
I would like to work on this

hmtosi:
Closed by #543

PR #543, opened 2026-01-06
feat(backend): Allow code blocks to have the same name for input and output objects 

ada333:
To allow code blocks having same input and output name I had to change 2 things in backend:

1. I added "_input" and "_output" to params of step function so something like this wouldn´t happen (python error because function can´t have 2 same params)
Before:
image

After:
image

2. Also there was a problem in Auto-generated pipeline function when initializing task and giving it input artifact,  the object referenced itself before it was created. So I changed it to find the correct ancestor of the artifact in compiler.py (line 220-230) and then the ancestor is assigned in new_pipeline_template.jinja2 (line 23).
Before:
image

After:
image

### DIFF
--- a/examples/base/candies_sharing.ipynb
+++ b/examples/base/candies_sharing.ipynb
@@ -207,7 +207,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.13.11"
+   "version": "3.10.18"
   }
  },
  "nbformat": 4,
--- a/backend/kale/tests/assets/kfp_dsl/iris.py
+++ b/backend/kale/tests/assets/kfp_dsl/iris.py
@@ -10,7 +10,7 @@ from kfp.dsl import Input, Output, Dataset, HTML, Metrics, ClassificationMetrics
     pip_index_urls=['https://pypi.org/simple'],
     pip_trusted_hosts=[]
 )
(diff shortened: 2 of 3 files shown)


hmtosi:
@ederign could you please assign me as a reviewer too?

ederign:
@ada333, can you please sign off on your commits?

examples/base/candies_sharing.ipynb
```diff
@@ -207,7 +207,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.18"
```
jesuino, line 210: I think this change can be discarded for this PR

ederign:
@jesuino could you please scan the PR again?

hmtosi:
lgtm
I have further tested the PR and I can confirm that the original issue from #537 is fixed. I created a small [example pipeline ](https://gist.github.com/hmtosi/c976fb257a3941168cd31e31d35b126e) that uses the same `dataFrame` name throughout every step. It runs successfully in `kfp` and does not throw the `RPC Error` listed in the ticket.

Two screenshots are below. One shows the example pipeline (notice the same `df` dataFrame at every step). The other shows the `kfp` run (notice that each pipeline step generates its own output object, and does not get blocked by the same name).
image

image

PR #573, opened 2026-01-31
Implementing GPU limit on Pipeline template

jesuino:
This is the implementation of the field GPU. It applies the change to the template using the respective methods for the Pipeline component:

```
    create_matrix_task.set_accelerator_type("nvidia.com/gpu")
                                      .set_accelerator_limit(2)
```

Now the GPU button on UI will be applied to the template, which will have effect on the node.

If you are going to run this on KFP make sure that you have the corresponding accelerator type available, otherwise the pod will never be created.

### DIFF
--- a/backend/kale/pipeline.py
+++ b/backend/kale/pipeline.py
@@ -99,7 +99,7 @@ class PipelineConfig(Config):
                           validators=[validators.PipelineNameValidator])
     experiment_name = Field(type=str, required=True)
     pipeline_description = Field(type=str, default="")
-    docker_image = Field(type=str, default="")
+    base_image = Field(type=str, default="")
     volumes = Field(type=list, items_config_type=VolumeConfig, default=[])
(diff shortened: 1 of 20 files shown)


backend/kale/templates/new_pipeline_template.jinja2
```diff
@@ -39,9 +39,17 @@ def auto_generated_pipeline(
     {{ step.name }}_task.after({{ steps_list[loop.index0 - 1].name }}_task)
     {%- endif %}
     {% endif %}
-    
+
     {{ step.name }}_task.set_display_name("{{ component_names[step.name] }}-step")
 
+    {%- if step.config.limits %}
+    {%- for limit_key, limit_value in step.config.limits.items() %}
+    {%- if limit_key in ['nvidia.com/gpu', 'amd.com/gpu'] %}
```
ada333, line 47: Could we get these values from one place in the code? The same array is also in UI.

PR #582, opened 2026-02-02
bug(backend): Pipeline Version Link Returns "Cannot retrieve pipeline…

### LINKED ISSUE #581: [backend] Pipeline Version Link Returns "Cannot retrieve pipeline version" Error
### Environment

*  How did you deploy Kale?

Following contribute.md

- **Kubeflow Pipelines**: v2.4.0+
- **Installation**: Official KFP installation via [Kubeflow Pipelines Installation documentation](https://www.kubeflow.org/docs/components/pipelines/operator-guides/installation/)
- **Kale**: Current master branch

### Steps to reproduce

When clicking the "Done" link after uploading a **new** pipeline in Kale, the user sometimes is redirected to an error page:

Image

> Cannot retrieve pipeline version. Click Details for more information.

The issue only occurs on the **first** deploy (new pipeline). Subsequent deploys to the same pipeline work correctly.

### Expected result

Open the current run on KFP.
 

---

ederign:
## Summary

Fixes race condition in `upload_pipeline()` that caused "Cannot retrieve pipeline version" errors when creating new pipelines.

- Removed unnecessary version creation/deletion logic that was prone to race conditions
- New pipelines now use the default version created by `upload_pipeline()` directly
- Existing pipeline uploads unchanged (still create versions with random names)

https://github.com/user-attachments/assets/36813c34-bbc0-448f-a577-eabaf262834b

Fixes #581

## Root Cause

The previous code created a pipeline (which auto-creates a default version), then created a second version with a random name, then tried to delete the default version by sorting versions by `created_at` timestamp.

If both versions had the same timestamp (common with second-precision timestamps), the sort order was undefined and could delete the wrong version - the one we wanted to keep. The returned version ID then pointed to a deleted version, causing the "Cannot retrieve pipeline version" error.

## Test Plan

- [ ] Deploy a **new** pipeline via Kale and click the "Done" link - should navigate to the pipeline version without error
- [ ] Deploy to an **existing** pipeline - should continue to work as before

### DIFF
--- a/backend/kale/common/kfputils.py
+++ b/backend/kale/common/kfputils.py
@@ -127,37 +127,27 @@ def upload_pipeline(
     client = _get_kfp_client(host)
     log.info("Uploading pipeline '%s'...", pipeline_name)
     pipeline_id = get_pipeline_id(pipeline_name, host=host)
-    version_name = utils.random_string()
     if not pipeline_id:
-        # The first version of the pipeline is set to the pipeline name value.
-        # To work around this, upload the first pipeline, then another one
-        # with a proper version name. Finally delete the original pipeline.
+        # New pipeline: upload_pipeline creates a default version automatically
         upp = client.upload_pipeline(
             pipeline_package_path=pipeline_package_path, pipeline_name=pipeline_name
         )
         pipeline_id = upp.pipeline_id
-        log.info("Uploaded Pipeline '%s' id: %s", pipeline_name, pipeline_id)
-        upv = client.upload_pipeline_version(
-            pipeline_package_path=pipeline_package_path,
-            pipeline_version_name=version_name,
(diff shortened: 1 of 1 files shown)


ederign:
@StefanoFioravanzo do you recall why the 'upload_pipeline' has such complex logic?

jesuino:
We saw this sometimes bu we were never able to consistently reproduce this! I do believe this may be related: https://github.com/kubeflow/kale/issues/522

jesuino (approved):
I did test starting multiple versions or multiple new pipelines and never had an error with this change, hence I do believe it is working as expected.

ederign:
@StefanoFioravanzo when you have a chance to review this!

StefanoFioravanzo (approved):

Issue #585, opened 2026-02-03
[feature] Set a default Base Image


jesuino:
### What feature would you like to see?

Now that Kale has [support to base images](https://github.com/kubeflow/kale/pull/571) configuration, we should bring the default value (currently `python:3.12`) to the UI and avoid the empty field 

### What is the use case or pain point?

The empty field is confusing.

### Is there a workaround currently?

Possibly modify the notebook metadata with an image value.

---

Love this idea? Give it a 👍.

Issue #574, opened 2026-02-01
feat: Searchable runtime image dropdown with defaults for per-step image selection


ederign:
### Feature Area

/area frontend
/area backend

### What feature would you like to see?

Add a searchable runtime image selector for per-step container images, while keeping support for free-text input.

Kale should provide:

- A searchable dropdown with prefilled runtime images
- Ability to still manually type an image when needed
- Support for a default runtime image by notebook, automatically selected
- Ability to define the default image inside the container (so environments can ship sensible defaults)

This builds on the existing per-step runtime image support and improves usability and standardization.

### What is the use case or pain point?

Manually typing Docker image names per step is error-prone and slow, especially when teams already have a known set of supported images. A searchable dropdown improves speed, correctness, and discoverability, while still allowing advanced users full flexibility.

### Is there a workaround currently?

Yes. Users manually type image names in step metadata or copy/paste them from examples. There is no catalog, search, validation, or default mechanism.

---

Love this idea? Give it a 👍.

ederign:
### Notes / Ideas (to kick off the discussion)

- Replace the plain image text field with a searchable selector, while keeping free-text input.
- Support defaults to reduce repetition.

Possible default resolution order:

1. Step-level image
2. Pipeline-level default
3. Container-defined default

Config options (can be layered):
- Env var for simple defaults (easy to ship in container images)
- JSON/YAML catalog file mounted or baked into the image
- Notebook/workspace metadata for user overrides

Ya-shh:
Hey @ederign 

This would be a really nice upgrade to the per-step image selection #571 we recently added — makes choosing runtimes much smoother, especially for teams with multiple images.

Quick proposed implementation plan:

**Backend (Image Catalog + Resolution):**
- Add `backend/kale/config/images.py` to load images from:

-   `KALE_DEFAULT_IMAGE` env var
- Optional catalog file (~/.kale/images.yaml or /etc/kale/images.yaml)

- Backend exposes the image catalog to the frontend

**Resolution order would be:**

1.   Per-step image (cell tag)
2.   Pipeline/notebook default (from notebook metadata)
3.   Container default (from env var or catalog file)


**Frontend:**
- Replace current text field with MUI `Autocomplete` (freeSolo = true)
- Fetch catalog on panel mount via RPC
- Grouped options: “Suggested” / “Recent” / “All available” / “Custom entry”

Happy to iterate on:
- YAML vs JSON preference for the catalog file
[...]

PR #593, opened 2026-02-04
feat(labextension): Display default base image in cell metadata UI

Ya-shh:
the default base image (python:3.12) in the cell metadata UI

### DIFF
--- a/backend/kale/compiler.py
+++ b/backend/kale/compiler.py
@@ -101,6 +101,10 @@ class Compiler:
 
         Returns (str): A Python executable script
         """
+        # Fail early if there are no steps in the pipeline.
+        if not hasattr(self.pipeline, "steps") or not self.pipeline.steps:
+            raise ValueError("Task is missing from pipeline.")
+
         # List of lightweight components generated code
         lightweight_components = [
             self.generate_lightweight_component(step) for step in self.pipeline.steps
--- a/backend/kale/marshal/backend.py
+++ b/backend/kale/marshal/backend.py
@@ -91,7 +91,7 @@ class MarshalBackend:
         saved file.
         """
         abs_path = os.path.join(get_data_dir(), name + "." + self.file_type)
-        log.info(
+        log.debug(
             "Saving %s object using %s: %s to %s", self.display_name, self.name, name, abs_path
         )
         try:
@@ -124,7 +124,7 @@ class MarshalBackend:
         directly this function instead of `load`.
         """
         abs_path = os.path.join(get_data_dir(), name + "." + self.file_type)
-        log.info("Loading %s file using %s: %s", self.display_name, self.name, name)
+        log.debug("Loading %s file using %s: %s", self.display_name, self.name, name)
         try:
             return self.load(abs_path)
         except ImportError as e:
@@ -285,9 +285,9 @@ class Dispatcher:
                 and os.path.splitext(ls)[0] == basename
             )
         ]
-        log.info("Found %d entries for basename '%s': %s", len(entries), basename, entries)
+        log.debug("Found %d entries for basename '%s': %s", len(entries), basename, entries)
         if not entries:
-            log.info(
+            log.debug(
                 "Looking for unique file/folder with basename '%s' in %s", basename, get_data_dir()
             )
             raise ValueError(
--- a/backend/kale/rpc/nb.py
+++ b/backend/kale/rpc/nb.py
@@ -20,7 +20,7 @@ from tabulate import tabulate
 
 from kale import Compiler, NotebookProcessor, marshal
 from kale.common import astutils, kfputils, kfutils, podutils
-from kale.rpc.errors import RPCInternalError
+from kale.rpc.errors import RPCInternalError, RPCUnhandledError
 from kale.rpc.log import create_adapter
 
 KALE_MARSHAL_DIR_POSTFIX = ".kale.marshal.dir"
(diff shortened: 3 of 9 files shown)


Ya-shh:
Base-Image
@ederign  Could you please review this?

Ya-shh:
Fixes -  #585

Ya-shh:
Base-Image

@ederign could you please review this?

Ya-shh:
Fixes -  #585

Ya-shh:
@jesuino could you please review this ?

Ya-shh:
@ederign & @StefanoFioravanzo ,could you please take a look at this when you have a moment?

If this approach works for the current issue, I'd be happy to start implementing the new feature discussed in #574.

Really excited to keep contributing to the project!

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
```diff
@@ -63,6 +63,8 @@ export const RESERVED_CELL_NAMES_CHIP_COLOR: { [id: string]: string } = {
   functions: 'a32626',
 };
 
+export const DEFAULT_BASE_IMAGE = 'python:3.12';
```
jesuino, line 73: Could we retrieve this information from the backend to keep it aligned between backend and frontend?

You can see how Adam implemented a backend property to retrieve the KFP host information. The same could be done for the base image: https://github.com/kubeflow/kale/pull/520/changes#diff-8d069efbd791d2303d5b2c423b84ae69d18021bd67a288594dfd17ed55cb812dR22

This way we won't have to update in two different places if we want to change the default base image.

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
```diff
@@ -459,14 +462,27 @@ export class CellMetadataEditor extends React.Component<IProps, IState> {
...
-                  label={'Base Image'}
-                  updateValue={this.updateBaseImage}
-                  value={this.props.baseImage || ''}
-                  placeholder="e.g., python:3.11"
+                <Button
                   variant="outlined"
-                  style={{ width: '25%' }}
-                />
+                  size="small"
+                  disabled
```
jesuino: Why is this field disabled? I can't change the base image anymore.

Image

Ya-shh:
@jesuino Thanks for the review! I'll fix these issues asap.

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
Ya-shh: Good point @jesuino ! The field was initially disabled because a previous review flagged the empty input as confusing for users — so I replaced it with a styled button showing the default image. I kept it disabled since we plan to introduce a searchable runtime image dropdown with defaults for per-step image selection in a future iteration. 

That said, your feedback makes sense — users should still be able to override the base image now. I've updated it to open an editable dialog on click (same pattern as the GPU dialog), while still displaying the default clearly. The default is now also fetched from the backend via RPC to keep frontend and backend aligned.

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
Ya-shh: Done @jesuino !

Ya-shh:
@jesuino please let me know your view on this?

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
```diff
@@ -459,14 +479,27 @@ export class CellMetadataEditor extends React.Component<IProps, IState> {
...
-                  placeholder="e.g., python:3.11"
+                <Button
                   variant="outlined"
-                  style={{ width: '25%' }}
-                />
+                  size="small"
+                  onClick={() => this.toggleBaseImageDialog()}
+                  style={{
+                    textTransform: 'none',
+                    minWidth: '150px',
```
jesuino: Would you please  check the CSS for this button? In generating setting a fixed size is not good for responsible screens, which means that it may mess the inline editor in small screens or not.

I also do believe this should be a button just like to GPU button, that do not show additional information because the base image information is always visible at the cell, it looks redundant!

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
```diff
@@ -459,14 +479,27 @@ export class CellMetadataEditor extends React.Component<IProps, IState> {
...
+                    backgroundColor: '#f5f5f5',
```
jesuino: I would avoid changing style in inline CSS like this because this may be broken if we change the look and feel of the application using the underlying framework. Since you are opening a Dialog Box, I would keep this as a simple button just like the GPU one.

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
jesuino: Still about the minWidth. Please notice that the other componens have a relative width which summed give it a 100% result. I would make this a button, set its width as `5%` and change the `Depends on` to have `30%`.

jesuino (changes requested):
Thank you again, @Ya-shh 

We are almost there!

One last comment is about the field. Would you please make it responsive like the other fields and make it a button like the GPU one? I think we don't need to repeat the Base Image information because it is always shown on the top of the cell.

Image

Thanks!

Ya-shh:
Thanks a lot for the detailed review! @jesuino
I will fix these issues asap🚀

Ya-shh:
BASE-Image
@jesuino  does it look good now?

jesuino:
Thank you, @Ya-shh 

It all look good to me code related, let me just ask around about your initial UI component and the button. What do you think, @ederign @StefanoFioravanzo @hmtosi , which style looks better?

image
image

When you click the IMAGE button this is what is displayed:

image



Thanks!

hmtosi:
@jesuino I think the button is great, thanks!

ederign:
I prefer this one:
image

ederign:
tested it locally and works! Thank you

/approve
/lgtm

PR #601, opened 2026-02-11
Add download button for compiled pipeline file

Ya-shh:
- Added a "Download" button with icon next to the "Done" status after compilation, allowing users to download the compiled .kale.py file directly from the progress panel

### DIFF
--- a/labextension/src/widgets/deploys-progress/DeployProgress.tsx
+++ b/labextension/src/widgets/deploys-progress/DeployProgress.tsx
@@ -21,6 +21,7 @@ import UnknownIcon from '@mui/icons-material/Help';
 import PendingIcon from '@mui/icons-material/Schedule';
 import SkippedIcon from '@mui/icons-material/SkipNext';
 import SuccessIcon from '@mui/icons-material/CheckCircle';
+import GetAppIcon from '@mui/icons-material/GetApp';
 
 import StatusRunning from '../../icons/statusRunning';
 import TerminatedIcon from '../../icons/statusTerminated';
@@ -166,6 +167,22 @@ export const DeployProgress: React.FunctionComponent<
     }
   };
 
+  const handleDownloadClick = () => {
+    if (props.compiledPath && props.compiledContent) {
+      const fileName =
+        props.compiledPath.split('/').pop() || 'pipeline.kale.py';
+      const blob = new Blob([props.compiledContent], { type: 'text/plain' });
+      const url = URL.createObjectURL(blob);
+      const a = document.createElement('a');
+      a.href = url;
+      a.download = fileName;
+      document.body.appendChild(a);
+      a.click();
+      document.body.removeChild(a);
+      URL.revokeObjectURL(url);
+    }
+  };
+
   // Handle close click safely
   const handleCloseClick = () => {
     if (props.onRemove) {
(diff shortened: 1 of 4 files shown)


Ya-shh:
@ederign could you please review?
Download-button

labextension/src/widgets/deploys-progress/DeployProgress.tsx
```diff
@@ -215,6 +240,15 @@ export const DeployProgress: React.FunctionComponent<
             style={{ color: DeployUtils.color.success, height: 18, width: 18 }}
           />
         </a>
+        <a
+          onClick={handleDownloadClick}
+          style={{ cursor: 'pointer', marginLeft: 4 }}
+          role="button"
+          tabIndex={0}
+        >
+          Download
```
jesuino: I would remove the text and leave only the icon

Makefile
```diff
@@ -190,7 +190,7 @@ lock-upgrade: check-uv ## Upgrade all dependencies and update lock
 ##@ Development Helpers
 
 jupyter: ## Start JupyterLab
-	SETUPTOOLS_SCM_PRETEND_VERSION=$(DEV_VERSION) $(UV) run jupyter lab
+	SETUPTOOLS_SCM_PRETEND_VERSION=$(DEV_VERSION) $(UV) run jupyter lab --ContentsManager.allow_hidden=True
```
jesuino: Will this be required in production to have this working? Notice that some users may want to keep files hidden, hence we must make sure that we document that this feature only works with this flag.

ederign: @jesuino I wonder if we should also create this file for download in a temp folder just for download it? I guess jupyter supports there

Ya-shh:
> Thanks for your contribution! The feature seems awesome, but I could not get it to work locally, I had an error and I think it is because the flag `--ContentsManager.allow_hidden=True` is missing on `jupyter-kfp` make target. Would you please check it? Also, please check my comments. Thanks!
> 
> Image

Thanks for reporting this @jesuino ! The initial implementation relied on JupyterLab's Contents API, which doesn't work with hidden directories (.kale/) without that flag. I've since updated the approach , the compiled script content is now returned directly in the compile_notebook RPC response. The download creates a client-side blob from this content, so it no longer depends on the Contents API. This should resolve the error you encountered.

Makefile
Ya-shh: > Will this be required in production to have this working? Notice that some users may want to keep files hidden, hence we must make sure that we document that this feature only works with this flag.

Addressed @jesuino , the --ContentsManager.allow_hidden=True flag is no longer needed. The Makefile remains unchanged. The download now works by passing the compiled script content through the existing RPC response, making it independent of the Contents API and any server configuration. No additional flags are required in production or development.

Ya-shh: > @jesuino I wonder if we should also create this file for download in a temp folder just for download it? I guess jupyter supports there

Good point @ederign ! I addressed this by returning the compiled script content directly in the compile_notebook RPC response (script_content field). The frontend creates a client-side blob from this for download, so it doesn't need to access the hidden .kale/ directory at all , no temp folder or additional server configuration required.

Ya-shh:
hey @jesuino & @ederign ,
I want to mention one thing

-script_content is included in the compile_notebook response for all compile flows (Save, Upload, Run), even though the download button only appears after Compile and Save. The compiled scripts are typically a few KB, and this is a single-user Jupyter session, so the overhead is negligible. Kept it simple to avoid adding a flag parameter to the RPC ,happy to add that optimization if preferred.

labextension/src/widgets/deploys-progress/DeployProgress.tsx
Ya-shh: Thanks ,I removed that

PR #608, opened 2026-02-16
fix: disable deploy buttons on form validation errors and show friendly error messages

### LINKED ISSUE #605: [frontend] Invalid experiment name throws a RPC error
### Environment

*  How did you deploy Kale?

make jupyter-kfp (and kfp running)

main

### Steps to reproduce

Add a new experiment, put a invalid name (i.e. starting with Upper case), click in Run:

Image

### Expected result

We should validate and block the 'Compile and Run' 'Compile and save button' if the 'form' is invalid

### Materials and Reference

---

Ya-shh:
- Disables "Compile and Run/Upload/Save" buttons when pipeline name or experiment name has validation errors.

-  Also wraps raw RPC validation errors with user-friendly messages instead of showing technical error dialogs.

Closes #605


Ya-shh:
@ederign  could you please review this?

Ya-shh:
Fixes-#605

StefanoFioravanzo:
@Ya-shh can you share a short demo recording?

labextension/src/lib/Commands.ts
```diff
@@ -195,18 +195,32 @@ export default class Commands {
...
+        this._kernel,
+        'nb.validate_notebook',
+        validateNotebookArgs,
+      );
+      if (!validateNotebook) {
+        onUpdate({ notebookValidation: false });
+        return false;
+      }
+      onUpdate({ notebookValidation: true });
+      return true;
```
jesuino: Wouldn't be easier to just call the callback and return the `validateNotebook` value?

 ```
onUpdate({ notebookValidation: validateNotebook });
return validateNotebook;
```

File: labextension/src/widgets/LeftPanel.tsx [hunk 2]
```diff
@@ -98,6 +99,7 @@ export const DefaultState: IState = {
   namespace: '',
   kfpUiHost: '',
   defaultBaseImage: '',
+  formErrors: { pipelineName: false, experimentName: false },
```
jesuino: Why is an apparent string being set as boolean? Should it be something like `isExperimentValid`?

jesuino (changes requested):
Thanks for your contribution!

I see that the button is indeed disabled when there's validation problems

Image


The panel still continues very technical. To reproduce this just submit a notebook without step. 

Image

Ya-shh:
Thank you @jesuino as always! 🙌

**Changes summary:**

- `validateMetadata`
  -  now just calls the callback and directly returns `validateNotebook`
  - Added explicit `RPCError` handling in the `catch` block
  - Prioritizes `err_details` → falls back to `err_message` → generic fallback

- `compilePipeline`
  - Replaced `_legacy_executeRpcAndShowRPCError` with `_legacy_executeRpc` + manual `RPCError` catch
  - Applies the same error message priority: `err_details` → `err_message` → fallback
  - Eliminates the double dialog issue 

**Visible user impact:**

- Submitting a notebook with no tagged steps now shows a clear message:
  > The pipeline does not have any steps. Please tag a cell as a pipeline step or set `steps_defaults` in the notebook metadata.





please let me know if anything looks off ?

Ya-shh:
> @Ya-shh can you share a short demo recording?





Yeah for sure @StefanoFioravanzo !


https://github.com/user-attachments/assets/1948d371-cfa3-4696-aed1-f21afb01da19

Ya-shh:
> @Ya-shh thanks for the PR. Some things that we need to change: If you want to fix this issue only on disable deploy buttons, this is the blocker
> 
> * The dropdown toggle button isn't disabled
> * Please don't make NEVER things scroll horizontally.
> 
> If you also want to fix #605
> 
> * error-handling pattern in Commands.ts is almost duplicated. Instead of inline thec atch in every caller, we need to make it generic because there are other 7 other call sites still using _legacy_executeRpcAndShowRPCError. If the plan is to  migrate them all to friendly messages, each one will copy-paste this pattern. Better to centralize it now.
> * Is all type of errors as error instanceof RPCError? Shall we make it generic to blow with a nice message instead of a generic error message?
> 
> And then, the fun begins with React. Try to learn why this happens with React instead of asking LLM, you will learn a few things with controlled vs uncontrolled components, duplicated state, and props-driven side effects. This helps you learn about state transitions :)
> 
> think about whether formErrors should exist at all, or whether LeftPanel should derive button-disabled state directly from the current field values at render time. Should derived state that can be computed be stored?

Thanks for the detailed review @ederign , it was genuinely helpful.

Addressed all the points:

- Dropdown: disabled DropdownButton alongside the main button when the form is invalid

- Horizontal scroll: fixed

- Error handling: centralized into _legacy_executeRpcAndShowRPCError , it covers all BaseError subtypes, re-throws anything unexpected

- Input: removed local state, useEffect, and debounce ,propsValue drives the field directly

- formErrors: dropped from state, validity derived inline at render

Good callout on the React patterns , cleaning up the controlled component and derived state made the data flow noticeably simpler.

Ya-shh:
> LGTM!
> 
> 1) The message looks better. Later we should think about an actual error dialog, not a generic dialog
> 
> Image
> 
> 2) The button is now fully disabled:
> 
> Image
> 

Yes,Thank you @jesuino !

labextension/src/components/Input.tsx
```diff
@@ -166,7 +166,6 @@
 //   );
 // };
 import * as React from 'react';
-import { useDebouncedCallback } from 'use-debounce';
```
jesuino, line 169: I think `use-debounce` isn't used anymore, would you please also remove the dependency from `package.json` and let us know how it goes?

jesuino (changes requested):
I think we can remove the `use-debounce` dependency after your change!

labextension/src/components/Input.tsx
Ya-shh: Yes , done @jesuino !

ederign:
https://github.com/user-attachments/assets/48231fc4-10e2-4dbd-9fbb-788159dcfa7a

jesuino:
@Ya-shh Would you please check the issue mentioned by Eder? Please also merge your branch with main!

Thanks

ederign:
https://github.com/user-attachments/assets/fdda5235-73f0-4176-bddc-a9d7d3a2ca3a

labextension/src/components/Input.tsx
```diff
@@ -215,82 +212,63 @@ export const Input: React.FunctionComponent<IInputProps> = props => {
...
-        updateError(true);
+    setLocalValue(newValue);
+    if (onBeforeUpdate) {
+      if (onBeforeUpdate(newValue)) {
+        setBeforeUpdateError(true);
       } else {
-        updateError(false);
-        debouncedCallback(newValue, inputIndex || 0);
+        setBeforeUpdateError(false);
+        updateValue(newValue, inputIndex || 0);
```
jesuino: This is being called multiple times while we do type in the Kale Step Name, ending with multiple saves while typing and finally race condition between saving the the notebook state.

Image

jesuino:
Hello @Ya-shh 

I can also reproduce the behavior mentioned by Eder and I do believe the reason is because the change you made is leading the notebook be saved every time we change the input, see the video below.

Now bear in mind that this is a behavior that we are working to change on [this PR](https://github.com/kubeflow/kale/pull/627). Meanwhile would you please prevent this behavior and therefore the error seeing by Eder?

THanks! 


https://github.com/user-attachments/assets/3bba13e8-5b22-4d8c-a5fd-9fb1e65d6c85

labextension/src/components/Input.tsx
Ya-shh: yes, I have  resolved it @jesuino

Ya-shh:
> Hello @Ya-shh
> 
> I can also reproduce the behavior mentioned by Eder and I do believe the reason is because the change you made is leading the notebook be saved every time we change the input, see the video below.
> 
> Now bear in mind that this is a behavior that we are working to change on [this PR](https://github.com/kubeflow/kale/pull/627). Meanwhile would you please prevent this behavior and therefore the error seeing by Eder?
> 
> THanks!
> 
>  Screencast.From.2026-03-09.15-41-49.mp4

Yeah I can see that, thanks for the review @jesuino ! I have resolved this issue , please take a look

Ya-shh:
@ederign & @jesuino ,
Thanks for the review!
The issue was that i initially removed the `!error` guard from `handleBlur` globally. This was done to ensure invalid values could reach the parent state, allowing the deploy buttons to be properly disabled. However, it caused all Input components , including cell metadata step names to propagate invalid values to the parent, which triggered unnecessary notebook saves.

I have now added a targeted `updateOnError` prop to the `Input` component (defaults to `false`). Only the pipeline name and experiment name inputs use this prop, so all other inputs behave as they did before.

I also fixed a double-click bug in the cell metadata editor by changing the close button from `onClick` to `onMouseDown`. Do let me know whether it looks good or not?

Just aiming for the best possible user experience :)

examples/base/candies_sharing.ipynb
```diff
@@ -207,7 +207,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.19"
+   "version": "3.12.2"
```
jesuino: Would you please revert this change? This should stop happening after we fix the auto save

labextension/src/lib/TagsUtils.ts
```diff
@@ -193,20 +193,22 @@ export default class TagsUtils {
...
-    for (i = 0; i < notebookPanel.model!.cells.length; i++) {
-      const tags: string[] = CellUtils.getCellMetaData(
-        notebookPanel.content,
-        i,
-        'tags',
-      ) || [];
+    for (let i = 0; i < notebookPanel.model!.cells.length; i++) {
+      const tags: string[] =
+        CellUtils.getCellMetaData(notebookPanel.content, i, 'tags') || [];
+      if (!(tags || []).includes(prevTag)) {
```
jesuino: Is this `tags || []`really necessary? Won't tags be `[]` if it is undefined?

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
```diff
@@ -309,7 +309,13 @@ export class CellMetadataEditor extends React.Component<IProps, IState> {
       currentCellMetadata,
       false,
     ).then(() => {
-      TagsUtils.updateKaleCellsTags(this.props.notebook, oldBlockName, value);
+      TagsUtils.updateKaleCellsTags(
```
jesuino: I wonder why you are saving externally and not passing `true` to this method? Notice we are reviewing the auto save process, so I would not spread `notebook.context.save()` in the code

Ya-shh:
> * Now that the User editing the tag is not auto reflected and only when I click outside it is reflected, is confusing what will happen if I click "Run" when I modified the tag.

Yes @jesuino the behavior is correct, just the visual feedback is delayed until blur.

This is the original behavior of the Input component ,my changes didn't modify it. I only added updateOnError to the pipeline name and experiment name inputs.

I could fix it by making the tag update visually on every keystroke, but the problem is updateCurrentBlockName also calls notebook.context.save() , so we would be saving the notebook on every keystroke

To do it properly, i would need to:

1.Update the visual tag in real time without saving
2.Only save the notebook on blur

this is a separate UX improvement unrelated to my PR but i can open a separate PR for this. Do let me know what do you think about this?

examples/base/candies_sharing.ipynb
Ya-shh: yes reverted

labextension/src/lib/TagsUtils.ts
Ya-shh: Thanks for pointing out this, I have removed those tags

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
Ya-shh: yeah now setKaleCellTags handles saving internally with save: true, and updateKaleCellsTags also defaults to save: true

Ya-shh:
Thanks for pointing out these issues @jesuino . I have made the changes, please take a look

Ya-shh:
@jesuino , I have resolved the merge conflict, please take a look

jesuino:
@Ya-shh Would you please merge again so I can review? Thanks!

Ya-shh:
yeah, done @jesuino

labextension/src/components/ExperimentInput.tsx
```diff
@@ -106,13 +106,13 @@ export const ExperimentInput: React.FunctionComponent<
     return filtered.length === 0 ? '' : filtered[0].name;
   };
 
-  const updateSelected = (selected: string, index?: number) => {
+  const updateSelected = (selected: string, _index?: number) => {
```
jesuino: Is this refactor really needed for this PR?

labextension/src/components/ExperimentInput.tsx
```diff
@@ -106,13 +106,13 @@ export const ExperimentInput: React.FunctionComponent<
...
     const value = selected === NEW_EXPERIMENT.id ? '' : getName(selected);
     const experiment: IExperiment = { id: selected, name: value };
     props.updateValue(experiment);
   };
 
-  const updateValue = (value: string, index?: number) => {
+  const updateValue = (value: string, _index?: number) => {
```
jesuino: Same here, why rename this variable?

labextension/src/components/Input.tsx
```diff
@@ -215,106 +213,94 @@ export const Input: React.FunctionComponent<IInputProps> = props => {
...
     variant = 'outlined',
-    InputProps,
     updateValue,
-    onBeforeUpdate = undefined,
+    onBeforeUpdate,
+    updateOnError = false,
     ...rest
   } = props;
 
+  const [localValue, setLocalValue] = React.useState(String(propsValue));
```
jesuino: Would you please import `useState` and  `useEffect` from `React` and use them to avoid repeating `React` class?

labextension/src/components/Input.tsx
```diff
@@ -215,106 +213,94 @@ export const Input: React.FunctionComponent<IInputProps> = props => {
...
-    } else {
-      const hasError = onBeforeUpdate(newValue);
-      if (hasError) {
-        updateError(true);
+    setLocalValue(newValue);
+    if (updateOnError) {
+      updateValue(newValue, inputIndex || 0);
+    }
+    if (onBeforeUpdate) {
+      if (onBeforeUpdate(newValue)) {
```
jesuino: I would keep the `hasError` variable because it is clear to see why the return of `onBeforeUpdate` is being used in an `if`.

labextension/src/components/Input.tsx
```diff
@@ -215,106 +213,94 @@ export const Input: React.FunctionComponent<IInputProps> = props => {
...
+  };
+
   return (
     <StyledTextField
       {...rest}
       variant={variant}
       className={className}
       error={error}
-      value={value}
+      value={localValue}
```
jesuino: Would you please tell me why a `localValue` is required? I am just trying to understand

labextension/src/lib/Commands.ts
```diff
@@ -272,9 +272,7 @@ export default class Commands {
     );
     if (!compileNotebook) {
       onUpdate({ compiledPath: 'error' });
-      await NotebookUtils.showMessage('Operation Failed', [
-        'Could not compile pipeline.',
-      ]);
+      return null;
```
jesuino: I would avoid returning null because it may lead to `Uncaught TypeError: can't access property "x", object is null` errors if someone tries to use this object without verifying it.

I know this method was here before, but we can take this opportunity to type the return object of method `compilePipeline` and return something more meaningful to the caller?

labextension/src/lib/TagsUtils.ts
```diff
@@ -215,13 +217,16 @@ export default class TagsUtils {
...
         .filter(t => t !== '' && t !== 'prev:');
+      hasChanges = true;
       allPromises.push(
         CellUtils.setCellMetaData(notebookPanel, i, 'tags', newTags, false),
       );
     }
-    Promise.all(allPromises).then(() => {
-      notebookPanel.context.save();
-    });
+    if (hasChanges && save) {
```
jesuino: This part will likely conflict with my PR which should be merged before yours since mine is targeting a blocker:  https://github.com/kubeflow/kale/pull/707

I would wait my PR to be merged and later you can rebase yours and check how it behaves after my changes.

labextension/src/components/ExperimentInput.tsx
Ya-shh: The index param is unused but required by the Select/Input callback signature, so I prefixed it with _ following common convention. But since the project already allows unused args in ESLint, it's unnecessary. Thanks for pointing out , I  removed it

labextension/src/components/Input.tsx
Ya-shh: `localValue` replaces the old `useDebouncedCallback` approach. It holds the in progress input so the text field updates instantly while typing, and `updateValue` is only called on blur ,which avoids parent state updates on every keystroke. This also let us remove the `use-debounce` dependency.

labextension/src/lib/Commands.ts
Ya-shh: Thanks @jesuino !, I have replaced the null return with a typed result object that returns { success: false } on failure. The caller now checks result.success instead of a falsy check.

labextension/src/lib/TagsUtils.ts
Ya-shh: Sounds good, I'll wait for your PR to be merged and then rebase on top of main

labextension/src/components/Input.tsx
Ya-shh: Yeah , I fixed this

labextension/src/components/Input.tsx
Ya-shh: Yeah , I fixed this

Ya-shh:
Thanks for the detailed feedback @jesuino !, please take a look at my replies to your comments , I have made changes accordingly

jesuino:
Hello @Ya-shh I see that there are conflicts with the base branch, would you please fix it? I will review it right after the conflicts are solved and it is merged with main

Thanks!

Ya-shh:
yes @jesuino , I have fixed the conflicts, please take a look

labextension/src/components/Input.tsx
jesuino: Hello @Ya-shh is it still needed after I modified the way that the UI is updated? Please run Kale on Main and you will see:

Now Kale will listen to metadata change and update the UI accordingly.

I was hoping that this change would save us from doing the `onBlur` and the value replication. For me waiting for the blur to see the UI change is very weird, personally I don't like it and I think it is leading to a worse UX.

Thanks!

Ya-shh:
Thanks @jesuino ,you are right, localValue and onBlur are no longer needed after your metadata listening changes. I have removed the value replication and the input now updates directly on change. Also removed the updateOnError prop since it's no longer relevant

labextension/src/lib/TagsUtils.ts
```diff
@@ -193,19 +193,19 @@ export default class TagsUtils {
     oldBlockName: string,
     newBlockName: string,
   ) {
-    let i: number;
+    const prevTag = 'prev:' + oldBlockName;
```
jesuino: Why are changes on this class related to this? Could be this that is leading to problems with the step names?

Ya-shh:
@jesuino , yes TagsUtils.ts should not have been modified in this PR. The changes there (prevTag variable, continue optimization, let i refactor) were leftover from an earlier iteration where I added a save parameter to updateKaleCellsTags. After your auto save removal PR was merged, I resolved the merge conflict but didn't fully revert the surrounding refactors ,I only reverted the save logic. The continue statement changed the behavior, the upstream processes all cells and rewrites their tags, while my version skipped cells without a prevTag, which could leave stale metadata. I have now reverted TagsUtils.ts to exactly match main
 
 I have  fixed that issue, please take a look

Ya-shh:
hey @jesuino , I noticed the duplicate step name issue  exists on main. The backend handles it silently by merging the code of duplicate named cells into a single step (in `nbprocessor.py` file),  it doesn't cause a compilation or deployment error , but the user loses the separate step they intended. Would it make sense to open a  PR that detects duplicate step names and disables the deploy buttons until the user fixes them?

PR #614, opened 2026-02-18
Kale throws exception when all cells are deleted

### LINKED ISSUE #613: [frontend] Kale throws exception when all cells are deleted
### Environment

*  How did you deploy Kale?

N/a

Latest

### Steps to reproduce

* Create a new Notebook
* Enable Kale
* Delete the only empty cell
* Kale panel is now empty and there's an error on console

Image

### Expected result

Kale should just ignore that the cell was deleted and not break.

### Materials and Reference

---

jesuino:
closes #613 
Adding a check for the cell seems to be enough to fix #613


ada333:
I tried the change and it works as expected.

PR #615, opened 2026-02-18
Edit button goes below the Cell when adding new cell

### LINKED ISSUE #610: [frontend] Edit button goes below the Cell when adding new cell
When you click on "Inset Cell Below" button, the edit button (little pencil) is place below the generated cell, which is wrong, because the edit button is placed on top of the cell.

Image

Image

jesuino:
closes #610 

A possible solution for this with my good old friend Mutation Observer. There are other possible solution, such as change the way that we register the callback to add the inline editor to the cell.


ada333:
I tried these changes and it works as expected.

PR #620, opened 2026-02-19
ISSUE-559: Add cell types in ReadMe

ada333:
This PR adds clear information about cell types in ReadMe mainly for first-time user. I think it is important that we mention somewhere that imports work only in imports cell.

### DIFF
--- a/.github/workflows/build_labextension.yml
+++ b/.github/workflows/build_labextension.yml
@@ -19,6 +19,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+ 
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
 
       - name: Install uv
         uses: astral-sh/setup-uv@v4
@@ -43,6 +48,11 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
 
       - name: Install uv
         uses: astral-sh/setup-uv@v4
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@ to the project.
 
 - **Python 3.12+**
 - **uv** - Installed automatically by `make`, or manually: `curl -LsSf https://astral.sh/uv/install.sh | sh`
-- **Node.js 18+** and **jlpm** (for frontend development only - jlpm comes with JupyterLab)
+- **Node.js 22+** and **jlpm** (for frontend development only - jlpm comes with JupyterLab)
 - **Kubernetes cluster** (optional, for KFP testing only) - [minikube](https://minikube.sigs.k8s.io/) or [kind](https://kind.sigs.k8s.io/)
 
 ## Quick Start
--- a/backend/kale/pipeline.py
+++ b/backend/kale/pipeline.py
@@ -101,6 +101,7 @@ class PipelineConfig(Config):
     experiment_name = Field(type=str, required=True)
     pipeline_description = Field(type=str, default="")
     base_image = Field(type=str, default="")
+    enable_caching = Field(type=bool, default=True)
     volumes = Field(type=list, items_config_type=VolumeConfig, default=[])
     katib_run = Field(type=bool, default=False)
     katib_metadata = Field(type=KatibConfig)
--- a/backend/kale/processors/nbprocessor.py
+++ b/backend/kale/processors/nbprocessor.py
@@ -53,6 +53,11 @@ LIMITS_TAG = r"^limit:([_a-z-\.\/]+):([_a-zA-Z0-9\.]+)$"
 # Image tag for per-step Base image selection
 # E.g.: image:python:3.11-slim
 IMAGE_TAG = r"^image:(.+)$"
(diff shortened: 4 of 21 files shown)


README.md
```diff
@@ -103,6 +103,38 @@ jupyter lab
...
+## Cell Types
+
+Kale uses special cell types (tags) to organize your notebook into pipeline components. You can assign these types to cells using the Kale JupyterLab extension or by adding tags directly in the notebook metadata.
+
+### Cell Types Reference
+
+| Cell Type | Status | Description |
+|-----------|--------|-------------|
+| **Imports** | ✅ Works | Use for all import statements. **All imports must be placed in cells tagged as `imports`.** Importing libraries (pandas, tensorflow, etc.) in other cell types will cause pipeline execution errors. |
+| **Functions** | ✅ Works | Use for function and class definitions only. **Do not include** logic, print statements, or import statements in these cells. Only pure definitions should be here. |
```
ederign: Function cells can contain logic inside function bodies right? I think what you mean is that function cells should only contain definitions (function/class defs), not top-level
  executable statements

README.md
```diff
@@ -103,6 +103,38 @@ jupyter lab
...
+| **Pipeline Parameters** | ✅ Works | Define variables that will become pipeline parameters. |
+| **Pipeline Metrics** | ⚠️ Known Issue | Intended for defining pipeline metrics, but currently not working. |
+| **Step** | ✅ Works | Regular pipeline steps with custom names. This is the default cell type for your data processing and ML logic. Each step can have dependencies on other steps. Steps can also define their own image and GPU requirements. |
+| **Skip Cell** | ✅ Works | Cells marked as skip will be excluded from the pipeline. Useful for exploratory code or debugging that shouldn't be part of the production pipeline. |
+
+### Important Guidelines
+
+> [!WARNING]
+> **Import statements must be in `imports` cells only!**
```
ederign: A more accurate statement: imports outside imports cells won't be detected for automatic dependency installation,
  which causes ImportError at runtime if the package isn't pre-installed in the container image.

If the package is on the container image works isn't?

ederign (changes requested):
Just some minor comments.

README.md
ada333: good point, if the package is on the container image it will work. I will add it to Readme.

README.md
```diff
@@ -103,6 +103,37 @@ jupyter lab
...
+
+Kale uses special cell types (tags) to organize your notebook into pipeline components. You can assign these types to cells using the Kale JupyterLab extension or by adding tags directly in the notebook metadata.
+
+### Cell Types Reference
+
+| Cell Type | Status | Description |
+|-----------|--------|-------------|
+| **Imports** | ✅ Works | The code in this cell will be pre-pended to every step of the pipeline. Used for all import statements. **All imports must be placed in cells tagged as `imports`.** Importing libraries (pandas, tensorflow, etc.) in other cell types will cause pipeline execution errors. |
+| **Functions** | ✅ Works |The code in this cell will be pre-pended to every step of the pipeline, after **imports**. Used for function and class definitions only.  **Do not include** top-level executable statements |
+| **Pipeline Parameters** | ✅ Works | Define variables that will become pipeline parameters. |
```
hmtosi: Some testing on the `Pipeline Parameters` label showed that if more than on e `Pipeline Parameters` cell exists, and a parameter is defined in each cell, only the final value will be taken. For example,  `parameter cell 1: num_cats = 3` and `parameter cell 2: num_cats = 5` will result in `num_cats = 5` for the whole pipeline.

Please update this section to say that if the same parameter is defined in mulitple `Pipeline Parameter` cells, only the value in the last cell will be used.

README.md
```diff
@@ -103,6 +103,37 @@ jupyter lab
...
+| **Pipeline Metrics** | ⚠️ Known Issue | Intended for defining pipeline metrics, but currently not working. |
+| **Step** | ✅ Works | Regular pipeline steps with custom names. This is the default cell type for your data processing and ML logic. Each step can have dependencies on other steps. Steps can also define their own image and GPU requirements. |
+| **Skip Cell** | ✅ Works | Cells marked as skip will be excluded from the pipeline. Useful for exploratory code or debugging that shouldn't be part of the production pipeline. |
+
+### Important Guidelines
+
+> [!WARNING]
+> **Imports outside imports cells won't be detected for automatic dependency installation, which causes ImportError at runtime if the package isn't pre-installed in the container image.**
```
hmtosi: Maybe write this as "Imports outside `Imports` cells..." to be slightly clearer

README.md
```diff
@@ -103,6 +103,37 @@ jupyter lab
...
+
+
+**Best Practices:**
+- Place all imports at the beginning of your notebook in cells tagged as `imports`
+- Keep function definitions pure - no side effects, prints, or imports
```
hmtosi: what do you mean by side effects in this context?

ada333: I meant no changing of global variables / parameters.

ederign:
@hmtosi would mind taking a second look here?

ederign (approved):

README.md
hmtosi: resolved

README.md
hmtosi: resolved

README.md
hmtosi: thank you for explaining, lgtm

Issue #529, opened 2025-11-28
[feature] Allow users to disable/enable cache for pipeline components or for the whole pipeline


jesuino:
### Feature Area

/area frontend
/area backend

### What feature would you like to see?

Allow users to enable/disable cache for individual steps and for the whole pipeline.

### What is the use case or pain point?

One might need to run a pipeline from a fresh start after correcting some dependency. In these cases we need to disable cache, other kfp will use the last run from cache.

### Is there a workaround currently?

Modify the DSL to disable caching according to the documentation:

https://www.kubeflow.org/docs/components/pipelines/user-guides/core-functions/caching/

---

Love this idea? Give it a 👍.

PR #631, opened 2026-02-24
feat: Allow users to disable/enable cache for pipeline components or for the whole pipeline

jesuino:
In this PR we add support to Pipeline Cache config as described in Issue #529  - the code change was created along with AI assistance (Claude Code and local models), but not the description below.

How to test this:

* Take as example [this simple notebook](https://github.com/jesuino/kale_examples/blob/main/cache_test.ipynb), used for testing;
image

* First, bear in mind that Cache is set true by default, which means that if run a Pipeline twice, the second run will have all nodes cached. You can try this with the workflow above and you will see that first all nodes run and the second run will have the little cloud icon; all subsequent runs will be retrieved from the cache
image

image

Notice also that the second run is much faster then the first, due the cache:

image

* Let's start playing with cache. First disable cache for the whole Pipeline and run the pipeline again, now you will see that all nodes are fresh, running since it was the first time (not recovered from cache)

image

* Now let's turn on cache ONLY for the `mult_a` step, and if you run the pipeline you will see that only this node will run fresh, the other will be recovered from cache

image

image

In addition I also removed the need to print the base image if it is using the default image. It will be printed only if the user set something.

### DIFF
--- a/backend/kale/processors/nbprocessor.py
+++ b/backend/kale/processors/nbprocessor.py
@@ -55,7 +55,9 @@ LIMITS_TAG = r"^limit:([_a-z-\.\/]+):([_a-zA-Z0-9\.]+)$"
 IMAGE_TAG = r"^image:(.+)$"
 # Cache tag for per-step caching control
 # E.g.: cache:enabled or cache:disabled
-CACHE_TAG = r"^cache:(enabled|disabled)$"
+CACHE_ENABLED = "enabled"
+CACHE_DISABLED = "disabled"
+CACHE_TAG = rf"^cache:({CACHE_ENABLED}|{CACHE_DISABLED})$"
 
 _TAGS_LANGUAGE = [
     SKIP_TAG,
@@ -174,7 +176,7 @@ class NotebookConfig(PipelineConfig):
             if conf_type == "cache":
                 # Cache value is 'enabled' or 'disabled'
                 cache_value = parts.pop(0)
-                result["enable_caching"] = cache_value == "enabled"
+                result["enable_caching"] = cache_value == CACHE_ENABLED
         return result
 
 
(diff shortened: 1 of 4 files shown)


backend/kale/processors/nbprocessor.py
```diff
@@ -53,6 +53,9 @@
 # Image tag for per-step Base image selection
 # E.g.: image:python:3.11-slim
 IMAGE_TAG = r"^image:(.+)$"
+# Cache tag for per-step caching control
+# E.g.: cache:enabled or cache:disabled
+CACHE_TAG = r"^cache:(enabled|disabled)$"
```
ada333: I would make the values into constants for better code - at least the enabled one since we are asking for its value in few if statements

labextension/src/lib/TagsUtils.ts
```diff
@@ -118,11 +120,20 @@ export default class TagsUtils {
         baseImage = imageTag.substring(IMAGE_TAG.length);
       }
 
+      // Parse cache tag
+      let enableCaching: boolean | undefined;
+      const cacheTag = tags.find(v => v.startsWith(CACHE_TAG));
+      if (cacheTag) {
+        const cacheValue = cacheTag.substring(CACHE_TAG.length);
+        enableCaching = cacheValue === 'enabled' ? true : false;
```
ada333: here again I would use some constant

ederign:
I didn't run it locally, but I'll as soon as you address Adam's comments! btw, this is a great feature!

jesuino:
@ada333 thanks for your suggestions, I did modify the code to use constants

@ederign I also like this feature! The test you mention is not possible at the moment because KFP default behavior is to cache, so when we enable cache we just don't explicitly set the task to have no cache.

However, I notice that the test was lacking the full disable of cache, so I modified the Iris example to disable cache and make sure that all nodes have disabled cache in that case.

I am still checking how I could make a better look for the button. With help from Gemini I had this, but I think it is not there yet

image

ederign:
@jesuino lgtm! We can always improve the design of this page. I'm planning to take a look and improve it someday.

Let's wait @ada333 review

ada333 (approved):
@jesuino I think the design looks good for now! I would merge this PR, we can always change it additionally.

PR #639, opened 2026-02-26
ISUS-616: Fix deleting cells

ada333:
This PR fixes errors when deleting steps that are dependencies of some other steps.
There were 2 problems - in func removeOldDependencies in TagsUtils.ts was parameter thats value was always undefined and the function never deleted the step from dependecies - backend error when compiling notebook.

In UI the when deleting steps the dependencies were not filtered properly and it left the deleted step in MultiSelect of dependencies and the color dot.


labextension/src/lib/TagsUtils.ts
```diff
@@ -251,47 +250,36 @@ export default class TagsUtils {
...
-        if (Array.isArray(cellTags)) {
-          const newTags = cellTags.filter(e => e !== removedDependency);
-          cell.metadata['tags'] =  newTags;
+
+    const allBlocks = this.getAllBlocks(notebook.content);
+    const allBlocksSet = new Set(allBlocks);
+
+    for (let index = 0; index < cells.length; index++) {
+      const cell = cells.get(index);
+      const cellTags = cell?.metadata['tags'];
```
jesuino: Wouldn't it be better to use the method `CellUtils.getCellMetaData` ? This way we won't have to check if the tags are an array and we will not be required to check if `typeof tag === 'string'`

labextension/src/widgets/cell-metadata/InlineMetadata.tsx
```diff
@@ -251,7 +255,11 @@ export class InlineMetadata extends React.Component<IProps, IState> {
    * block
    */
   updateDependencies() {
-    const dependencies = this.props.stepDependencies.map((name, i) => {
+    const validDependencies = this.props.stepDependencies.filter(dep =>
```
jesuino: Is this double check really needed? I didn't test the code with it, but I do expect that once you remove orphaned dependencies using `removeOldDependencies` the dependencies list will be up-to-date removing the need to check if a dependency is valid or not.

If in your tests you had to duplicate this check, then I think that something may be wrong with how we are handling this React component state.

jesuino (changes requested):
Hello @ada333 great work, it is working well!

I have a code change suggestion and a question. I do believe we should not double check the dependencies again when rendering the metadata. Something could be wrong if we are required to do that. If you found that the UI is getting somehow out of sync, I would say that we should analyze it in another issue.

Thanks!

labextension/src/lib/TagsUtils.ts
jesuino: More possible improvements:

* Use getKaleCellTags  and the `previousBlocks` property, which will bring it already handled, so no need to repeat the logic. 
* Instead getting the tags all the time, you could modify `getKaleCellTags` to include the tags, which will improve performance and code readability. The reason is because getKaleCellTags already calls `CellUtils.getCellMetaData(notebook, index, 'tags') `, so why not include it in the  `IKaleCellTags` object and reuse it instead increase the [algorithm complexity](https://en.wikipedia.org/wiki/Computational_complexity_theory)?

ada333:
@jesuino thanks for review, and you are right, the double check was not needed. I updated the PR with your suggestions.

labextension/src/lib/TagsUtils.ts
```diff
@@ -251,47 +250,40 @@ export default class TagsUtils {
...
-        const cellTags = cell?.metadata['tags'];
-        if (Array.isArray(cellTags)) {
-          const newTags = cellTags.filter(e => e !== removedDependency);
-          cell.metadata['tags'] =  newTags;
+
+    const allBlocks = this.getAllBlocks(notebook.content);
+    const allBlocksSet = new Set(allBlocks);
+
+    for (let index = 0; index < cells.length; index++) {
+      const kaleTags = this.getKaleCellTags(notebook.content, index);
```
jesuino, line 281: I do believe only one check for the object should be enough, then you can remove all the other checks:

```
    for (let index = 0; index < cells.length; index++) {
      const kaleTags = this.getKaleCellTags(notebook.content, index);
      if (!kaleTags) {
        continue;
      }
      const deletedDeps = kaleTags.prevBlockNames.filter(
        dep => !allBlocksSet.has(dep)
      );

      const newPrevBlockNames = kaleTags.prevBlockNames.filter(
        dep => !deletedDeps.includes(dep)
      );

///....

```

labextension/src/lib/TagsUtils.ts
```diff
@@ -251,47 +250,40 @@ export default class TagsUtils {
...
+      if (kaleTags && kaleTags.prevBlockNames.length > 0) {
+        const deletedDeps = kaleTags.prevBlockNames.filter(
```
jesuino: Do we really need this intermediate variable?, wouldn't it be okay if we just verify if a dependency is existing in the current block set?

```
const newPrevBlockNames = kaleTags.prevBlockNames.filter(
        dep => allBlocksSet.has(dep)
      );
```

labextension/src/lib/TagsUtils.ts
```diff
@@ -251,47 +250,40 @@ export default class TagsUtils {
...
+          dep => !allBlocksSet.has(dep)
+        );
+
+        if (deletedDeps.length > 0) {
+          const newPrevBlockNames = kaleTags.prevBlockNames.filter(
+            dep => !deletedDeps.includes(dep)
+          );
+
+          const updatedMetadata = {
```
jesuino: I think you should be able to use the [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) here:

```
      const updatedMetadata = {
        ...kaleTags,
        prevBlockNames: newPrevBlockNames,
      };
```

ada333:
@jesuino I absolutely agree, definitely better code quality, thank you for good review. The changes are there.

Issue #641, opened 2026-02-27
[frontend] Changes to the Pipeline Metadata in Left Pane are ignored


jesuino:
### Steps to reproduce

* Open a notebook
* Modify its information on the Left Pane
* Notice that the Pipeline metadata remains untouched. Reopening the notebook and the old value will still be there, ignoring the change in UI

<img width="1119" height="334" alt="Image" src="https://github.com/user-attachments/assets/625ae952-2ed9-4d77-aef6-15ff289342ef" />

### Expected result

Notebook should be dirty after modifying this and users should be able to save the changes.

Workaround is to modify the notebook metadata manually

ai-naymul:
Hi @jesuino 


i have created a fix at #642 pls check that out

ederign:
closed as done!

PR #642, opened 2026-02-27
fix(frontend): persist pipeline metadata changes to notebook file

ai-naymul:
addressing: #641

 Root cause: In JupyterLab 4, notebookPanel.model.metadata returns a read-only copy. The code was doing:                                                                                     
  // This writes to a throwaway copy - changes go nowhere                                                                                                                                     
  metadata[key] = value;

                                                                                                                                                                                             
  Fixes done: Updated getMetaData() and setMetaData() in NotebookUtils.tsx to use JupyterLab 4's proper API:                                                                                         
  // This actually persists and marks notebook as dirty                                                                                                                                       
  notebookPanel.model.setMetadata(key, value);                                                                                                                                                
  notebookPanel.model.getMetadata(key);


ai-naymul:
@jesuino @ederign

labextension/src/lib/NotebookUtils.tsx
```diff
@@ -259,13 +259,17 @@ export default class NotebookUtilities {
...
-      const metadata = notebookPanel.model.metadata as any;
-      if (typeof metadata.has === 'function' && metadata.has(key)) {
-        return metadata.get(key);
+    if (notebookPanel.model) {
+      // JupyterLab 4+: use model.getMetadata() which is the proper API.
+      // model.metadata returns a read-only copy that won't reflect changes.
+      if (typeof notebookPanel.model.getMetadata === 'function') {
+        return notebookPanel.model.getMetadata(key) || null;
+      }
+      // Fallback for older JupyterLab versions
```
jesuino: Which Jupyter Lab will use this feedback? If this is Jupyter 3.x I would say we don't need to keep this Fallback because there are other places where we removed support for older Jupyter versions, so this would fail anyway.

labextension/src/lib/NotebookUtils.tsx
```diff
@@ -292,19 +296,23 @@ export default class NotebookUtilities {
...
-    const metadata = notebookPanel.model.metadata as any;
+    // JupyterLab 4+: use model.setMetadata() which is the proper API.
+    // model.metadata returns a read-only copy, so direct assignment
+    // on it has no effect.
     let oldVal: any;
-
-    if (typeof metadata.set === 'function') {
-      oldVal = metadata.set(key, value);
+    if (typeof notebookPanel.model.setMetadata === 'function') {
+      oldVal = notebookPanel.model.getMetadata(key);
```
jesuino: I don't think this line of code is necessary

labextension/src/lib/NotebookUtils.tsx
```diff
@@ -292,19 +296,23 @@ export default class NotebookUtilities {
...
+      notebookPanel.model.setMetadata(key, value);
     } else {
-      // Fallback for different metadata implementations
-      oldVal = (metadata as any)[key];
-      (metadata as any)[key] = value;
+      // Fallback for older JupyterLab versions
```
jesuino: Same here. The NotebookModel API was introduced in Jupyter Lab 4.x which I do believe is our target to support. I would remove fallback code

https://jupyterlab.readthedocs.io/en/4.0.x/api/classes/notebook.NotebookModel-1.html

labextension/src/lib/NotebookUtils.tsx
```diff
@@ -292,19 +296,23 @@ export default class NotebookUtilities {
...
+      const metadata = notebookPanel.model.metadata as any;
+      oldVal = metadata[key];
+      metadata[key] = value;
+      notebookPanel.model.dirty = true;
```
jesuino: Notice that this code is not invoked with Jupyter 4.x because it won't pass on the else clause, so it is not getting to a dirty state.

jesuino (changes requested):
Hello @ai-naymul 

Thanks for your contribution!

I am suggesting a few changes, in summary we don't need to keep fallback for Jupyter 4.x, so feel free to remove that code. Also I noticed that the dirty status is not working, perhaps we need to wait for [this](https://github.com/kubeflow/kale/pull/627/) to be merged to test this behavior

ai-naymul:
Here is whats done in the new commit;

  1. Removed all JupyterLab 3.x fallback code (typeof checks, else branches with manual metadata access)
  2. Now directly uses model.getMetadata() / model.setMetadata() since Kale targets JupyterLab 4.x
  3. No manual "dirty = true" needed... setMetadata() handles it automatically through the YJS shared model event chain
  
  @jesuino  ready for another look when you get a chance.

ederign:
@ai-naymul, while reviewing this PR, I took a careful look and found that the problem is much more deep: https://github.com/kubeflow/kale/issues/648

So we need to define our long-term strategy with that and patches carefully those cases until we do a proper solution. @ai-naymul I'll look at your PR as soon as possible.

ai-naymul:
> @ai-naymul, while reviewing this PR, I took a careful look and found that the problem is much more deep: #648
> 
> So we need to define our long-term strategy with that and patches carefully those cases until we do a proper solution. @ai-naymul I'll look at your PR as soon as possible.

 @ederign Thanks for the detailed write up on #648, that makes a lot of sense. The hooks based approach would eliminate this entire class of bugs.
                                                                                                                                                                                              
 I understand #648 is on hold until after the 2.0 release. In the meantime, are there any open issues I can pick up?

ai-naymul:
> Great catch, thanks!
> 
> Would you please just remove the comments? We are in a process of getting rid of Jupyter 3.x API, so the comments are not really needed.

 @jesuino 
 
 Removed the comments as requested. Ready for another look.

jesuino:
@ai-naymul Thanks again - would you please update your branch so we can merge this?

Issue #501, opened 2025-10-28
[bug] security violation of PSS baseline/restricted


juliusvonkohout:
### Environment

You can set https://github.com/kubeflow/manifests/blob/dfe02ca826fe49d0411d35ca5c20452f944c1ec7/common/kubeflow-namespace/base/namespace.yaml#L8 on the namespace where you want to enforce it. You can also configure only warnings etc. Here is the general documentation https://kubernetes.io/docs/concepts/security/pod-security-standards/ and a typical example is https://github.com/kubeflow/manifests/blob/dfe02ca826fe49d0411d35ca5c20452f944c1ec7/applications/pipeline/upstream/base/pipeline/ml-pipeline-ui-deployment.yaml#L19.

Here is an example for what must be set on the pods created by Kale:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: ml-pipeline-ui
spec:
  template:
    metadata:
      labels:
        app: ml-pipeline-ui
      annotations:
        cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
    spec:
      securityContext:
        seccompProfile:
          type: RuntimeDefault
      - image: ghcr.io/kubeflow/kfp-frontend:dummy
        name: ml-pipeline-ui
        securityContext:
          allowPrivilegeEscalation: false
          runAsNonRoot: true
          runAsUser: 1000
          runAsGroup: 0
          capabilities:
            drop:
            - ALL```

An LLM can also generate such a tight security context for you. You just have to set it on the generated pipeline containeroperations.


### Steps to reproduce

Run a Kale pipeline and check the securitycontext and failures on a secure namespace.

### Expected result

PSS baseline/restrcted compatibility

### Materials and reference

<!-- Help us debug this issue by providing resources such as: sample code, background context, or links to references. -->

### Labels
<!-- Please include labels below by uncommenting them to help us better triage issues -->

<!-- /area frontend -->
<!-- /area backend -->
<!-- /area testing -->
<!-- /area samples -->

---

<!-- Don't delete message below to encourage users to support your issue! -->
Impacted by this bug? Give it a 👍.

juliusvonkohout:
@StefanoFioravanzo

juliusvonkohout:
See also https://github.com/kubeflow/pipelines/pull/12577 as the KFP first step.

hmtosi:
@ederign please assign this to me, I am working on it

jesuino:
[...]

PR #653, opened 2026-03-04
ISSUE-652: Fix permission error by setting marshal path outside the root

ada333:
This PR just changes the path for marshal path to be outside the root dir so when Kale is run in Docker the "jovyan" user does not get Permission Error.


ada333:
#501 I want for this issue to be resolved first because it should be related.

ada333:
merged into -> #656

PR #656, opened 2026-03-04
Issue 501: PSS baseline - set security context

### LINKED ISSUE #501: [bug] security violation of PSS baseline/restricted

hmtosi:
Closes #501 

This PR implements the missing security logic in the pipeline template so that it is assigned to each pipeline step.


hmtosi:
@jesuino @ederign @ada333 this is ready to be reviewed/tested.

This PR makes the following changes:
- adds the `kfp[kubernetes]>=2.16.0` dependency which is required to add the `securityContext`
- sets the `securityContext` in `pipeline_template.jinja2` as requested by Julius 
- adjusts the default marshal directory to `/tmp/marshal` instead of `/marshal` so that pipelines run successfully 
      - this includes the changes from #653 
- updates the tests to reflect these changes

I believe Julius made three requests, and we have covered one of them, which is setting the security context. We have NOT...
- enabled PSS restricted in our namespace for tests
- set PSS restricted on the namespace we are using 
     - `kubectl label --overwrite namespace <namespace-where-pipelines-run> pod-security.kubernetes.io/enforce=restricted`

Please advise on whether those should be done in this PR as well. I am not familiar with how to do those.

backend/kale/templates/pipeline_template.jinja2
```diff
@@ -30,6 +31,14 @@ def auto_generated_pipeline(
     {%- endif %}
     )
 
+    security_context.set_security_context(
+        task={{ step.name }}_task,
+        run_as_user=65534,
+        run_as_group=0,
+        run_as_non_root=True
+    )
+    {{ step.name }}_task.set_env_variable(name="HOME", value="/tmp")
```
ada333, line 40: NOTE: I think that this is here is necessary because at the start of every step pip installs all the `packages_to_install` to somewhere like `$HOME/.cache/pip` , for the user with some number (here run_as_user=65534) HOME=nonexistent and the pip install fails, thats why we have to specify the `tmp` folder

ada333 (approved):
@hmtosi LGTM! Code looks good and the changes work as expected.

- For the first question - I think the best place to add restricted mode would be our e2e test - right when setting up the KinD cluster. I can add it there since I have an open issue with it anyway.
- Second question - Everyone can set it on their cluster individually - we could add it to ReadMe as recommendation. Other solution could be some general cluster/namespace setting in Makefile? -> some Make command that would set the restricted mode or other security flags in namespace user is running kfp...

backend/kale/templates/pipeline_template.jinja2
hmtosi: Thank you for clarifying that!

jesuino (approved):
I did test this on my Local cluster and it is working fine.

I don't have too much to say about the security side of it, but [reading this](https://kubernetes.io/docs/concepts/security/pod-security-standards/) I would say that Kale is doing its part.

Thanks!

ederign (approved):

ederign:
rebasing because this PR got stucked...

PR #665, opened 2026-03-06
Update Metrics to be compatible with KFP v2 APIs

### LINKED ISSUE #525: [backend] Investigate Pipeline Metrics generation
### Environment

*  How did you deploy Kale?

Built from main 

*  Kale version: 

### Steps to reproduce

* Create a notebook with a pipeline which at least one cell that contains the `pipeline-metrics` label
* Upload it to KFP v2, run the pipeline and I can't find the metrics anywhere

### Expected result

* Metrics should be visible after the Pipeline Run

It seems that the Kale code to handle metrics is not up to date with latest Kfp v2 metrics APIs: https://github.com/kubeflow-kale/kale/blob/main/backend/kale/common/kfputils.py#L287

### Materials and Reference

KFP v2 Metrics docs: https://www.kubeflow.org/docs/components/pipelines/legacy-v1/sdk/output-viewer/#v2-sdk-use-sdk-visualization-apis
KFP Metrics example: https://github.com/kubeflow/pipelines/blob/kfp-kubernetes-2.14.4/samples/test/metrics_visualization_v2.py

---

jesuino:
closes #525 

In this Pull Request we update Kale Metrics to be displayed in KFP UI. The previous way of handling metrics is not compatible with latest KFP.

Notice also that it only supports [scalar](https://www.kubeflow.org/docs/components/pipelines/legacy-v1/sdk/output-viewer/#scalar-metrics) metrics, because it is what was supported previously. However, in a near future we way want to support other metrics types. Also, bear in mind that a call to `print(${metric})` is required so Kale can translate it to KFP metric.

In practice this:

image

Should be translated to this in KFP UI:

image

NOTE: There's an additional field that I didn't add here, but I am checking with KFP if the field `store_session_info` can be considered as a bug, because I can reproduce this behavior without Kale

image

### DIFF
--- a/backend/kale/common/kfputils.py
+++ b/backend/kale/common/kfputils.py
@@ -283,8 +283,11 @@ def generate_mlpipeline_metrics(metrics):
         metrics (dict): a dictionary where the key is the metric name and the
             value is its value.
     """
-    with open(KFP_UI_METRICS_FILE_PATH, "w", encoding="utf-8") as _kale_mf:
-        json.dump(metrics, _kale_mf)
+    try:
+        with open(KFP_UI_METRICS_FILE_PATH, "w", encoding="utf-8") as _kale_mf:
+            json.dump(metrics, _kale_mf)
+    except (OSError, TypeError, PermissionError) as e:
+        log.error("Not able to create metrics file, an unexpected error happened: %s", e)
 
 
 def load_mlpipeline_metrics(output):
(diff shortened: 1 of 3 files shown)


backend/kale/common/kfputils.py
```diff
@@ -283,36 +283,20 @@ def generate_mlpipeline_metrics(metrics):
...
-                )
-                continue
-        metadata.append(
-            {
-                "name": name,
-                "numberValue": value,
-                "format": "RAW",
-            }
-        )
+    with open(KFP_UI_METRICS_FILE_PATH, "w", encoding="utf-8") as _kale_mf:
```
ada333: I would add some check here - either try/catch or if os.path.exists

jesuino: Because I am opening it with "w", then it will [create the file if it doesn't exist](https://www.w3schools.com/python/python_file_handling.asp), but I did add error handling for other reasons, thanks for the suggestion!

jesuino:
I did update the candies sharing example to include metrics!

image

PR #670, opened 2026-03-08
Merge backend and labextension into single kubeflow-kale package

ederign:
## Summary

This PR closes https://github.com/kubeflow/kale/issues/671 

Right now Kale ships as two separate wheels: `kubeflow-kale` (backend) and `kubeflow-kale-labextension` (JupyterLab extension). This made the install story confusing — users had to install both packages, keep versions in sync manually, and the workspace setup added complexity to our build and release process.

This PR merges both into a single `kubeflow-kale` package with an optional `[jupyter]` extra:

```bash
pip install kubeflow-kale              # lean backend: CLI, compiler, marshal
pip install kubeflow-kale[jupyter]     # backend + JupyterLab extension
```

This is the standard pattern used by JupyterLab extensions like [jupyterlab-git](https://github.com/jupyterlab/jupyterlab-git), [jupyterlab-lsp](https://github.com/jupyter-lsp/jupyterlab-lsp), and [elyra](https://github.com/elyra-ai/elyra). One package, one version, one install command. The extension is bundled as a prebuilt labextension inside the wheel using `hatch-jupyter-builder`.

This also unblocks the release workflow — with two packages we'd need coordinated releases, version pinning between them, and two PyPI uploads. With one package it's just a single `uv build` and a single publish.

### Package structure

**Before:**
```
backend/pyproject.toml          → kubeflow-kale (setuptools + setuptools_scm)
labextension/pyproject.toml     → kubeflow-kale-labextension (hatchling)
pyproject.toml                  → uv workspace (virtual, not published)
```

**After:**
```
pyproject.toml                  → kubeflow-kale (hatchling + hatch-jupyter-builder)
labextension/package.json       → TypeScript source (built into the wheel)
```

The `backend/` directory is gone — `kale/` now lives at the repo root. The `labextension/` directory stays but only contains the TypeScript source and build config. The built extension output goes to `jupyterlab_kubeflow_kale/labextension/` which gets bundled into the wheel.

### Build system

Switched from setuptools + setuptools_scm to hatchling + hatch-jupyter-builder. This is what JupyterLab's [official extension template](https://github.com/jupyterlab/extension-template) generates and what most extensions use. The jupyter-builder hook runs `jlpm build:prod` during `uv build` to compile the TypeScript and bundle it into the wheel.

For development, we skip the hook with `SKIP_JUPYTER_BUILDER=1` and build the labextension separately — same pattern as before, just cleaner.

### Version management

Before: setuptools_scm derived the version from git tags, which meant you needed `SETUPTOOLS_SCM_PRETEND_VERSION` everywhere. Now: version lives in `kale/__init__.py` as `__version__ = "2.0.0a1"` and hatchling reads it directly. Simple, explicit, no magic.

The labextension version in `labextension/package.json` uses semver (`2.0.0-alpha.1`). `make check-versions` validates they match.

### Server extension

The server extension module moved from `kubeflow_kale_labextension` (its own package) to `jupyterlab_kubeflow_kale` (a module inside the main package). The jupyter server config was updated to point to the new module path.

### Files removed

- `backend/pyproject.toml`, `backend/setup.py`, `backend/MANIFEST.in` — replaced by root `pyproject.toml`
- `labextension/pyproject.toml` — no longer a separate package
- `labextension/kubeflow_kale_labextension/` — moved to `jupyterlab_kubeflow_kale/`
- uv workspace config (`[tool.uv.workspace]`, source overrides) — not needed with single package

### Other cleanups

- Fixed stale `.gitignore` entries
- Removed dead jupyter-releaser config from `pyproject.toml`
- Fixed GitHub URLs (`kubeflow-kale/kale` → `kubeflow/kale`)
- Fixed e2e test path to `examples/`
- Updated CI workflows, Makefile, and docs for the new structure

## How to review

Most of the diff is file moves (`backend/kale/` → `kale/`). The actual code changes are in:
- `pyproject.toml` — the new single-package config
- `kale/__init__.py` — version string + package metadata
- `jupyterlab_kubeflow_kale/__init__.py` — conditional import guard for server extension
- `Makefile` — updated targets for single package
- `CONTRIBUTING.md` — updated instructions

## Test plan

- [ ] `make dev` sets up the environment
- [ ] `make test` passes (backend + labextension)
- [ ] `make lint` passes
- [ ] `make build` produces a single wheel with the labextension bundled
- [ ] Install works in a clean environment (see below)
- [ ] `make docker-build && make docker-run` works
- [ ] `make check-versions` validates Python/npm versions match

### Testing the wheel in a clean environment

Build the wheel first:
```bash
make build    # produces dist/kubeflow_kale-*.whl
```

**Lean backend (no jupyter):**
```bash
python -m venv /tmp/kale-test && source /tmp/kale-test/bin/activate
pip install /path/to/kale/dist/kubeflow_kale-*.whl
kale --help                          # CLI should work
python -c "from kale import __version__; print(__version__)"
```

**With JupyterLab extension:**
```bash
# Option 1: local venv
python -m venv /tmp/kale-jupyter && source /tmp/kale-jupyter/bin/activate
WHEEL=$(ls /path/to/kale/dist/kubeflow_kale-*.whl)
pip install "jupyterlab>=4.0.0,<5" "${WHEEL}[jupyter]"
jupyter lab
# Kale panel should appear in the left sidebar

# Option 2: clean Docker container
docker run --rm -p 8888:8888 \
  --add-host=host.docker.internal:host-gateway \
  -e KF_PIPELINES_ENDPOINT=http://host.docker.internal:8080 \
  -e KF_PIPELINES_UI_ENDPOINT=http://localhost:8080 \
  -v /path/to/kale/dist:/wheels quay.io/jupyter/scipy-notebook:latest \
  bash -c 'WHEEL=$(ls /wheels/kubeflow_kale-*.whl) && pip install "${WHEEL}[jupyter]" && jupyter lab --ip=0.0.0.0'
# Open the URL printed in the terminal, check that the Kale panel loads
```

### DIFF
--- a/.github/workflows/build_backend.yml
+++ b/.github/workflows/build_backend.yml
@@ -12,6 +12,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # Node is required by make dev (builds the JupyterLab extension)
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
       - name: Install uv
         uses: astral-sh/setup-uv@v4
         with:
@@ -22,8 +28,8 @@ jobs:
         with:
           python-version: "3.12"
 
-      - name: Install dependencies
-        run: SKIP_JUPYTER_BUILDER=1 uv sync --all-extras --frozen
+      - name: Set up development environment
+        run: make dev
 
       - name: Run backend tests
         run: make test-backend
@@ -33,6 +39,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # Node is required by make dev (builds the JupyterLab extension)
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+
       - name: Install uv
         uses: astral-sh/setup-uv@v4
         with:
@@ -43,8 +55,8 @@ jobs:
         with:
           python-version: "3.12"
 
-      - name: Install dependencies
-        run: SKIP_JUPYTER_BUILDER=1 uv sync --all-extras --frozen
+      - name: Set up development environment
+        run: make dev
 
       - name: Lint backend
         run: make lint-backend
--- a/.github/workflows/build_labextension.yml
+++ b/.github/workflows/build_labextension.yml
@@ -32,10 +32,8 @@ jobs:
         with:
           python-version: "3.12"
 
-      - name: Install dependencies
-        run: |
-          SKIP_JUPYTER_BUILDER=1 uv sync --all-extras --frozen
-          cd labextension && uv run jlpm install
+      - name: Set up development environment
+        run: make dev
 
       - name: Lint labextension
         run: make lint-labextension
@@ -61,17 +59,11 @@ jobs:
         with:
           python-version: "3.12"
 
-      - name: Install dependencies
-        run: |
-          SKIP_JUPYTER_BUILDER=1 uv sync --all-extras --frozen
-          cd labextension && uv run jlpm install
+      - name: Set up development environment
+        run: make dev
 
-      - name: Build labextension
-        run: |
-          cd labextension && uv run jlpm build:prod
-
-      - name: Register extension for verification
-        run: uv run jupyter labextension develop --overwrite .
+      - name: Build wheel
+        run: make build
 
       - name: Verify extension
         run: |
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.11"
 resolution-markers = [
     "python_full_version >= '3.14'",
@@ -105,11 +105,11 @@ wheels = [
 
 [[package]]
 name = "async-lru"
-version = "2.1.0"
+version = "2.2.0"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/ef/c3/bbf34f15ea88dfb649ab2c40f9d75081784a50573a9ea431563cab64adb8/async_lru-2.1.0.tar.gz", hash = "sha256:9eeb2fecd3fe42cc8a787fc32ead53a3a7158cc43d039c3c55ab3e4e5b2a80ed", size = 12041, upload-time = "2026-01-17T22:52:18.931Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/05/8a/ca724066c32a53fa75f59e0f21aa822fdaa8a0dffa112d223634e3caabf9/async_lru-2.2.0.tar.gz", hash = "sha256:80abae2a237dbc6c60861d621619af39f0d920aea306de34cb992c879e01370c", size = 14654, upload-time = "2026-02-20T19:11:43.848Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/2e/e9/eb6a5db5ac505d5d45715388e92bced7a5bb556facc4d0865d192823f2d2/async_lru-2.1.0-py3-none-any.whl", hash = "sha256:fa12dcf99a42ac1280bc16c634bbaf06883809790f6304d85cdab3f666f33a7e", size = 6933, upload-time = "2026-01-17T22:52:17.389Z" },
+    { url = "https://files.pythonhosted.org/packages/13/5c/af990f019b8dd11c5492a6371fe74a5b0276357370030b67254a87329944/async_lru-2.2.0-py3-none-any.whl", hash = "sha256:e2c1cf731eba202b59c5feedaef14ffd9d02ad0037fcda64938699f2c380eafe", size = 7890, upload-time = "2026-02-20T19:11:42.273Z" },
 ]
 
 [[package]]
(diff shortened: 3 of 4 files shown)


ederign:
@jesuino William, when you test this, can you test both in an 'old' environment and also with a fresh, new cloned environment? 

just to make sure that jupyter don't act weird with the rename

.github/workflows/build_labextension.yml
```diff
@@ -37,7 +34,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          uv sync --all-packages --all-extras --frozen
+          SKIP_JUPYTER_BUILDER=1 uv sync --all-extras --frozen
```
jesuino: I don't see any other reference to this variable besides on Makefile, which is simple setting it. Is it still needed?

.github/workflows/build_labextension.yml
```diff
@@ -66,60 +63,18 @@ jobs:
...
           cd labextension && uv run jlpm install
 
       - name: Build labextension
-        run: make build-labextension
+        run: |
+          cd labextension && uv run jlpm build:prod
```
jesuino: why not just build everything `make build` or keep the `make build-labextension`  but adjusted accordingly to your changes? I am thinking about this failing and contributors not being experienced on this command to reproduce it.

.github/workflows/build_labextension.yml
ederign: SKIP_JUPYTER_BUILDER is an env var recognized by https://github.com/jupyter/hatch-jupyter-builder, the build hook plugin. Here's how it works, When SKIP_JUPYTER_BUILDER=1 is set, the plugin sees it and skips that automatic npm build entirely (plugin.py: "Skipping the build hook since SKIP_JUPYTER_BUILDER was set").

ederign: I'll remove this as makes sense your last comment (to improve reproducibility of the PR)

.github/workflows/build_labextension.yml
ederign: I did this fix william, can you review again? The only drawback is to use make dev (requirement for make build) I had to install note in the GHA, but just a small slowness for great benefit as you described.

ederign:
@jesuino the first commit changed the hash because I had to do a rebase.

.github/workflows/build_backend.yml
```diff
@@ -25,8 +28,8 @@ jobs:
         with:
           python-version: "3.12"
 
-      - name: Install dependencies
-        run: uv sync --all-packages --all-extras --frozen
```
ada333, line 29: is it okay that `make dev `does not use the `--frozen `flag?

uv.lock
```diff
@@ -1194,31 +1161,31 @@ wheels = [
 
 [[package]]
 name = "kfp-pipeline-spec"
-version = "2.16.0"
+version = "2.15.2"
```
hmtosi: I believe we need to require `kfp==2.16.0`

ederign:
@ada333 about your comment on check-versions, it was a left over when I split the PR into two: it's added: https://github.com/kubeflow/kale/pull/673/changes/ca91aad0901dd5b7f28312e1143bd68820b117ce#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

.github/workflows/build_backend.yml
ederign: @ada333 it's intentional:

 --frozen in CI: Ensures reproducibility. Fails if uv.lock is out of date — you want CI to catch lockfile drift, not silently fix it.
  - No --frozen in make dev: Allows uv sync to update uv.lock if pyproject.toml changed. For local dev, this is what you want — a developer pulls changes, runs make dev, and it just works.

  If make dev used --frozen, devs would get confusing errors every time someone added a dependency and they'd have to manually run make lock first.

uv.lock
ederign: Nice catch:

Fixed after rebase:

[[package]]
name = "kfp-pipeline-spec"
version = "2.16.0"
source = { registry = "https://pypi.org/simple" }
dependencies = [
    { name = "protobuf" },
]

ederign:
@hmtosi thanks for the review:

  1. make dev "run make jupyter" message:

  The message is correct — make dev → make jupyter is still the local development workflow. make dev sets up the environment (installs deps, builds the extension, links it),
  then make jupyter starts JupyterLab. No changes there.
  
   2. make build fails with venvs:

  Should be fixed. Can you give it a try?  I've added .venv* to the sdist exclude list in pyproject.toml. Previously, hatchling only excluded .venv by default, so custom-named venvs (like .venv-py312-kfp216)
  were being included in the source distribution and hitting the absolute symlink error. Now any .venv* directory is excluded.
  
    3. Wheel + KFP pipeline execution:

  This is expected right? but for should we should improve our docs. Any suggestion? For now, it's a matter of using the KFP dev workflow. make build creates a distributable wheel, but to test pipelines against a KFP cluster you need the dev   environment with the wheel server running. The workflow is documented in CONTRIBUTING.md#testing-with-kfp-clusters:
  
  Terminal 1: Serve the local wheel
  make kfp-serve

  Terminal 2: Start JupyterLab with KFP env vars
  make jupyter-kfp

  For minikube specifically, you need to override KFP_HOST_ADDR since we default to host.docker.internal (Docker Desktop / Kind):

  KFP_HOST_ADDR=host.minikube.internal make jupyter-kfp

  This is also documented in the CONTRIBUTING.md#extension-workflow-jupyterlab-ui.

uv.lock
ederign: updated with rebase: https://github.com/ederign/kale/blob/eb0f828b0c3df00667312c51b7ea50061293dd3e/uv.lock#L1163

hmtosi: Thank you, this has been resolved and is working properly

PR #678, opened 2026-03-09
fix(backend): use json format for xgboost model marshalling

### LINKED ISSUE #675: [frontend] Iris xgboost example does not produce train model artifact
I noticed that this warning/error is only in xgboost Iris - the SKLearn works fine.
I think the fix will be one-liner.

Image

---

ai-naymul:
Fixes #675

@ada333 @ederign @jesuino 

before:
<img width="2935" height="1427" alt="560345128-cf7747b6-7be4-4fcb-8765-61a2d9b9e5ef" src="https://github.com/user-attachments/assets/8a8db72f-ff06-4deb-beda-a4f8f4b61476" />


after:
<img width="1466" height="965" alt="Screenshot (14)" src="https://github.com/user-attachments/assets/08a3aa9a-1323-4be1-91d8-a51062044cd9" />

<img width="1491" height="942" alt="Screenshot (13)" src="https://github.com/user-attachments/assets/18355dcc-51d6-450c-94d8-5ac79d2c8dc2" />




xgboost 2.x shows a deprecation warning when saving models with the old .bst format. This warning ends up in the train_model HTML report as a pink error box, which is what we see in #675. 

Fixed by switching the save format from .bst to .json which is the recommended format now. Also added a missing skip tag on the pip install cell in the xgboost iris notebook to match the   sklearn one.                                                                                                                                                                                
                                                                                                                                                                                              
Tested locally on a KFP cluster and the pipeline runs clean with no warnings.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,285 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      target:
+        description: 'Release target'
+        type: choice
+        default: 'dry-run'
+        options:
+          - dry-run
+          - testpypi
+          - testpypi+pypi
+
+concurrency:
+  group: release-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  prepare:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      tag: ${{ steps.version.outputs.tag }}
+      prerelease: ${{ steps.version.outputs.prerelease }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Extract version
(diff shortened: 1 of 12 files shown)


ai-naymul:
btw I noticed prow removes lgtm when I rebase to update the branch. Should I wait until both lgtm and approved are set before rebasing, or is there a preferred way to handle this?

cc: @ederign @jesuino @ada333

ai-naymul:
@jesuino @hmtosi @ada333 

Rebased onto latest main after #670 path changes.

ai-naymul:
@jesuino @ada333 @ederign  this one fixes #675 (xgboost deprecation warning in iris example). Small change, tested on local KFP.

PR #689, opened 2026-03-12
Upgrade e2e test

### LINKED ISSUE #501: [bug] security violation of PSS baseline/restricted

### LINKED ISSUE #565: chore(testing): Add verification of artifact generation in e2e test
## Chore description
We should add some more checks in e2e test - definitely a check that artifacts were generated correctly and maybe that the pipeline parameters work as expected.

/area testing

Love this idea? Give it a 👍.

ada333:
This PR upgrades e2e test, for review/testing I left the workflow_dispatch tag (so it can be run manually) so anyone can run it and try it.
Changes I made:

1. Added e2e to CI - triggered for every push and PR
2. Added PSS restricted mode for kubeflow namespace - should close #501
3. Added check for output artifact generation (that they exist and are not empty) - old issue where output artifacts were not created / empty in MinIO (https://github.com/kubeflow/kale/issues/511) -> should close #565 
4. Added simple check for metrics (PR added recently) - just that they were created.


hmtosi (approved):
/lgtm

The code changes look good, and I see that PSS has been assigned as requested in the ticket. The action is passing on the runner

Image

jesuino (approved):
LGTM!

Thanks for this new test. With this we are making sure that the marshallers are being correctly executed for and generating an artifact.

.github/workflows/e2e-test.yml
```diff
@@ -25,17 +32,29 @@ jobs:
 
       - name: Install KFP
         run: |
-          export PIPELINE_VERSION=${{ inputs.pipeline_version }}
+          export PIPELINE_VERSION=${{ env.KFP_VERSION }}
```
juliusvonkohout, line 56: Here you have to run Kubectl create namespace kubeflow and annotate it with PSS restricted. Only that way all pods that are later deployed are checked. It must happen before you deploy any kfp manifests.

ederign: @hmtosi @ada333 ^

ada333:
@juliusvonkohout thanks for review, changes added as requested

.github/workflows/e2e-test.yml
juliusvonkohout: Please make sure to use the overwrite flag and I think you should have a negative test. So crease a dummy pod with the root user and see that it fails.

juliusvonkohout:
@ederign for final review.

ederign (approved):

PR #695, opened 2026-03-12
Removing Outdated UI Metadata method

### LINKED ISSUE #664: [backend] Investigate if 'update_uimetadata' is still needed in KFP v2
### Environment

KFP 2.x

*  Kale version:
Unreleased 2.0

Investigate if the `UI metadata` related methods are still needed for KFP v2.16. See [KFP Utils](https://github.com/kubeflow/kale/blob/main/backend/kale/common/kfputils.py#L237)

### LINKED ISSUE #688: [backend] Pipeline steps fail when pod name exceeds 63 characters (HOSTNAME truncation)
### Environment

I'm not sure if this reproduces only on OSX or also in Linux. Folks please double check.

Also, I never saw this issue before the KFP 2.16 upgrade. This used to work on KFP 2.14.3 / Argo v3.6.7. After upgrading to KFP 2.16.0 / Argo v3.7.3 with a fresh install in my local, they fail. This could indicate a change in the pod naming scheme (e.g. longer suffixes in Argo v3.7.x), but I could not verify this since the old KFP deployment was deleted before investigation.

### Problem

Pipeline steps fail with a `404 Not Found` error when the Argo workflow generates pod names longer than 63 characters. The step's user code executes successfully, but the post-execution `update_uimetadata` call crashes.

```
kubernetes.client.exceptions.ApiException: (404)
Reason: Not Found
"pods \"candies-sharing-216-fresh-8bzv4-system-container-impl-404075885\" not found"

### How to reproduce

Use any pipeline name — Argo appends suffixes like `-system-container-impl-XXXXXXXXX` to pod names, so even moderately long pipeline names push the total pod name past 63 chars.

### Test results

| Pipeline name | Pod name length | HOSTNAME | Result |
|---|---|---|---|
| `small-name` | < 63 chars | matches pod name | Success |
| `candies-sharing-216-fresh` | 64 chars | truncated to 63 | 404 failure |
| `loooonngggnamebiggerthan65chaaaarrrssssssssss` | 83 chars | truncated to 63 | 404 failure |

Kubernetes seems to silently truncates the container hostname to 63 characters when the pod name exceeds this limit. The `HOSTNAME` env var inside the container does not match the actual pod name.
---

jesuino:
closes #664 
closes #688

UI Metadata is not used anymore by KFP v2 hence I am removing this because this was also the cause for the error from #688 

To test this PR just compile notebooks and run pipelines from it as usual. You may also want to test a pipeline with a very long name so we will know if it is running correctly. Here's a string:

`loooonngggnamebiggerthan65chaaaarrrssssssssss-loooonngggnamebiggerthan65chaaaarrrssssssssss`

I didn't add any additional validation because I see that KFP is working correctly after removing UI metadata related method.

### DIFF
--- a/labextension/src/lib/CellUtils.ts
+++ b/labextension/src/lib/CellUtils.ts
@@ -110,8 +110,7 @@ export default class CellUtilities {
     notebookPanel: NotebookPanel,
     index: number,
     key: string,
-    value: any,
-    save: boolean = false,
+    value: any
   ): Promise<any> {
     if (!notebookPanel || !notebookPanel.model) {
       throw new Error('Notebook was null!');
@@ -128,10 +127,6 @@ export default class CellUtilities {
       if (metadata) {
         oldVal = metadata[key];
         cell.setMetadata(key, value);
-      }
-
-      if (save) {
-        return notebookPanel.context.save();
       }
       return Promise.resolve(oldVal);
     } catch (error) {
--- a/labextension/src/lib/NotebookUtils.tsx
+++ b/labextension/src/lib/NotebookUtils.tsx
@@ -105,6 +105,27 @@ export default class NotebookUtilities {
     );
   }
 
+/**
+   * Opens a pop-up dialog in JupyterLab to display a simple message.
+   * @param title The title for the message popup
+   * @param msg The message as an array of strings
+   * @param buttonLabel The label to use for the button. Default is 'OK'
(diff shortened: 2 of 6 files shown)


jesuino:
@ederign Hannah found new issues, but she opened a new ticket to work on these later. Would it be okay to go ahead and merge this?

PR #696, opened 2026-03-13
Improve error message when KFP server doesn't support securityContext

### LINKED ISSUE #687: [bug] Improve error message when KFP server doesn't support securityContext
## Problem

After PR #656 (PSS baseline - set security context), Kale compiles pipelines with `securityContext` in the kubernetes executor config. When the KFP server is too old to recognize this field, the pipeline submission fails with a cryptic error:

```
failed to unmarshal kubernetes config: proto: (line 1:2): unknown field "securityContext"
```

This is a server-side proto deserialization error that gives no indication of what the user should do to fix it.

## Root Cause

The `kfp.kubernetes.security_context.set_security_context()` call (added in PR #656) generates a `securityContext` field in the compiled pipeline YAML. The KFP Python SDK (`kfp[kubernetes]>=2.16.0`) supports this, but the KFP **server** must also be a version that recognizes the field in its protobuf definitions.

Example version mismatch:
- Client SDK: `kfp[kubernetes] 2.16.0` (supports `securityContext`)
- Server: `kfp-api-server 2.14.3` (does NOT support `securityContext`)

## Proposed Fix

Catch the `securityContext` unknown field error in the RPC layer and return a user-friendly message, e.g.:

> "Your KFP server does not support the `securityContext` field. Please upgrade your Kubeflow Pipelines deployment to version >= 2.16.0."

## Screenshot

Image

## References

- PR #656: Issue 501: PSS baseline - set security context
- KFP server image: `ghcr.io/kubeflow/kfp-api-server`
- KFP releases: https://github.com/kubeflow/pipelines/releases

### Labels

  /area frontend  
  /area backend  

---

CodeVishal-17:
Fixes #687

This PR improves the error message shown when the KFP server
does not support the `securityContext` field.

Previously users would see a cryptic protobuf error:

"failed to unmarshal kubernetes config: proto: unknown field 'securityContext'"

This change catches the exception raised during pipeline submission
and provides a clearer message suggesting upgrading Kubeflow Pipelines
to version >= 2.16.0.

### DIFF
--- a/kale/common/kfputils.py
+++ b/kale/common/kfputils.py
@@ -61,9 +61,7 @@ def get_pipeline_id(pipeline_name: str, host: str = None) -> str:
         pipelines = client.list_pipelines(page_token=token)
         token = pipelines.next_page_token
         f = next(
-            filter(
-                lambda x: x.display_name == pipeline_name, pipelines.pipelines or []
-            ),
+            filter(lambda x: x.display_name == pipeline_name, pipelines.pipelines or []),
             None,
         )
         if f is not None:
@@ -71,9 +69,7 @@ def get_pipeline_id(pipeline_name: str, host: str = None) -> str:
     return pipeline_id
 
 
-def get_pipeline_version_id(
-    version_name: str, pipeline_id: str, host: str = None
-) -> str:
+def get_pipeline_version_id(version_name: str, pipeline_id: str, host: str = None) -> str:
     """List through the versions and filter by version name.
 
     Args:
@@ -155,9 +151,7 @@ def upload_pipeline(
             pipeline_id=pipeline_id,
         )
         version_id = upv.pipeline_version_id
-        log.info(
-            "Uploaded version '%s' for pipeline '%s'.", version_name, pipeline_name
-        )
+        log.info("Uploaded version '%s' for pipeline '%s'.", version_name, pipeline_name)
     return pipeline_id, version_id
 
 
@@ -201,9 +195,7 @@ def run_pipeline(
 
     if not run_name:
         run_name = f"{pipeline_name}-{version_name}-{utils.random_string()}"
-    display_version = "({}version: '{}')".format(
-        "" if version_id else "default ", version_name
-    )
+    display_version = "({}version: '{}')".format("" if version_id else "default ", version_name)
     log.info(
         "Submitting new pipeline run '%s' for pipeline '%s' %s ...",
         run_name,
@@ -232,13 +224,10 @@ def run_pipeline(
                     ):
                         message = body.get("message", "").lower()
 
-                        if (
-                            "failed to unmarshal kubernetes config" in message
-                            and re.search(
-                                r"unknown field.*securitycontext",
-                                message,
-                                re.IGNORECASE,
-                            )
+                        if "failed to unmarshal kubernetes config" in message and re.search(
+                            r"unknown field.*securitycontext",
+                            message,
+                            re.IGNORECASE,
                         ):
                             raise RuntimeError(
                                 "Your KFP server does not support the 'securityContext' field. "
@@ -269,9 +258,7 @@ def generate_mlpipeline_metrics(metrics):
         with open(KFP_UI_METRICS_FILE_PATH, "w", encoding="utf-8") as _kale_mf:
             json.dump(metrics, _kale_mf)
     except (OSError, TypeError, PermissionError) as e:
-        log.error(
-            "Not able to create metrics file, an unexpected error happened: %s", e
-        )
+        log.error("Not able to create metrics file, an unexpected error happened: %s", e)
 
 
 def load_mlpipeline_metrics(output):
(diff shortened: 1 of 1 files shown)


kale/common/kfputils.py
```diff
@@ -194,13 +194,24 @@ def run_pipeline(
...
+    try:
+        run = client.create_run_from_pipeline_package(
+            pipeline_file=pipeline_package_path,
+            arguments=kwargs,
+            run_name=run_name,
+            experiment_name=experiment_name,
+        )
+    except Exception as e:
+        if 'unknown field "securityContext"' in str(e):
+            log.error(
```
ada333: is there a reason to log the error and raise the Exception with the same error message? I think the error log might be redundant.

CodeVishal-17: Good catch ,you're right that the extra log was redundant since the exception already surfaces the message. I've removed the `log.error` and pushed the update.

kale/common/kfputils.py
```diff
@@ -194,13 +194,20 @@ def run_pipeline(
...
-    )
+    try:
+        run = client.create_run_from_pipeline_package(
+            pipeline_file=pipeline_package_path,
+            arguments=kwargs,
+            run_name=run_name,
+            experiment_name=experiment_name,
+        )
+    except Exception as e:
+        if 'unknown field "securityContext"' in str(e):
```
ederign: String matching is a little dangerous. If the protobuf error message changes wording, gets localized, or the KFP server changes how it formats the error, this silently stops working.

We can either:
- Check server version upfront in run_pipeline. Same idea but at submission time, check compatibility and give the friendly error before attempting submission.
- At minimum, catch the specific exception type (if there is one)

CodeVishal-17: Thanks for the suggestion  I’ve updated the implementation to catch `ApiException` and now check the error using the response body instead of relying on string matching.

This should make it more robust against changes in error formatting. Let me know if you'd prefer checking server compatibility upfront instead.

kale/common/kfputils.py
```diff
@@ -194,13 +195,20 @@ def run_pipeline(
...
-    )
+    try:
+        run = client.create_run_from_pipeline_package(
+            pipeline_file=pipeline_package_path,
+            arguments=kwargs,
+            run_name=run_name,
+            experiment_name=experiment_name,
+        )
+    except ApiException as e:
+        if e.status == 400 and "securityContext" in str(e.body or ""):
```
jesuino: Hello @CodeVishal-17 I still believe that we could be more specific.

Checking only the string on the body is too broad and it may lead to false results, which is very bad for  users in production environments. For example, let's say that the error is:

`The field userId was not found in the list of fields abc, xyz, securityContext`

This is not an actual scenario, but notice that with the error message above users would see `Your KFP server does not support the 'securityContext' field.`, which is misleading.

CodeVishal-17:
> Hello @CodeVishal-17
> 
> Thanks for your contribution. Please see my comment below.
> 
> While the change works and will likely be correct for most scenarios, we must be more specific as possible to avoid giving the users misleading messages.
> 
> Do we know what is returned when such errors happens? If now, I would suggest you to first reproduce the error to inspect the response and make it sure that we are very specific before changing the user message otherwise users may see some message that is not the actual problem they are facing.

@jesuino Thanks for the feedback ,that makes sense.

I've updated the check to match the specific protobuf error (`unknown field "securityContext"`) instead of broadly checking for "securityContext" in the response. This should avoid false positives and prevent misleading messages.

Please let me know if you'd prefer a more flexible check.

CodeVishal-17:
Hi! @jesuino  Thanks a lot for the detailed explanation and reproduction steps — that really helped.

I’ve updated the implementation to follow your suggested approach:

* Parse the `ApiException` body as JSON
* Check the top-level `code == 13`
* Validate the `details` entries for `type.googleapis.com/google.rpc.Status` and `code == 2`
* Finally inspect the error message to detect the `failed to unmarshal kubernetes config` with the `unknown field "securityContext"` case

I also removed the dependency on the HTTP status code since, as you pointed out, it can vary (e.g., 500 instead of 400).

Let me know if this looks good or if you’d like me to refine the checks further!

jesuino:
Please run `make format-backend` and merge your branch with updated main.

Thanks!

CodeVishal-17:
> Please run `make format-backend` and merge your branch with updated main.
> 
> Thanks!

@jesuino Thanks!, I've applied the backend formatting and synced with the latest main branch.

jesuino:
Hello @CodeVishal-17 

Thanks for doing this. For some reason the lint modified a lot of files unrelated to your fix. Which command did you run to fix the lint and what Python version are you using? Unfortunately the linting is still failing and now also pytest seems to be failing.

Thanks!

CodeVishal-17:
> Hello @CodeVishal-17
> 
> Thanks for doing this. For some reason the lint modified a lot of files unrelated to your fix. Which command did you run to fix the lint and what Python version are you using? Unfortunately the linting is still failing and now also pytest seems to be failing.
> 
> Thanks!

@jesuino Thanks for the review!

I’ve cleaned up the branch to remove unrelated changes, and now both linting and tests are passing.

Please let me know if there’s anything else I should improve.

PR #698, opened 2026-03-13
fix(backend): fix inverted condition in podutils and loop bug in rpc/kfp

### LINKED ISSUE #697: [backend] Inverted condition in get_container_name function makes return unreachable
ref:  https://github.com/kubeflow/kale/blob/95e97ac0eaf432f7c875fe0610fa562e4aaeed83/kale/common/podutils.py#L123

When filtering out sidecar containers leaves exactly one candidate (the expected success case), the function raises `"No container names left"` instead of returning the candidate. When candidates is empty, it falls through to `candidates[0]` which raises an IndexError.

Expected: if len(candidates) == 0 so that one remaining candidate is returned successfully.

ai-naymul:
Fixes #697

cc: @ada333 @jesuino @ederign

### DIFF
--- a/kale/common/podutils.py
+++ b/kale/common/podutils.py
@@ -117,10 +117,10 @@ def get_container_name():
     candidates = [c for c in container_names if all(x not in c for x in potentially_sidecar_names)]
     if len(candidates) > 1:
         raise RuntimeError(
-            "Too many container candidates.Cannot infer the"
+            "Too many container candidates. Cannot infer the"
             f" name of the current container from: {candidates} "
         )
-    if len(candidates) > 0:
+    if len(candidates) == 0:
         raise RuntimeError(
             "No container names left. Could not infer the name of the running container."
         )
--- a/kale/rpc/kfp.py
+++ b/kale/rpc/kfp.py
@@ -84,10 +84,10 @@ def _get_pipeline_id(pipeline_name):
     client = _get_client()
     token = ""
     pipeline_id = None
-    while pipeline_id is None or token is not None:
+    while pipeline_id is None and token is not None:
         pipelines = client.list_pipelines(page_token=token)
         token = pipelines.next_page_token
-        f = next(filter(lambda x: x.display_name == pipeline_name, pipelines.pipelines), None)
+        f = next(filter(lambda x: x.display_name == pipeline_name, pipelines.pipelines or []), None)
         if f is not None:
             pipeline_id = f.pipeline_id
     return pipeline_id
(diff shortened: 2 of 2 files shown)


ai-naymul:
@jesuino @ada333  @ederign  this fixes the inverted condition in podutils and a loop bug in rpc/kfp. Happy to address any feedback.

ederign:
how did you reproduce this bug? can you record a video of it?

ai-naymul:
> how did you reproduce this bug? can you record a video of it?

@ederign sorry for the late reply. Here's the walkthrough: https://www.loom.com/share/c43060725515427ba506d59aa9b25582                        
                                         
Also wanted to flag, while auditing this I noticed `_get_pipeline_id` in `rpc/kfp.py` is actually never called anywhere. The live path for upload goes through `kfputils.upload_pipeline`, which uses the correct version in kfputils. So another option is to just delete it instead of ffixing it. Happy either way, let me know what you prefer.                                                                                     
                                                            
  cc @jesuino @ada333

kale/common/podutils.py
```diff
@@ -117,10 +117,10 @@ def get_container_name():
     candidates = [c for c in container_names if all(x not in c for x in potentially_sidecar_names)]
     if len(candidates) > 1:
         raise RuntimeError(
-            "Too many container candidates.Cannot infer the"
+            "Too many container candidates. Cannot infer the"
```
jesuino, line 120: Do you think this method is used at all? I see it in `get_container_name`, which is used in `pipeline._set_base_image` and when there's an error, it defaults to the base image (see `pipeline._set_base_image`)- and in my tests it always have error - I would to remove this code and keep the base image as the default one and soon use the mechanism introduced on this PR: https://github.com/kubeflow/kale/pull/737

kale/rpc/kfp.py
```diff
@@ -84,10 +84,10 @@ def _get_pipeline_id(pipeline_name):
     client = _get_client()
     token = ""
     pipeline_id = None
-    while pipeline_id is None or token is not None:
+    while pipeline_id is None and token is not None:
```
jesuino: Same for this change, it seems that the method `_get_pipeline_id` is not used anymore - we can simply delete it. Would you please confirm?

jesuino (changes requested):
Hello @ai-naymul  - thanks again for another contribution!

It seems that we are dealing with dead/unused code, would you please confirm? If this is the case we can simply delete these methods. For the base image name, we can use the default.

Thanks!

ai-naymul:
hey @jesuino confirmed it's dead code. `_get_pipeline_id` is never called anywhere, `upload_pipeline` goes through `kfputils` directly. went ahead and deleted it. the `podutils.py` fix stays since `get_container_name` is actively used. updated the PR.

jesuino (approved):
LGTM!

Later we can refactor the backend and remove unused code.

Thanks for your contribution!

ederign (approved):
/approve
/lgtm

PR #702, opened 2026-03-16
issue 680: configure `kale` to use different `kfp` server

### LINKED ISSUE #680: [feature] Configure Kale to use different KFP server
### Feature Area

/area frontend
/area backend

* **KFP endpoint is configurable:** When a user sets a custom KFP API endpoint (host/port), then Kale uses that endpoint for all pipeline submissions and run tracking instead of the default in-cluster service.
* **Configuration persists across sessions:** When a user saves the KFP endpoint configuration, then it persists across Jupyter kernel restarts and notebook sessions without requiring re-entry.
* **Authentication credentials are supported:** When the target KFP server requires authentication (e.g., bearer token, cookie, or mTLS), then the user can provide those credentials as part of the configuration, and Kale uses them transparently on each request.
* **Connection is validated before use:** When a user saves the new KFP endpoint, then Kale performs a connectivity check and surfaces a clear success or error message indicating whether the server is reachable and responding.
* **Invalid configuration is rejected gracefully:** When the provided endpoint is malformed or unreachable, then Kale displays a descriptive error and does not attempt pipeline compilation or submission.
* **Default behavior is preserved:** When no custom endpoint is configured, then Kale falls back to the in-cluster KFP service discovery (e.g., ml-pipeline.kubeflow.svc.cluster.local) as it does today.

* Let developers use same KFP Server
* Let the users to us different KFP Server flavours/locations on their kubernetes instances.

No there is no workaround, if KFP Server is in a different location, or non-standard location it won't work.

---

Love this idea? Give it a 👍.

hmtosi:
Closes #680 
This is the backend ticket only.
UI will be addressed in #713 

#### Quick Manual Check
1. Start your `kfp` server on a known port (default is 8080)
2. Build the `kale` extension and start JupyterLab
3. In the `Kale` tab, verify that KFP connection status is `Connected`
4. From the Launcher page, open a new Terminal and update the configuration
```python
    # change directory to kale
    cd kale

    # launch a new interactive session
    python

    >>> # save a new configuration
    >>> from kale.config import kfp_server_config
    >>> kfp_server_config.save_config({"host": "http://localhost:8081"})
```
[Note] - Any of the six configuration parameters can be adjusted in this way - any parameters you do not adjust will remain as the default

5. Verify that the KFP connection status has changed to `Disconnected` (you may need to refresh JupyterLab)
6. Find the `kfp_server_config.json` file on your local machine (it should be at `home/.kale/kfp_server_config.json`), and make sure it has been modified with the latest changes.

#### Unit Test
In development


kale/common/kfputils.py
```diff
@@ -40,8 +41,55 @@
...
+    if credentials is None:
+        credentials = config.credentials
+    if existing_token is None:
+        existing_token = config.existing_token
+    if namespace is None:
+        namespace = config.namespace or "kubeflow"
+    if ssl_ca_cert is None:
+        ssl_ca_cert = config.ssl_ca_cert
+
+    return kfp.Client(
```
jesuino: It looks like that the client is also created on file `kfp.py`. Would it make sense to use this code there or create a new file  to act as kfp client factory which both `kfp.py` and `kfputils.py` would use to create `kfp.Client()`. Please let me know what you think.

jesuino (changes requested):
Hello Hannah,

Thanks for your PR, it seems that everything is on place for allowing us to make a fine configuration of the `kfp.Client()` creation!

I mentioned another place where the kfp.Client is called, would you please check my comment and let me know your thoughts?

Thanks!

kale/common/kfputils.py
hmtosi: @jesuino thank you for pointing this out. I believe my changes have addressed it.

kale/common/kfp_client_factory.py
```diff
@@ -0,0 +1,75 @@
...
+    config = kfp_server_config.load_config()
+
+    # Use parameter if provided, otherwise fall back to config
+    if host is None:
+        host = config.host
+    if cookies is None:
+        cookies = config.cookies
+    if credentials is None:
+        credentials = config.credentials
+    if existing_token is None:
```
ada333: When testing Kale in RHOAI, I noticed that `existing_token` value changes a lot. The default should be taken from `/var/run/secrets/kubernetes.io/serviceaccount/token` - this way the token will automatically update.

kale/config/kfp_server_config.py
```diff
@@ -0,0 +1,84 @@
...
+        with open(config_path) as f:
+            config_dict = json.load(f)
+        log.info("Loaded KFP server config from %s", config_path)
+        return KFPServerConfig(**config_dict)
+    except (json.JSONDecodeError, OSError) as e:
+        log.warning("Failed to load KFP server config from %s: %s. Using defaults.", config_path, e)
+        return KFPServerConfig()
+
+
+def save_config(config: KFPServerConfig | dict[str, Any]) -> None:
```
ada333, line 118: I believe this func is not used anywhere, do we want to keep it?

kale/common/kfp_client_factory.py
ada333: And I believe the existing_token should not be in config file at all, so it doesnt stay there still without update

kale/config/kfp_server_config.py
jesuino: I think we should keep for the UI integration

kale/config/kfp_server_config.py
```diff
@@ -0,0 +1,83 @@
...
+    host = Field(type=str, default=None)
+    auth_type = Field(type=str, default="none")
+    auth_params = Field(type=dict, default=None)
+    namespace = Field(type=str, default="kubeflow")
+    ssl_ca_cert = Field(type=str, default=None)
+
+
+def get_config_path() -> str:
+    """Get the path to the KFP server configuration file."""
+    kale_dir = os.path.join(os.path.expanduser("~"), ".config/kale")
```
jesuino, line 52: would it be possible to have a env var to change the config path if user wants to?

jesuino (changes requested):
It is looking great and working well locally! I just have one final solution.

Thanks!

kale/config/kfp_server_config.py
hmtosi: yes let's keep for the ui integration

kale/config/kfp_server_config.py
hmtosi: good suggestion, I have added it above. Please let me know if it looks good.

jesuino (approved):
LGTM!

kale/config/kfp_server_config.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Create .kale directory if it doesn't exist
+    os.makedirs(kale_dir, exist_ok=True)
+
+    config_dict = config.to_dict()
+    with open(config_path, "w") as f:
+        json.dump(config_dict, f, indent=2)
+
+    # Set restrictive permissions for security (only owner can read/write)
+    os.chmod(config_path, 0o600)
```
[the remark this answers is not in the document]
hmtosi: I have adjusted this in the code

kale/common/kfp_client_factory.py
hmtosi: I believe these token updates are system-specific, so I will leave it as-is for now. The user is able to manually update the `existing_token` in the config if needed.

jesuino (approved):
LGTM!

hmtosi:
@ederign @StefanoFioravanzo this is ready for your final review/approval when you have a moment, thank you!

kale/common/kfp_client_factory.py
```diff
@@ -0,0 +1,78 @@
...
+    namespace: str | None = None,
+    ssl_ca_cert: str | None = None,
+) -> "Client":
+    """Create a KFP client with configuration.
+
+    Loads saved configuration from ~/.config/kale/kfp_server_config.json and allows
+    parameter overrides. Explicit parameters override saved config if they are provided.
+
+    Authentication is handled by creating credentials at runtime using the authenticator
+    module, so credential objects are never serialized to disk.
```
ederign: We kind of do this indirectly in the JSON, right?

hmtosi:
Hi @ada333 @jesuino, this is ready for review again. I have made some changes to the authentication security so that...
- secrets/tokens/cookies are no longer written directly to the disk through the normal workflow
- the config file has guardrails that prevent a user from accidentally writing these
- the unit tests verify these security measures
- the `README.md` is updated with some information about how to adjust authentication

----
I did the following checks on my local kale installation:

:white_check_mark: Verify that Kale is available
-build extension and open JupyterLab
-see that the `kale` icon is there
-open a notebook and toggle kale to enable it

:white_check_mark: Run a pipeline
-verify that the notebook opened in the last test has all kale metadata components
-hit "Compile and Run" and check that pipeline completes

:white_check_mark: Assign new `host`
-open a terminal in the JupyterLab session
-start an interactive Python session and then run:
```python 
from kale.config import kfp_server_config
kfp_server_config.save_config({"host": "http://localhost:8081"})
```
-this saves successfully, and then the KFP connectivity check shows that KFP is suddenly unreachable
-reversing this and reassigning `port=8080` also saves successfully, and KFP is reachable again

:white_check_mark: Try to assign new `cookies` directly
-in the same interactive Python session, run:
```python
kfp_server_config.save_config(
            	{
                	"auth_type": "dex",
                	"auth_config": {"cookies": "session=abc"},  # ← Should fail
            	}
```
-this fails with `ValueError: auth_config contains fields that look like secrets`
-the config file is not updated on my local machine

:white_check_mark: Assign new `token` with environment variable
-in the same interactive Python session, run:
```python
kfp_server_config.save_config(
            	{
                	"auth_type": "existing_bearer_token",
                	"auth_config": {"env_var": "KF_PIPELINES_TOKEN"},  # ← Should pass
            	}
```
-this succeeds  with the message `Saved KFP server config to /home/htosi/.config/kale/kfp_server_config.json`
-checking the local config file, it is updated with these values
-the KFP connectivity status becomes Disconnected because we have not set the environment var value


----
Further functionality will be added with the UI ticket. One known issue is that running the `kfp_server_config.save_config()` function overwrites the current config value. Thus, if a user saves one parameter and then later saves a different parameter (ie. assigns a new host and then separately assigns a token env var), only the last one will be saved, and all others will be overwritten/set to default. This behavior will be handled in the UI ticket.

kale/common/kfp_authenticator.py
```diff
@@ -0,0 +1,251 @@
...
+            return AuthResult(existing_token=token)
+
+        # No token found
+        raise ValueError(
+            f"Bearer token not found. Set {env_var} environment variable "
+            f"or provide file_path in auth_config"
+        )
+
+
+class DexAuthenticator:
```
ada333: just a better code practice, but we could add a general class called Authenticator with one method - authenticate. And all these specific authenticator could be child classes of this one

kale/common/kfp_authenticator.py
```diff
@@ -0,0 +1,251 @@
...
+            AuthResult with bearer token
+
+        Raises:
+            ValueError: If token cannot be resolved
+        """
+        config = config or {}
+
+        # Try env_var first
+        env_var = config.get("env_var", "KF_PIPELINES_TOKEN")
+        if env_var:
```
ada333: do we actually want to try env_var first? if user configures a file path in his `kfp_server_config.json` file doesnt it make more sense to try the file path first?

kale/config/kfp_server_config.py
```diff
@@ -0,0 +1,157 @@
...
+    config_path = get_config_path()
+    if not os.path.exists(config_path):
+        log.debug("No KFP server config found at %s, using defaults", config_path)
+        return KFPServerConfig()
+
+    try:
+        with open(config_path) as f:
+            config_dict = json.load(f)
+        log.info("Loaded KFP server config from %s", config_path)
+        return KFPServerConfig(**config_dict)
```
ada333, line 71: I think before the return there should be called a `_validate_auth_config` on the `**config_dict` object - if user manually sets the token to some value it wont give him any warning like this.

ada333 (changes requested):
@hmtosi your PR looks good, I think generally it works very good but I found some potential issues/comments.

kale/common/kfp_authenticator.py
hmtosi: Thank you, this is good feedback. I have adjusted the code

kale/common/kfp_authenticator.py
hmtosi: I am not certain the order we expect, but the code was not doing this consistently. I have changed it so it always checks the config first and then checks the env vars, in case a user has specified something in the config but there is an inherited env var that they didn't know about

kale/config/kfp_server_config.py
hmtosi: Can you explain a bit more about the security scenario you're worried about here? This is the load function, so the return function is showing what is already saved in the file (this function is not saving new information to the file). If a user manually set the token in the json file and saved it (bypassing the `_validate_auth_config` run on `save_config`), that will be loaded here. Is that what you mean, that we should warn the user they have entered a secret and saved it to disk?

ada333: yes, that is what I meant - if user manually sets the token in the json file there will be no validation, right?

hmtosi: That is correct. If a user manually sets a raw token in the json file, it will save with no warnings and no error. I just tested, and doing this also changes the kfp connection status, suggesting that the token is being read. 

Authenticating at this point (pulling the raw token out of the file) means that the token is already saved to the disk (ie. already leaked), so this is not the best place for that check in my opinion.

If we do want to authenticate that, it should be at the point of manual file save of `kfp_server_config.json`. However, I think it may be overkill to validate what users are manually entering into the json.

I think we need the team to weigh in @jesuino @ederign - we have safeguards on what users can save using the command-line functions, but we do not have any validation on what they can manually save into the json. Do we want to add that? Or do we want to at least give them a warning, as Adam is suggesting above?(note that at this point they would have already leaked the token)

adrielparedes: For me, we should not do any kind of check if the user want to explicitly do it. I agree Kale and Kubeflow shouldn't do it, or show it on the UI, but if the user in the end wants to change anything, should be it's responsibility to do that. We may have some "analyzer" in the future looking for those kind of things, but I don't think we should care right now

ada333 (approved):
LGTM! @hmtosi thanks for working on this.

jesuino (approved):
I can confirm this is working as expected:

* I did create my own configuration (`myconfig.json')
* Then I configured Kale to use it:

```
export KALE_CONFIG_PATH=/home/wsiqueir/projects/kale/myconfig.json
```

* When starting Kale I see clearly it reading the configuration instead the default because I did on purpose set an invalid URL and errors were thrown:

```
  File "/home/wsiqueir/projects/kale/.venv/lib/python3.12/site-packages/urllib3/connection.py", line 219, in _new_conn
    raise NewConnectionError(
urllib3.exceptions.NewConnectionError: HTTPConnection(host='localhost', port=9190): Failed to establish a new connection: [Errno 111] Connection refused

```

What I didn't do was test all the auth schema, but since the configuration is being correctly read I can say that this works as expected.

ederign:
@hmtosi @jesuino I wonder, as this doesn't have a UI component yet, if we can wait to merge this after the 2.0 release, just to reduce the risk.

My plan is as soon as I review/merge the docs from Stefano (this weekend), start to release 2.0

jesuino:
@ederign I would say we should merge this and start the UI work right after the 2.0 release!

ederign (approved):

Issue #706, opened 2026-03-17
[frontend] Users should be able to configure if they want the notebook to be auto saved when running it


jesuino:
After #612 fix is merged, we can make the option to avoid repeated dialog when running/compiling a notebook using Jupyter Lab settings (which should be introduced by PR #685

jesuino:
Fix is now merged on main

PR #707, opened 2026-03-17
Remove auto saving and prompt users to save when Running/Compiling the notebook

### LINKED ISSUE #612: [frontend] Kale auto saves notebook when editing the cells but not saving the changes
### Environment

Latest

### Steps to reproduce

* Open a notebook
* Enable Kale
* Modify one of the tags, but don't save the notebook and close it. Notice that the notebook was saved anyway:
```
[I 2026-02-17 21:19:53.501 ServerApp] Saving file at /examples/base/candies_sharing.ipynb
```

### Expected result

The Notebook should not be saved, instead it should be in a dirty state ans user should be prompted for saving it.

### Materials and Reference

---

jesuino:
closes #612 

I had to hurry this because the settings PR was blocking this from getting solved. On this PR:

* We remove all `notebookPanel.context.save()` and kept only the one before deployment;
* We also add a metadata listener for the active cell, so we update the panel accordingly
* The user is now prompted to save before running/compiling. I created an issue to track this -> #706

https://github.com/user-attachments/assets/6c055c49-585b-4a71-b7b0-6230e04c8519


alikhere:
HI @jesuino 
I  noticed when we create a notebook and try to change the cell type, the change is not reflected in the UI until I save the notebook. The change is only reflecting after saving the notebook.


https://github.com/user-attachments/assets/5a2b159d-f865-4a45-abed-b2c3b6d93e5f

jesuino:
@alikhere Thanks for reviewing it. The video is file corrupted and I can't watch it, but I will try to reproduce

alikhere:
> @alikhere Thanks for reviewing it. The video is file corrupted and I can't watch it, but I will try to reproduce

@jesuino See the Video:
https://drive.google.com/file/d/1jpZBD2qpclmmEx4eiYKqF_Re5BT1B2Gw/view?usp=sharing

jesuino:
@alikhere Thank you, I am able to reproduce it now and I am working on it!

jesuino:
@alikhere Great catch!

The issue was that when a notebook has only one cell, Jupyter Lab makes the cell active before we can register listeners to it, so we are not able to listen to metadata, so the changes in Cell metadata are ignored by Kale.

Would you please review it again?

Sayan4496:
Hi @jesuino 
Previously `setCellMetaData` allowed controlling persistence via the `save` flag, but this PR removes that capability.
Is this intentional? It might reduce flexibility for cases where immediate save is required.
Maybe we should ensure that critical metadata updates are still persisted safely.

jesuino:
Hello @Sayan4496 

That's intentional. When user changes the notebook we should make it clear that the notebook is in dirty state. There's no damage on the document content because:

* We listen to metadata changes and update the InlineMetadata accordingly;
* The only place that I do believe where save immediately is required is when running/compiling a notebook, for this I added an dialog prompting the user. Later this behavior will be configurable, meaning that users can let Kale auto save the notebook if they want;
* We can implement "silent auto save"  in Kale, but please notice that JupyterLab has its own very well documented auto save feature in place and I do believe we should rely on that if users want to auto save their notebook  -> https://github.com/jupyterlab/jupyterlab/issues/16374#top

> It might reduce flexibility for cases where immediate save is required.

I don't see use case for auto saving the metadata because Jupyter Lab APIs keep track of the users changes and we can use that for validation or for building the UI. Please let me know if you think otherwise.

>  Maybe we should ensure that critical metadata updates are still persisted safely.

I do agree with that, we should have automated e2e tests that will ensure that the UI changes are reflected into the metadata read by Kale and that Kale is reading that correctly. I even mentioned this here -> https://github.com/kubeflow/kale/pull/660

Sayan4496:
Thanks @jesuino for the detailed explanation, that makes sense 👍
Relying on JupyterLab’s auto-save and prompting only during run/compile seems like a clean approach.

alikhere:
> @alikhere Great catch!
> 
> The issue was that when a notebook has only one cell, Jupyter Lab makes the cell active before we can register listeners to it, so we are not able to listen to metadata, so the changes in Cell metadata are ignored by Kale.
> 
> Would you please review it again?

@jesuino Now its working fine.

labextension/src/lib/NotebookUtils.tsx
```diff
@@ -105,6 +105,27 @@ export default class NotebookUtilities {
...
   }
 
+/**
+   * Opens a pop-up dialog in JupyterLab to display a simple message.
+   * @param title The title for the message popup
+   * @param msg The message as an array of strings
+   * @param buttonLabel The label to use for the button. Default is 'OK'
+   * @param buttonClassName The classname to give to the 'ok' button
+   * @returns Promise<void> - A promise once the message is closed.
+   */
```
ada333, line 115: why not use showYesNoDialog?

Image

jesuino: Thank you for this suggestion - I modified the code to use the Yes/No dialog.

ada333: thank you!

PR #683, opened 2026-03-11
fix: replace deprecated ast.Num, ast.Str, ast.NameConstant with ast.Constant

asmaa34937:
Fixes #682

Replace deprecated AST classes that were removed in Python 3.12+:
- ast.Num → ast.Constant
- ast.Str → ast.Constant  
- ast.NameConstant → ast.Constant

PR #716, opened 2026-03-19
Issue 683 remove deprecated code

### LINKED ISSUE #682: [backend] Remove Deprecated ast.[Classes]
I deployed Kale on Eder's branch `merge-single-package` following the instructions from #670.
Following the instructions, I was able to install and start the kale extension. The `titanic-ml-dataset/titanic_dataset_ml` example successfully compiled, uploaded, and started to run. However, the `base/candies_sharing` example fails to compile with the following error:

Image

Researching online, it seems that this is a known breaking change that occurs because the `ast.Num` class was deprecated in `3.8` and is not supported by later python verisons. I believe it will be the same for `ast.Str`. these values are both present in `kale/common/astutils.py`. We need to determine the best replacement and update the code to align with later versions of python.

hmtosi:
Closes #682
Includes the changes from #683 and also fixes the linting issues


jesuino (approved):

jesuino:
Hello @asmaa34937 

We are going to merge this PR with your commit. Your contribution is still here, but with the linting fix by Hannah. 

Please apologize for doing another PR, however, we must merge this for our Alpha release.


Thanks for understanding!

ederign (approved):

PR #751, opened 2026-04-10
Checking KFP version to add security context if supported

jesuino:
closes #721 

**Overview**
This PR adds dynamic detection of the Kubeflow Pipelines (KFP) server version to conditionally apply security context settings only when the KFP version supports it (v2.16+).

Key Changes

1. KFP Version Detection 
Added get_kfp_server_version(host) - fetches KFP version from the /healthz endpoint
Added supports_security_context(host) - returns True if KFP version > 2.16
Added get_kfp_host() - retrieves KFP host URL from client configuration

2. Conditional Security Context
The compiler now checks if the KFP server supports security context before generating pipelines
Template conditionally imports and applies security_context.set_security_context() only when supported

3. Code Refactoring 
Refactored get_ui_host() to use the new centralized kfputils.get_kfp_host() method

4. CI/CD Enhancement
Added kfp_version input parameter for workflow dispatch to enable testing with different KFP versions

5. Test Updates

Added mock for supports_security_context to ensure tests pass with security context enabled

**Why This Matters**

This change ensures backward compatibility with older KFP versions that don't support the security context feature, while still enabling it for newer versions.

### DIFF
--- a/kale/common/kfputils.py
+++ b/kale/common/kfputils.py
@@ -327,8 +327,12 @@ def load_mlpipeline_metrics(output):
 
 def get_kfp_host():
     """Get the KFP host URL from the client."""
-    client = _get_kfp_client()
-    return getattr(client, "_uihost", None) or getattr(client, "host", None)
+    try:
+        client = _get_kfp_client()
+        return getattr(client, "_uihost", None) or getattr(client, "host", None)
+    except Exception as e:
+        log.warning("Failed to get KFP host: %s", e)
+        return None
 
 
 def get_experiment_from_run_id(run_id: str):
--- a/kale/common/kfputils.py
+++ b/kale/common/kfputils.py
@@ -65,7 +65,7 @@ def get_kfp_server_version(host: str) -> str | None:
         response.raise_for_status()
         data = response.json()
         return data.get("apiServerTagName")
-    except (requests.RequestException, ValueError, KeyError) as e:
+    except Exception as e:
         log.warning("Failed to fetch KFP server version from %s: %s", healthz_url, e)
         return None
 
--- a/kale/common/kfputils.py
+++ b/kale/common/kfputils.py
@@ -25,6 +25,8 @@ from typing import Any
 
 import kfp
 from kfp_server_api.exceptions import ApiException
+from packaging.version import Version
+import requests
 
 from kale.common import utils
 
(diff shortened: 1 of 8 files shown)


kale/common/kfputils.py
```diff
@@ -44,6 +46,56 @@ def _get_kfp_client(host=None, namespace: str = "kubeflow"):
...
+
+
+def supports_security_context(host: str) -> bool:
+    """Check if the KFP server supports security_context (version > 2.16).
+
+    Args:
+        host: The KFP host URL
+
+    Returns:
+        True if KFP version is greater than 2.16, False otherwise
```
hmtosi, line 80: I would change this to:
True if KFP version >= 2.16 or cannot be determined, False if KFP version < 2.16

hmtosi (changes requested):
@jesuino great work so far, I'm still testing on kfp 2.15. In the meantime, just one comment to clarify the documentation

hmtosi (commented):
@jesuino I have finished testing.

**2.16.0**
The code works as expected on `kfp 2.16`. The DSL compiles successfully, it contains the security context and shows `kfp = 2.16`, and the pipelines run.

**2.15.x**
I was able to get the code to work on `kfp 2.15` as well, but it took some adjusting. To set up, I installed kfp using their [recommended steps](https://www.kubeflow.org/docs/components/pipelines/operator-guides/installation/) I used the `platform-agnostic config`, and `PIPELINE_VERSION=2.15.0`. I adjusted `pyproject.toml` to have `kfp[kubernetes]==2.15.2` instead of `2.16.0`. 

However, the kfp image provided by this method has `"apiServerCommitHash":"unknown", "apiServerTagName":"unknown"` which meant  the kfp host sees the version as `Unknown`. Because this code gets the version from the kfp host, this will return `unknown` as well, which the code is defaulting to `supports_security_context ==True`. This was causing the DSL not to compile, because it was looking for the `security_context` field, but that doesn't exist in this version of pipelines. (This could have been caused by the `2.15.0<>2.15.2` mismatch in my code - I can test further.) To finish testing, I hardcoded `kfp==2.15.2` in `kfputils.py`.

Once the proper version was assigned, the DSL compiled successfully, it showed `kfp = 2.15` and did not contain the security context, and the pipelines ran successfully.

**Final Thoughts**
These changes work as expected and look ok to me. However, we should discuss if this is indeed the default behavior we want for unknown pipeline versions

kale/common/kfputils.py
```diff
@@ -44,6 +46,56 @@ def _get_kfp_client(host=None, namespace: str = "kubeflow"):
...
+    if not version_str:
+        # If we can't determine the version, assume it's supported to avoid blocking users with newer versions
+        return True
+
+    try:
+        # Handle version strings that may have prefixes like 'v' or suffixes
+        clean_version = version_str.lstrip("v").split("-")[0]
+        kfp_version = Version(clean_version)
+        min_version = Version("2.16")
+        return kfp_version > min_version
```
ada333, line 92: Agree with Hannah - kfp_version >= min_version

ada333 (changes requested):
@jesuino looks good thanks - only one comment but it is stoping me from testing it out on my cluster since I have 2.16.0 kfp.

jesuino:
Thank you @hmtosi and @ada333 

I am going ahead and closing this because this was discussed that we are going to continue supporting KFP >= 2.16.0 hence this will not make sense. ([community discussion](https://app.slack.com/client/T08PSQ7BQ/C08KJBVDH5H))

THanks!

PR #753, opened 2026-04-13
feat(dev): add lightweight KFP local dev cluster via k3d

cordeirops:
## Summary

Running a full Kubeflow Pipelines stack for local development currently requires minikube (or equivalent), `kubectl port-forward`, and enough memory to host both the Kubernetes control plane (~2 GB) and the KFP services (~2 GB). This makes it impractical for contributors with resource-constrained machines or those who only need to work on Kale's frontend or backend — not on actual pipeline execution.

This PR introduces a **k3d-based local development cluster** as a lightweight, one-command alternative.

[k3d](https://k3d.io) runs [k3s](https://k3s.io) (a certified, minimal Kubernetes distribution) inside Docker containers. Its control plane uses an embedded SQLite store instead of etcd, reducing cluster overhead from ~2 GB (minikube) to ~512 MB, bringing the total footprint for a full KFP stack down to approximately **2.5 GB**.

### `scripts/kfp-dev-setup.sh` _(new)_

An idempotent Bash script that handles the full first-time setup:

1. Verifies `docker` and `kubectl` are available
2. Installs `k3d` automatically via its official install script if not present
3. Creates a k3d cluster named `kale-kfp` (Traefik ingress disabled to save memory)
4. Deploys KFP v2.16 using the official `platform-agnostic` kustomize manifests from `github.com/kubeflow/pipelines`
5. Waits for all KFP pods to reach `Ready` state
6. Starts a `kubectl port-forward` in the background, tracking its PID in `.kfp-dev-pf.pid` for clean teardown
7. Smoke-tests the UI endpoint and prints the next steps

The script is safe to re-run — it skips cluster creation and KFP deployment if they already exist.

### `Makefile` _(updated)_

Five new targets under a dedicated **"KFP Local Cluster"** section:

| Target | Description |
|---|---|
| `make kfp-dev-setup` | First-time cluster creation and KFP deployment (~5 min) |
| `make kfp-dev-start` | Daily driver: start the cluster and port-forward UI to `localhost:8080` |
| `make kfp-dev-stop` | Stop the port-forward and pause the cluster (all pipeline data is preserved) |
| `make kfp-dev-delete` | Wipe the cluster entirely and free all resources |
| `make kfp-dev-status` | Show k3d cluster state and KFP pod status |

All targets are configurable via Makefile variables (`KFP_CLUSTER_NAME`, `KFP_PIPELINE_VERSION`, `KFP_LOCAL_PORT`).

The `clean` target is updated to also remove the port-forward PID file.

## Developer workflow

```bash
# First time only (~5 min)
make kfp-dev-setup

# Every day
make kfp-dev-start
make kfp-run NB=examples/base/candies_sharing.ipynb KFP_HOST=http://localhost:8080

# End of day
make kfp-dev-stop

# When done with the cluster entirely
make kfp-dev-delete
```

## Requirements

- Docker (Docker Desktop on macOS/Windows, or Docker Engine on Linux)
- `kubectl` (available via `brew install kubectl` or bundled with Docker Desktop)
- k3d — installed automatically by `kfp-dev-setup` if not present

## Resource comparison

| Setup | Cluster overhead | Total (cluster + KFP) |
|---|---|---|
| minikube + KFP | ~2 GB | ~4–6 GB |
| **k3d + KFP (this PR)** | **~512 MB** | **~2.5 GB** |

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
         lint lint-backend lint-labextension format-labextension format-backend \
         build \
         kfp-build kfp-serve kfp-compile kfp-run \
-        kfp-dev-setup kfp-dev-start kfp-dev-stop kfp-dev-delete kfp-dev-status \
+        kfp-dev-setup kfp-dev-start kfp-dev-stop kfp-dev-delete kfp-dev-status kfp-dev-upgrade \
         clean clean-venv lock lock-upgrade check-uv \
         jupyter jupyter-kfp watch-labextension \
         docker-build docker-run \
@@ -157,13 +157,18 @@ KFP_LOCAL_PORT ?= 8080
 KFP_PID_FILE := $(CURDIR)/.kfp-dev-pf.pid
 
 kfp-dev-setup: ## Create k3d cluster + install KFP standalone (~5 min, first time only)
-	@bash scripts/kfp-dev-setup.sh "$(KFP_CLUSTER_NAME)" "$(KFP_PIPELINE_VERSION)" "$(KFP_LOCAL_PORT)" "$(KFP_PID_FILE)"
+	@bash scripts/kfp-dev-setup.sh setup "$(KFP_CLUSTER_NAME)" "$(KFP_PIPELINE_VERSION)" "$(KFP_LOCAL_PORT)" "$(KFP_PID_FILE)"
+
+kfp-dev-upgrade: ## Upgrade KFP on existing cluster (usage: make kfp-dev-upgrade KFP_PIPELINE_VERSION=2.17.0)
+	@bash scripts/kfp-dev-setup.sh upgrade "$(KFP_CLUSTER_NAME)" "$(KFP_PIPELINE_VERSION)" "$(KFP_LOCAL_PORT)" "$(KFP_PID_FILE)"
 
 kfp-dev-start: ## Start existing cluster and port-forward KFP UI to localhost:8080
 	@printf "$(BLUE)Starting k3d cluster '$(KFP_CLUSTER_NAME)'...\n$(NC)"
 	@k3d cluster start $(KFP_CLUSTER_NAME) 2>/dev/null || { \
 		printf "$(YELLOW)Cluster '$(KFP_CLUSTER_NAME)' not found. Run 'make kfp-dev-setup' first.\n$(NC)"; exit 1; \
 	}
+	@printf "$(BLUE)Switching kubectl context to k3d-$(KFP_CLUSTER_NAME)...\n$(NC)"
+	@kubectl config use-context k3d-$(KFP_CLUSTER_NAME)
 	@if [ -f $(KFP_PID_FILE) ]; then \
 		kill $$(cat $(KFP_PID_FILE)) 2>/dev/null || true; \
 		rm -f $(KFP_PID_FILE); \
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
         lint lint-backend lint-labextension format-labextension format-backend \
         build \
         kfp-build kfp-serve kfp-compile kfp-run \
+        kfp-dev-setup kfp-dev-start kfp-dev-stop kfp-dev-delete kfp-dev-status \
         clean clean-venv lock lock-upgrade check-uv \
         jupyter jupyter-kfp watch-labextension \
         docker-build docker-run \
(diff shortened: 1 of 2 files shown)


scripts/kfp-dev-setup.sh
```diff
@@ -0,0 +1,188 @@
...
+    ok "Cluster '${CLUSTER_NAME}' created"
+}
+
+# ---------------------------------------------------------------------------
+# KFP deployment
+# ---------------------------------------------------------------------------
+
+deploy_kfp() {
+    # Idempotent: if the kubeflow namespace already has the ML pipeline CRD, skip
+    if kubectl get namespace kubeflow >/dev/null 2>&1 && \
```
ada333: I would do this differently so KFP can be upgraded on existing cluster without the need to delete it - or maybe a new function like upgrade_kfp can be added?

scripts/kfp-dev-setup.sh
```diff
@@ -0,0 +1,188 @@
...
+
+# ---------------------------------------------------------------------------
+# Cluster
+# ---------------------------------------------------------------------------
+
+create_cluster() {
+    if k3d cluster list 2>/dev/null | grep -q "^${CLUSTER_NAME}[[:space:]]"; then
+        warn "Cluster '${CLUSTER_NAME}' already exists — skipping creation."
+        info "Starting cluster in case it was stopped..."
+        k3d cluster start "${CLUSTER_NAME}"
```
ada333, line 77: before this we should switch kubectl context - something like
`kubectl config use-context k3d-kale-kfp`

Makefile
```diff
@@ -148,6 +149,58 @@ kfp-run: ## Compile and run on KFP with local wheel (usage: make kfp-run NB=...
...
+KFP_CLUSTER_NAME ?= kale-kfp
+KFP_PIPELINE_VERSION ?= 2.16.0
+KFP_LOCAL_PORT ?= 8080
+KFP_PID_FILE := $(CURDIR)/.kfp-dev-pf.pid
+
+kfp-dev-setup: ## Create k3d cluster + install KFP standalone (~5 min, first time only)
+	@bash scripts/kfp-dev-setup.sh "$(KFP_CLUSTER_NAME)" "$(KFP_PIPELINE_VERSION)" "$(KFP_LOCAL_PORT)" "$(KFP_PID_FILE)"
+
+kfp-dev-start: ## Start existing cluster and port-forward KFP UI to localhost:8080
+	@printf "$(BLUE)Starting k3d cluster '$(KFP_CLUSTER_NAME)'...\n$(NC)"
```
ada333, line 166: also here: `kubectl config use-context k3d-kale-kfp`

Makefile
```diff
@@ -148,6 +149,58 @@ kfp-run: ## Compile and run on KFP with local wheel (usage: make kfp-run NB=...
...
+	@if [ -f $(KFP_PID_FILE) ]; then \
+		kill $$(cat $(KFP_PID_FILE)) 2>/dev/null || true; \
+		rm -f $(KFP_PID_FILE); \
+	fi
+	@k3d cluster delete $(KFP_CLUSTER_NAME) 2>/dev/null || true
+	@printf "$(GREEN)Cluster deleted. Run 'make kfp-dev-setup' to start fresh.\n$(NC)"
+
+kfp-dev-status: ## Show cluster and KFP pod status
+	@printf "$(BLUE)k3d clusters:\n$(NC)"
+	@k3d cluster list 2>/dev/null || printf "$(YELLOW)k3d not installed\n$(NC)"
```
ada333, line 205: also here: `kubectl config use-context k3d-kale-kfp`

ada333 (changes requested):
Hi, @cordeirops I really like this idea - thank you for making this PR!
I left a few suggestions.

cordeirops:
Thanks for the thorough review, @ada333! I've addressed all three suggestions in the latest commit ([933ff5b](https://github.com/cordeirops/kale/commit/933ff5b)).

---

### Changes made

**1. `kubectl config use-context k3d-{cluster}` — two locations**

Added an explicit context switch in both places flagged:

- In `create_cluster()` (script) — after cluster creation or start, kubectl is immediately pointed at `k3d-${CLUSTER_NAME}` before any `kubectl apply` or `kubectl wait` calls. This prevents commands from silently targeting a different cluster (e.g. a leftover minikube context).
- In `kfp-dev-start` (Makefile) — same switch added right after `k3d cluster start`, so the daily-driver workflow is also safe for users who switch between multiple clusters.

**2. KFP upgrade without deleting the cluster**

Refactored `deploy_kfp()` by extracting the shared manifest-apply logic into a private helper `_apply_kfp_manifests()`. Two distinct functions now exist:

- `deploy_kfp()` — unchanged behaviour: skips if KFP is already present, avoids re-applying on every setup run.
- `upgrade_kfp()` — always calls `_apply_kfp_manifests()`, allowing `kubectl apply -k` to reconcile the diff between the currently installed version and the target version in-place, preserving all experiment and run history.

A new `make kfp-dev-upgrade` Makefile target exposes this:

```bash
# Upgrade to a newer KFP version on an existing cluster
make kfp-dev-upgrade KFP_PIPELINE_VERSION=2.17.0
```

---

Let me know if anything else needs adjusting!

cordeirops:
> **Build failure (browser_check timeout)**

The `build` job failed with a `page.waitForSelector: Timeout 100000ms exceeded` error in `uv run python -m jupyterlab.browser_check`. This step spins up a headless browser to verify JupyterLab loads, it is unrelated to the changes in this PR (Makefile + bash script, no frontend code touched).

The same workflow passed on PR #754 which ran earlier the same day. This is a known flaky behaviour in CI environments when the runner is under resource pressure.

I've pushed an empty commit to re-trigger the workflow.

ada333 (approved):

jesuino (changes requested):
Hello @cordeirops 

Thanks for this! k3d sounds like a good lightweight alternative for running Kale in dev mode.

It seems to be working well, I did start my cluster and I can access KFP, but Kale does not seem to connect from my jupyterlab:

Image


I see this in Kale logs:
```
urllib3.exceptions.NewConnectionError: HTTPConnection(host='localhost', port=80): Failed to establish a new connection: [Errno 111] Connection refused
```

Is any other configuration required to run this with Jupyter Lab?

Thanks!

cordeirops:
Hello @jesuino, sorry for the late reply. I don't recall needing any extra configuration for this to work. I'll investigate the cause and get back to you here.

cordeirops:
Hi @jesuino!

The `Connection refused` on port 80 usually means the port-forward wasn't active when JupyterLab started. Kale uses the default KFP port with no extra configuration needed, but the port-forward process needs to be running in the background first.

The full workflow that works for me:

```bash
# 1. First-time setup (only needed once)
make kfp-dev-setup

# 2. Start the cluster and port-forward
make kfp-dev-start

# 3. In the same terminal, start the dev environment and JupyterLab
make dev
make jupyter
```

The key is to keep that terminal open. Closing it kills the port-forward, and Kale loses the connection to KFP.

Let me know if that helps!

ederign:
@jesuino would you mind to take a look at this?

jesuino (approved):
Hello,

PLease aplogize for the delay to provide a response for this.

I did have issues with k3d and it took me a while to fix. While I can't consistently use this tool locally on my Fedora, I do understand that the problems I am facing are not related to Pedro's change, hence I am going ahead and approving this PR.

Thanks!

ederign (approved):

Issue #749, opened 2026-04-09
[frontend] Non-existent RPC call throws warning in web console


ada333:
When running Kale locally there are always these two logs in web console:

<img width="1205" height="632" alt="Image" src="https://github.com/user-attachments/assets/7ae7a3a0-407c-4f0b-ae56-e50d12e54a82" />

The first one should work when running jupyter in Kubeflow - but I would still make it only a warning, not an error and maybe there should be some better message.
The second one is worse because it is a RPC call to backend for function that does exist. 

<img width="1205" height="632" alt="Image" src="https://github.com/user-attachments/assets/568ce74b-e9c6-4546-b1e8-fb4ee1efe2a7" />
We should either delete this call or implement the backend function. I would implement the function - it is meant to retrieve the Docker image that the current Jupyter Notebook Server is running on . This could be useful -- but there are #737 and #574 so we need to be careful to think it through. 

---

Impacted by this bug? Give it a 👍.

PR #754, opened 2026-04-13
Fix log in web console and delete non existing RPC call to backend

ada333:
this PR solves #749  - 

1. deletes non-existent RPC getBaseImage call that threw warning everytime Kale was turned on
2. turns error to warning when there is no Kubeflow namespace found, and makes better log message
3. makes all the RPC command logs only in debug mode -> this was filling up console with nonsense logs

Before:
Screenshot From 2026-04-09 13-39-08

After:
Screenshot From 2026-04-13 12-58-54


jesuino (approved):
LGTM!

ederign (approved):

PR #761, opened 2026-04-15
Remove deprecated `block:` cell tag

StefanoFioravanzo:
### Summary
  Drops the long-deprecated `block:` cell tag from Kale's notebook grammar and cleans up the labextension's internal naming to match. The `block:` tag was superseded by `step:` a while back and kept only as a compatibility alias — this
  branch finishes the migration.

  ### Changes

  **Commit 1 — `Remove deprecated block: cell tag`** (`7798229`)
  - `kale/processors/nbprocessor.py`: drop `BLOCK_TAG` regex and its entry in `_TAGS_LANGUAGE`; reword the `prev` validation error to mention only `step`.
  - `labextension/src/lib/CellUtils.ts`: fix `getStepName`, which was still stripping `block:` instead of `step:` (a leftover that silently broke step-name lookup for the exploration feature).
  - `examples/openvaccine-kaggle-competition/open-vaccine.ipynb`: replace the one stray `"block:"` tag with `"step:"`.

  **Commit 2 — `Rename block identifiers to step in labextension`** (`95ceec4`)
  Pure mechanical rename across 5 files, no behavioural change:
  - `blockName` → `stepName`, `prevBlockNames` → `prevStepNames`
  - `getAllBlocks` → `getAllSteps`, `getPreviousBlock` → `getPreviousStep`
  - `BlockDependencyChoice` → `StepDependencyChoice`, `blockDependenciesChoices` → `stepDependenciesChoices`
  - `updateCurrentBlockName` → `updateCurrentStepName`, `updatePrevBlocksNames` → `updatePrevStepsNames`
  - HTML id `select-previous-blocks` → `select-previous-steps`
  - Updated surrounding doc-comments ("block name" → "step name", etc.)

  Kept as a separate commit so the first one stays easy to revert if needed.

  ### Why split
  The labextension had already migrated its read/write paths to `step:` (`getKaleCellTags` only parses `step:`, `setKaleCellTags` only writes `step:`) — only the grammar entry, one leftover prefix strip, one example notebook, and the
  internal identifier naming remained. Rather than bundle a pure rename with an actual grammar change, the rename lives in its own commit so the deprecation removal can be reviewed (and reverted) independently.

### Breaking change
  Notebooks that still carry `block:<name>` tags will no longer be recognised as pipeline steps. Users should re-tag those cells as `step:<name>`. All example notebooks shipped in `examples/` have been checked; only `open-vaccine.ipynb`
  had a stray occurrence and it's fixed in this PR.

### DIFF
--- a/examples/openvaccine-kaggle-competition/open-vaccine.ipynb
+++ b/examples/openvaccine-kaggle-competition/open-vaccine.ipynb
@@ -672,7 +672,7 @@
    "execution_count": null,
    "metadata": {
     "tags": [
-     "block:"
+     "step:"
     ]
    },
    "outputs": [],
--- a/kale/processors/nbprocessor.py
+++ b/kale/processors/nbprocessor.py
@@ -33,13 +33,7 @@ SKIP_TAG = r"^skip$"
 IMPORT_TAG = r"^imports$"
 FUNCTIONS_TAG = r"^functions$"
 PREV_TAG = r"^prev:[_a-z]([_a-z0-9]*)?$"
-# `step` has the same functionality as `block` and is
-# supposed to be the new name
 STEP_TAG = r"^step:([_a-z]([_a-z0-9]*)?)?$"
-# Extension may end up with 'block:' as a tag. We handle
-# that as if it was empty.
-# TODO: Deprecate `block` tag in future release
-BLOCK_TAG = r"^block:([_a-z]([_a-z0-9]*)?)?$"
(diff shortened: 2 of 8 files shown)


ederign:
@StefanoFioravanzo we need to fix the lint for this one.

jesuino (approved):
LGTM!

Made the basic examples and created one from scratch and it worked correctly!

StefanoFioravanzo:
@ederign linting fixed!

ederign (approved):

PR #763, opened 2026-04-15
ISSUE-715: Validate pipeline name length in UI (max 124 characters)

### LINKED ISSUE #715: [feature] Add character length to pipeline name validation
### Feature Area

/area frontend

Currently the pipeline name is validated on the UI for certain characters, but it is not validated for character length. We need to adjust it so that it limits the user to 124 or fewer characters.

Image

Providing a pipeline name of 125 or more characters leads to runtime errors as shown in #695 

Yes, the workaround is entering a pipeline name with fewer characters. If too many characters are entered, a clear error message is displayed. However, it would be better to validate this on the UI field rather than in the pipeline run.

cordeirops:
## Summary

Fixes #715.

Pipeline names longer than 124 characters cause a runtime error in KFP. Previously there was no UI-level validation, so users only discovered the limit after a failed pipeline run with a cryptic error.

### `labextension/src/components/Input.tsx`
- Added optional `maxLength?: number` and `maxLengthErrorMsg?: string` props to `IInputProps`
- When `value.length > maxLength`, the field enters error state: label, input text, and helper text all render in red
- `getErrorMessage()` surfaces the length error with priority over regex errors, since it is the more actionable fix
- Soft-validation approach (show error, do not hard-block input) is intentional — consistent with existing regex validation and allows paste-then-edit workflows

### `labextension/src/widgets/LeftPanel.tsx`
- Pass `maxLength={124}` and `maxLengthErrorMsg` to the Pipeline Name field
- Extended `pipelineNameValid` to include `length <= 124` so the compile/run button is disabled for over-limit values — this also covers names pre-populated from existing notebook metadata that were saved before this validation existed

## Behaviour

| Scenario | Before | After |
|---|---|---|
| Typing a name > 124 chars | No feedback | Error state on field, button disabled |
| Pasting a name > 124 chars | No feedback | Error state on field, button disabled |
| Opening notebook with saved name > 124 chars | No feedback until runtime error | Error state on field, button disabled |
| Submitting pipeline with valid name | Works | Unchanged |

Screenshot 2026-04-15 at 10 16 43

## Test plan

- [ ] Type a pipeline name longer than 124 characters — field shows red label, red text, red error message, compile button disabled
- [ ] Reduce the name to ≤ 124 characters — error clears, button re-enables
- [ ] Paste a 200-character string — same error behaviour as typing
- [ ] Open a notebook whose metadata contains a `pipeline_name` longer than 124 characters — error shown on load
- [ ] Existing regex validation (invalid characters) still works as before

### DIFF
--- a/labextension/src/widgets/LeftPanel.tsx
+++ b/labextension/src/widgets/LeftPanel.tsx
@@ -42,6 +42,7 @@ const KFP_STATUS_REFRESH_MS = 30_000;
 
 const KALE_NOTEBOOK_METADATA_KEY = 'kubeflow_notebook';
 const DEFAULT_UI_URL = 'http://localhost:8080';
+const PIPELINE_NAME_MAX_LENGTH = 124;
 
 export interface IExperiment {
   id: string;
@@ -654,9 +655,10 @@ export class KubeflowKaleLeftPanel extends React.Component<IProps, IState> {
         experimentInputValue = selectedExperiments[0].name;
       }
     }
-    const pipelineNameValid = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/.test(
-      this.state.metadata.pipeline_name,
-    );
+    const pipelineNameValid =
+      /^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/.test(
+        this.state.metadata.pipeline_name,
+      ) && this.state.metadata.pipeline_name.length <= PIPELINE_NAME_MAX_LENGTH;
     const experimentNameRegex = /^[a-z]([-a-z0-9]*[a-z0-9])?$/;
     const experimentNameValid =
       experimentInputSelected !== NEW_EXPERIMENT.id ||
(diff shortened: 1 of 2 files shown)


labextension/src/widgets/LeftPanel.tsx
```diff
@@ -654,9 +654,10 @@ export class KubeflowKaleLeftPanel extends React.Component<IProps, IState> {
         experimentInputValue = selectedExperiments[0].name;
       }
     }
-    const pipelineNameValid = /^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/.test(
-      this.state.metadata.pipeline_name,
-    );
+    const pipelineNameValid =
+      /^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/.test(
+        this.state.metadata.pipeline_name,
+      ) && this.state.metadata.pipeline_name.length <= 124;
```
ada333: I would make the 124 as a constant

ada333 (commented):
I tried it out and works good, thanks!
I have only one suggestion otherwise LGTM

jesuino (approved):
LGTM!

Tested locally and works as expected. 

THanks again for your contribution, @cordeirops

labextension/src/widgets/LeftPanel.tsx
cordeirops: Can you take another look at this? I just implemented it in the last commit

cordeirops:
Thanks @ada333 and @jesuino!

ada333 (approved):

ederign (approved):
/approve
/lgtm

ederign:
thanks @cordeirops and sorry for the delay! we were cutting 2.0 release.

PR #764, opened 2026-04-16
ISSUE-759: Add labels and annotations into DSL

ada333:
This PR:

- adds labels in the nbprocessor.py
- adds annotations and labels in jinja template

DSL now correctly registers labels and annotations:
image


jesuino (approved):
LGTM!

PR #765, opened 2026-04-17
chore(labextension): remove commented-out MUI v4 blocks

### LINKED ISSUE #502: [frontend] Remove unused and deprecated code from labextension
### Environment

*  How did you deploy Kale?

 Local deploy

2.0 (unreleased)

### Steps to reproduce

Using VS Code `Find All References`  and other tools we can find code that is not being and also commented code that should be removed. Example: `CellUtils.runAndDelete` is not used

Image

### Expected result

Unused, deprecated or commented code should be removed.

### Materials and Reference

N/A

---

ai-naymul:
Fixes #502.

Each of Select, SelectMulti, Input, ExperimentInput, and AdvancedSettings has a large block of the old MUI v4 makeStyles implementation commented out at the top of the file, preserved as reference when the migration to MUI v7 styled components happened. Removing them now, the live code below each block is what's actually in use.

about 530 lines across 5 files, pure deletion.

cc @jesuino @ederign @StefanoFioravanzo @ada333


ada333 (approved):
LGTM - but just to be safe we need to agree that this code will never be useful

ai-naymul:
@ada333 agreed. each block is the pre-migration MUI v4 makeStyles version of the same component, and the live MUI v7 styled() code directly below each block is the replacement. the old code also imports from @material-ui/core which isn't a dependency anymore.

git history keeps the reference if anyone ever wants to look at it.

cc @jesuino @StefanoFioravanzo @ederign  for a second look before merge.

ederign (approved):

ederign:
@ada333 we can always check git history if necessary.

PR #784, opened 2026-05-06
Adding a new RAG with Langchain example

jesuino:
Adding a RAG with langchain example. This was built for building a reusable "RAG file" so I can load locally on my LLM client and make it run scheduled on KFP.

This example is using almost all Kale features and doing a good test of data passing between steps and when working on this I did identify some improvements that could be done in Kale:

https://github.com/kubeflow/kale/issues/782
https://github.com/kubeflow/kale/issues/783

Once these two issues are implemented, we can revisit this example to make sure that the performance is improved as intended.

When running this example, make it sure that you have at least ~15g free on your cluster for it non cached run. It used pytorch and pars of langchain that are large pip packages and they are downloaded for each step execution, so takes a good amount of disk space and the whole pipeline takes around 30 minutes to run. Once the above issues are implemented, we will see a good performance improvement for this pipeline!

### DIFF
--- a/examples/rag/portable_rag_langchain.ipynb
+++ b/examples/rag/portable_rag_langchain.ipynb
@@ -1,0 +1,253 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "7b89e11d-1d54-47f2-905d-0d43e2367d4d",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "# RAG Example\n",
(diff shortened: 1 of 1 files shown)


ada333 (approved):
I really like this example - lot of Kale features in one example! I only read the notebook code though - I haven't tried to run it due to disk space issues on my cluster.

ederign (approved):
/approve
/lgtm

Issue #781, opened 2026-05-06
feature(backend): Make security context configurable by environment variables


jesuino:
### Feature Area

/area backend

### What feature would you like to see?

I would like to make the security context configurable using env variables

### What is the use case or pain point?

Currently the security context is added to all generated pipelines:

```
    security_context.set_security_context(
        task=my_task,
        run_as_user=65534,
        run_as_group=0,
        run_as_non_root=True
    )
```
However, users can't control what users are going to be used or either turn this off. This is a RFE to allow users to turn this on/off using system properties

### Is there a workaround currently?

* Generate the Pipeline
* Modify the DSL to adjust the security context as you want

---

Love this idea? Give it a 👍.

jesuino:
This was done in: https://github.com/kubeflow/kale/pull/788

PR #788, opened 2026-05-08
feat: Make security context configurable by environment variables

jesuino:
In this change we add Security Context configuration as a Jupyter Configuration and environment variables:

Source | Settings
-- | --
JupyterLab Settings | Settings > Advanced Settings Editor > Kale
Environment Variables | KALE_SECURITY_CONTEXT_ENABLED, KALE_SECURITY_CONTEXT_RUN_AS_USER, KALE_SECURITY_CONTEXT_RUN_AS_GROUP, KALE_SECURITY_CONTEXT_RUN_AS_NON_ROOT

<strong>Precedence</strong>: JupyterLab settings → env vars → defaults (enabled=true, run_as_user=65534, run_as_group=0, run_as_non_root=true)

This is how it looks like from Jupyter Lab settings:

image

A video showing it in action:

https://github.com/user-attachments/assets/fa1b1b0d-8fb8-4e39-9bd1-f9b5e77c7739

**This change was co-authored by Claude**


ada333 (changes requested):
@jesuino looks good on the first look, could you please fix the tests?

kale/common/utils.py
```diff
@@ -277,3 +277,45 @@ def compute_trusted_hosts() -> list[str]:
...
+
+
+def get_security_context_from_env() -> dict:
+    """Read security context configuration from KALE_ environment variables.
+
+    Environment variables:
+        KALE_SECURITY_CONTEXT_ENABLED:
+            Boolean-like flag (1/true/yes/on) to enable/disable security context.
+        KALE_SECURITY_CONTEXT_RUN_AS_USER:
+            Integer UID to run containers as.
```
hmtosi: small update to documentation - on this line and in the next section, could you please clarify... to run containers as what?

kale/pipeline.py
```diff
@@ -100,6 +100,19 @@ class KatibConfig(Config):
...
 
 
+class SecurityContextConfig(Config):
+    """Configuration for Kubernetes security context settings.
+
+    These settings control the security context applied to all pipeline steps.
+    Can be configured via JupyterLab settings or ``KALE_*`` environment variables.
+    """
+
+    enabled = Field(type=bool, default=True)
```
ada333, line 110: there are lot of the defaults (4 I think) it would be nice to have it in one place

kale/pipeline.py
```diff
@@ -194,6 +209,23 @@ def _set_marshal_path(self):
...
+        env_config = utils.get_security_context_from_env()
+
+        if self.security_context is None:
+            # No metadata provided - use env vars merged with defaults
+            self.security_context = SecurityContextConfig(**env_config)
+        elif env_config:
+            # Metadata exists but may be incomplete - env vars fill in missing
+            for key, value in env_config.items():
+                current = getattr(self.security_context, key, None)
+                if current is None:
```
ada333: I tested this and I believe the env vars in UI are never applied because security_context is always set to default values (from either settings default or widget default

ada333 (changes requested):
@jesuino thanks for working on this! I have 2 comments - one is just code duplicity, the other is about functionality which I found when testing this PR (I hope I did not test it wrong)

hmtosi:
Closes #781

kale/pipeline.py
jesuino: Thank you for finding this, you are right - I did refactor the code to make it sure that the env var is respected - please let me know if it works for you.

Thanks!

kale/common/utils.py
hmtosi: addressed, thank you

jesuino:
I am fixing an issue with the UI/Backend sync

/hold

jesuino:
@hmtosi @ada333 The PR is ready for review!

hmtosi (approved):
/lgtm

@jesuino thank you for this work, the code looks good to me. I did the following tests:
1. build kale and run it without adjusting security context - successful run, yaml includes security context
2. use settings menu and disable security context - successful run, yaml doesn't include security context (shows that settings correctly control security context)
3. use settings menu and reenable security context - successful run, yaml includes security context
4. open local `kale-settings.jupyterlab-settings` file and delete security context, export `KALE_SECURITY_CONTEXT_ENABLED=False` - successful run, doesn't include security context (this shows that the env vars also correctly control security context)

ada333 (approved):

ederign (changes requested):
William, some comments: 
- Looks like there is a circular import between `utils.py` and `pipeline.py`**;
- No error handling for invalid integer env vars i nsecurity_context.run_as_user = int(run_as_user
- In `widget.tsx`, when `jlSecurityContextSetting.user === undefined`, backend env var defaults are written to the JupyterLab settings store via `loadedSetting.set(...)

Would mind to take a look on it?

jesuino:
@ederign 

Thanks for the feedback.

> In widget.tsx, when jlSecurityContextSetting.user === undefined, backend env var defaults are written to the JupyterLab settings store via `loadedSetting.set(...)


Thanks for pointing this. We have to override the user values, otherwise the UI will not reflect the values in env vars. I did some research of alternative values, but still I could not find a feasible way to do this. In fact, Claude is saying that it is not possible:

```
this will always show schema defaults. That's the tradeoff - we can't change what the Settings UI shows without writing to user settings.
```

So I think we should keep this approach and document it? Or we can ignore the env var on the user settings and show the default values.

The other points were addressed, thanks!

ederign (approved):

PR #796, opened 2026-05-15
fix(backend): only set metrics flag on steps that have assigned metrics

### LINKED ISSUE #787: [backend] Metrics artifact is generated to all leaf nodes in DAG
Found on Kale 2.0
When creating example pipeline, I noticed that kale_metrics_artifact node was created for all the leaf nodes in generated DAG, even though they do not define any of the metrics, f.e.

Image

---

ai-naymul:
Fixes #787

image

cc @ederign @jesuino @ada333 

pls review this and let me know

### DIFF
--- a/kale/processors/nbprocessor.py
+++ b/kale/processors/nbprocessor.py
@@ -691,9 +691,9 @@ class NotebookProcessor:
                     )
                 )
                 anc_step.source.append(code)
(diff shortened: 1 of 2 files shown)


ada333 (approved):
LGTM! 
I expected the fix to be bigger but this is clean and looks like it works fine

ederign (approved):

PR #802, opened 2026-05-19
Default base image config

ada333:
This PR solves #727 

Adds env var (DEFAULT_BASE_IMAGE) and settings (with settings having priority) support of configuring default base image for all steps.

So now the resolution of image for each step should have this order:
per-step specified image > jupyterlab setting > env var > hardcoded default python 3.12

Also now in image dialog the default image should be visible:
image

Also I noticed some unused legacy code mainly in podutils.py and Commands.ts and took a chance to delete it since the functions were never called and I think we will not have use for them.

also I had to adjust gitignore because it was ignoring the lib/ folder in which we have lot of utils

### DIFF
--- a/labextension/schema/kale-settings.json
+++ b/labextension/schema/kale-settings.json
@@ -21,6 +21,7 @@
       "title": "Default base image",
       "description": "Default container image for pipeline steps. Leave empty to use the environment variable KALE_DEFAULT_BASE_IMAGE or the python:3.12.",
       "default": ""
+    },
     "securityContext": {
       "type": "object",
       "title": "Security Context",
--- a/labextension/schema/kale-settings.json
+++ b/labextension/schema/kale-settings.json
@@ -19,7 +19,7 @@
     "defaultBaseImage": {
       "type": "string",
       "title": "Default base image",
-      "description": "Default container image for pipeline steps. Leave empty to use the environment variableKALE_DEFAULT_BASE_IMAGE or the python:3.12.",
+      "description": "Default container image for pipeline steps. Leave empty to use the environment variable KALE_DEFAULT_BASE_IMAGE or the python:3.12.",
       "default": ""
     },
     "outputPath": {
--- a/labextension/schema/kale-settings.json
+++ b/labextension/schema/kale-settings.json
@@ -21,6 +21,42 @@
       "title": "Default base image",
       "description": "Default container image for pipeline steps. Leave empty to use the environment variable KALE_DEFAULT_BASE_IMAGE or the python:3.12.",
       "default": ""
+    "securityContext": {
+      "type": "object",
+      "title": "Security Context",
+      "description": "Kubernetes security context settings for pipeline steps. These settings control how containers run in the cluster.",
+      "default": {
+        "enabled": true,
+        "run_as_user": 65534,
+        "run_as_group": 0,
+        "run_as_non_root": true
+      },
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "title": "Enable Security Context",
+          "description": "When enabled, applies security context to all pipeline steps.",
+          "default": true
+        },
+        "run_as_user": {
+          "type": "integer",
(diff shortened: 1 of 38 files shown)


labextension/schema/kale-settings.json
```diff
@@ -16,6 +16,12 @@
       "description": "When enabled, Kale saves the notebook automatically before compile, run, or upload without prompting. When disabled, you are asked to save if there are unsaved changes.",
       "default": false
     },
+    "defaultBaseImage": {
+      "type": "string",
+      "title": "Default base image",
+      "description": "Default container image for pipeline steps. Leave empty to use the environment variableKALE_DEFAULT_BASE_IMAGE or the python:3.12.",
```
jesuino: missing space on `variableKALE_DEFAULT_BASE_IMAGE`

labextension/src/widgets/hooks/useNotebookLoader.ts
```diff
@@ -186,8 +183,10 @@ export function useNotebookLoader({
 
         const defaultPipelineName = getNotebookFileName(notebook);
         const sanitized = sanitizePipelineName(defaultPipelineName);
+        const { base_image: _legacyBaseImage, ...notebookMetadataRest } =
```
jesuino: Could you please remove this `_legacyBaseImage`

labextension/src/widgets/hooks/useNotebookMetadataPersistence.ts
```diff
@@ -40,7 +40,8 @@ export function useNotebookMetadataPersistence({
       prevMetadataJsonRef.current = json;
       const notebook = tracker.currentWidget;
       if (notebook) {
-        NotebookUtils.setMetaData(notebook, metadataKey, metadata);
+        const { base_image: _baseImage, ...metadataToPersist } = metadata;
```
jesuino, line 44: Could you please check this warning?

labextension/src/widgets/hooks/useDeployment.ts
```diff
@@ -157,9 +159,7 @@ export function useDeployment({
       JSON.stringify(metadataRef.current),
     );
 
-    if (metadata.base_image === '') {
-      metadata.base_image = DefaultState.metadata.base_image;
-    }
+    metadata.base_image = resolvedDefaultBaseImage;
```
jesuino, line 162: What if user had already a base_image in place, wouldn't this be overriding the user setting?

jesuino (changes requested):
Hello Adam, 

Thanks for your work on this. Here are a few suggested small changes. Please let me know what you think.

labextension/src/widgets/hooks/useDeployment.ts
ada333: I did this on purpose, I think we should not consider the default base image from metadata. 
It would be confusing for user if he sets some default image in the settings or by env var, but because he is using maybe an old notebook with base_image in metadata the settings or the env var never apply.

ada333:
@jesuino thanks for the review. I fixed most of the things and left one comment that should answer your review question.

jesuino (changes requested):
Hello @ada333 Thanks for the changes, would you please fix the merge conflicts?

One last thing is that I think that the `DEFAULT_BASE_IMAGE` from `constants.ts` is not being used anymore, would you please check it and if it is not used remove it?

Thanks!

ada333:
@jesuino thanks for review, I resolved the conflicts and deleted the constant.

labextension/src/widgets/hooks/useDeployment.ts
jesuino: I do understand your point: users will not be able to know what is the default image unless they check the config and not overriding it may mislead users.

In the near future we can think about showing the metadata configured image to users, but for now it sounds like an appropriate solution to override the notebook metadata image.

jesuino (approved):
LGTM!

PR #804, opened 2026-05-20
Fix dark mode styling for disabled GPU dialog

### LINKED ISSUE #785: [frontend] The disabled GPU dialog in Jupyter Lab dark mode looks weird
GOOD FIRST ISSUE:
When opening GPU dialog for a step in Jupyter Lab dark mode the dialog looks strange when the require GPU for this step is disabled:

Image

I would probably fix the text _GPU Count_ to have same formatting as GPU Type.

---

CodeVishal-17:
fixes #785 

## Summary

Fixes dark mode UI inconsistencies in the disabled GPU dialog.

### Changes

* Improved disabled input visibility in dark mode
* Added better contrast for disabled fields
* Unified styling between GPU Count and GPU Type fields
* Improved readability and overall dark mode consistency


ada333 (changes requested):
@CodeVishal-17 thanks for the changes!

the input fields look different:

Image

also there is code duplication since you are using the same style twice

Could you please try to fix the code changes so the input fields (on all dialogs) look the same and there is minimal code duplication?

CodeVishal-17:
Thanks for the review! I updated the implementation to reduce code duplication and aligned the disabled input styling more closely with the existing dialog styles.

ada333 (approved):

jesuino (approved):
Thanks, it is looking good to me now

Image

PR #811, opened 2026-05-27
feat(frontend): Add clear metadata button to cell metadata editor

### LINKED ISSUE #766: [bug] cannot restore cell to unlabeled state after labeling it
### Steps to reproduce
- Open JupyterNotebook
- enable Kale
- add a cell to the notebook
- click the pencil icon to edit
- choose a cell type from the drop-down and select it
- click on the pencil icon again --> no way to remove the label, can only choose a different one from the dropdown

This prevents the user from returning it to its "unlabeled" state, where it will just be combined into the step above it. The user is forced to make it a new, separate step.

### Expected result

User should be allowed to restore it to its unlabeled, default state

Sayan4496:
## Summary

This PR adds a clear metadata action to the Kale cell metadata editor, allowing users to quickly remove Kale-specific metadata from a notebook cell.

### UI Updates

* Added a new "Clear Metadata" button to `CellMetadataEditor.tsx`
* Used `LayersClearIcon` for better UX clarity
* Kept the existing close editor action separate from metadata clearing
* The button is only shown when the cell contains Kale metadata or reserved Kale cell types

### Hook Updates

* Added `clearCellMetadata()` to `useCellTags.ts`
* Exposed the action through the existing metadata update hook

### Utility Updates

* Added `removeAllKaleTags()` to `TagsUtils.ts`
* Removes only Kale-related tags while preserving unrelated/custom notebook tags

## Behavior

The new clear action removes:

* `step:` tags
* dependency (`prev:`) tags
* limit tags
* image tags
* cache tags
* reserved Kale metadata tags

Non-Kale/custom tags remain untouched.

## Why This Change

Previously, users had to manually reset or modify metadata field-by-field. This improvement provides a faster and more user-friendly way to completely clear Kale metadata from a cell.

## Testing

Tested locally by:

* adding metadata to cells
* clearing metadata using the new button
* verifying reserved Kale cells
* confirming editor close behavior still works
* ensuring unrelated notebook tags are preserved
* running eslint and pre-commit hooks successfully
## Screenshots

Screenshot 2026-05-27 183644

Closes #766

### DIFF
--- a/labextension/src/lib/TagsUtils.ts
+++ b/labextension/src/lib/TagsUtils.ts
@@ -261,10 +261,13 @@ export default class TagsUtils {
         'tags',
       ) || [];
 
-    const kaleTagPrefixes = KALE_TAG_PREFIXES;
+    // Extract the step name from 'step:<name>' tag before wiping
+    const stepTag = currentTags.find(tag => tag.startsWith('step:'));
+    const clearedStepName = stepTag ? stepTag.replace('step:', '') : null;
 
+    // Clear all Kale tags from the active cell
     const filteredTags = currentTags.filter(
-      tag => !kaleTagPrefixes.some(prefix => tag.startsWith(prefix)),
+      tag => !KALE_TAG_PREFIXES.some(prefix => tag.startsWith(prefix)),
     );
 
     CellUtils.setCellMetaData(
@@ -272,7 +275,13 @@ export default class TagsUtils {
       activeCellIndex,
       'tags',
       filteredTags,
-    );
+    ).then(() => {
+      // If this cell had a step name, remove all prev:<stepName> refs
+      // from dependent cells using the existing updateKaleCellsTags
+      if (clearedStepName) {
+        TagsUtils.updateKaleCellsTags(notebook, clearedStepName, '');
+      }
+    });
   }
 
   public static cellsToArray(notebook: NotebookPanel) {
--- a/labextension/src/lib/TagsUtils.ts
+++ b/labextension/src/lib/TagsUtils.ts
@@ -15,6 +15,7 @@
 import { Notebook, NotebookPanel } from '@jupyterlab/notebook';
 import CellUtils from './CellUtils';
 import { RESERVED_CELL_NAMES } from '../widgets/cell-metadata/CellMetadataEditor';
+import { KALE_TAG_PREFIXES } from '../widgets/cell-metadata/constants';
 
 const IMAGE_TAG = 'image:';
 const CACHE_TAG = 'cache:';
@@ -249,7 +250,7 @@ export default class TagsUtils {
       TagsUtils.updateKaleCellsTags(notebook, oldStepName, value);
     });
   }
-    public static removeAllKaleTags(
+  public static removeAllKaleTags(
     notebook: NotebookPanel,
     activeCellIndex: number,
   ) {
@@ -260,18 +261,7 @@ export default class TagsUtils {
         'tags',
       ) || [];
 
-    const kaleTagPrefixes = [
-      'step:',
-      'prev:',
-      'limit:',
-      'image:',
-      'cache:',
-      'skip',
-      'imports',
-      'functions',
-      'pipeline-parameters',
-      'final',
-    ];
+    const kaleTagPrefixes = KALE_TAG_PREFIXES;
 
     const filteredTags = currentTags.filter(
       tag => !kaleTagPrefixes.some(prefix => tag.startsWith(prefix)),
--- a/labextension/src/lib/TagsUtils.ts
+++ b/labextension/src/lib/TagsUtils.ts
@@ -249,6 +249,41 @@ export default class TagsUtils {
       TagsUtils.updateKaleCellsTags(notebook, oldStepName, value);
     });
   }
+    public static removeAllKaleTags(
+    notebook: NotebookPanel,
+    activeCellIndex: number,
+  ) {
+    const currentTags: string[] =
+      CellUtils.getCellMetaData(
+        notebook.content,
+        activeCellIndex,
+        'tags',
+      ) || [];
+
+    const kaleTagPrefixes = [
+      'step:',
+      'prev:',
+      'limit:',
+      'image:',
+      'cache:',
+      'skip',
+      'imports',
+      'functions',
+      'pipeline-parameters',
+      'final',
+    ];
+
+    const filteredTags = currentTags.filter(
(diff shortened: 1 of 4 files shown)


labextension/src/lib/TagsUtils.ts
```diff
@@ -249,6 +249,41 @@ export default class TagsUtils {
       TagsUtils.updateKaleCellsTags(notebook, oldStepName, value);
     });
   }
+    public static removeAllKaleTags(
```
ada333: I think there is indentation error here (2 spaces more)

labextension/src/lib/TagsUtils.ts
```diff
@@ -249,6 +249,41 @@ export default class TagsUtils {
...
+      'prev:',
+      'limit:',
+      'image:',
+      'cache:',
+      'skip',
+      'imports',
+      'functions',
+      'pipeline-parameters',
+      'final',
+    ];
```
ada333: It would be cool if this was added to constants.ts so it is not hardcoded here - please also check for other occurences of these and switch them to the constants.

ada333 (changes requested):
Hi @Sayan4496, it works great and code looks mostly good, thank you!
I left two comments for improvement. Is this PR connected to some issue?

labextension/src/lib/TagsUtils.ts
jesuino: TO add more to the Adam's comment: check the constant `CELL_TYPES` on `constants.ts`, you could use that here

jesuino (changes requested):
Thanks for this contribution, it looks like a good feature to have for Kale.

Overall it looks good. Please make the changes suggested by Adam.

Thanks!

Sayan4496:
Thanks @ada333 &@jesuino  for the feedback! I've fixed the indentation issue and updated the implementation to reuse constants from constants.ts instead of maintaining a separate hardcoded list.

labextension/src/widgets/cell-metadata/constants.ts
```diff
@@ -28,6 +28,18 @@ export const RESERVED_CELL_NAMES = [
...
+  'step:',
+  'prev:',
+  'limit:',
+  'image:',
+  'cache:',
+  'skip',
+  'imports',
+  'functions',
+  'pipeline-parameters',
+  'final',
```
jesuino: I didn't find `final` as a Kale metadata. Could you please confirm it? If it is not a Kale metadata, we should not have it here because it may accidentally remove user metadata.

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
```diff
@@ -246,7 +247,19 @@ export const CellMetadataEditor: React.FC<IProps> = props => {
               </>
             )}
 
-            <IconButton aria-label="delete" onMouseDown={closeEditor}>
+            {(hasStepName || RESERVED_CELL_NAMES.includes(stepName)) && (
```
jesuino, line 250: is `hasStepName` really necessary here?

labextension/src/widgets/cell-metadata/constants.ts
```diff
@@ -28,6 +28,18 @@ export const RESERVED_CELL_NAMES = [
   'pipeline-metrics',
   'skip',
 ];
+export const KALE_TAG_PREFIXES = [
+  'step:',
```
jesuino, line 32: I am wondering if we should update the dependent cells when a step dependency is removed. For example, let's say step A depends on B, so it will have `prev:B`. If we clear the metadata in B, A will be orphaned.

jesuino (changes requested):
Hello,

Thanks for the update. The functionality seems to be working as expected, but there are a few points that we should fix/improve.

Please let me know your thoughts!

ederign:
@Sayan4496 are you still interested in working on this? If so, can you please take a look on the @jesuino commennts?

Sayan4496:
Hi @ederign,

Yes, I'm still interested in working on this PR.

I had to pause working on it due to my university exams, which are still ongoing and will finish on July 10. Once they're over, I'll address @jesuino's review comments and push the requested updates.

Thanks for checking in and for your patience!

ederign:
@Sayan4496 for sure! Please ping us when it's ready!

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
Sayan4496: hasStepName is necessary because without it, the clear button only appears for reserved cell types (like pipeline-parameters, skip) but not for regular pipeline steps with a custom name. Both conditions are needed to cover all labeled cell states.

Sayan4496:
Hi @jesuino @ada333,

All review comments have been addressed:
- Removed `'final'` from `KALE_TAG_PREFIXES` (confirmed no Kale usage in the codebase)
- `removeAllKaleTags` now handles orphaned dependents by calling `updateKaleCellsTags` with `''` after clearing, so any `prev:<stepName>` refs in other cells are cleaned up automatically
- Fixed PR title to `feat(frontend):` — CI checks are now passing ✅

Ready for re-review! Thanks for your patience.

labextension/src/widgets/cell-metadata/constants.ts
```diff
@@ -28,6 +28,17 @@ export const RESERVED_CELL_NAMES = [
...
+  'prev:',
+  'limit:',
+  'image:',
+  'cache:',
+  'skip',
+  'imports',
+  'functions',
+  'pipeline-parameters',
```
ada333, line 40: pipeline-metrics is missing

ada333 (changes requested):
Hi @Sayan4496 , we are almost there - I think only thing that is missing is the pipeline-metrics, otherwise I tried it out and looks good!

labextension/src/widgets/cell-metadata/constants.ts
Sayan4496: Thanks @ada333! Added pipeline-metrics to KALE_TAG_PREFIXES. Ready for re-review!

ada333 (approved):

jesuino (approved):
Hello Sayan,

The code is looking good, thanks for your changes!

I found a behavior that I can reproduce without your feature (orphaned cells are not immediately cleaned, only when saving the notebook), but I will open a specific issue for that!

https://github.com/user-attachments/assets/c51cac9b-10b0-4408-976f-961a71f68cf5

PR #821, opened 2026-06-03
Add env vars for customization of run and upload links

ada333:
This PR adds two env vars - KALE_UPLOAD_LINK, KALE_RUN_LINK
If these env vars are set, their value will be used for upload and run links.
Currently in the Kale code there is KFP_UI_ENDPOINT env var and the links then are build KFP specific:
`${base}/#/pipelines/details/${pipeline.pipeline.pipelineid}/version/${pipeline.pipeline.versionid}`
this URL structure is KFP specific, which can be problem when using Kale somewhere else (RHOAI in our case). That is why this functionality was added. The links also have placeholders for dynamic/pipeline-specific values that any link could use.

Example use:
`export KALE_RUN_LINK="https://test.example.com/run/ns={namespace}/run={run_id}"`


labextension/src/widgets/deploys-progress/DeploysProgress.tsx
```diff
@@ -27,6 +27,11 @@ export type RunPipeline = {
   status: string | null;
 };
 
+export type CustomLinks = {
```
jesuino: This type seems to be a duplicate of `ICustomLinks`. Would you please centralize all in a single type?

labextension/src/widgets/hooks/useNotebookLoader.ts
```diff
@@ -99,6 +100,9 @@ export function useNotebookLoader({
       const host = (await commands.getKfpUiHost()) || DEFAULT_UI_URL;
       setKfpUiHost(host);
 
+      const links = await commands.getCustomLinks();
```
jesuino: I wonder if we could use a single `Promise.all` for both the custom links and `kfpUIHost` calls

labextension/src/widgets/hooks/useNotebookMetadata.ts
```diff
@@ -121,6 +132,7 @@ export function useNotebookMetadata({
     setIsEnabled(false);
     setNamespace('');
     setKfpUiHost('');
+    setCustomLinks({ upload: '', run: '' });
```
jesuino: I would call this a more specific name, such as `DeployPanelCustomLinks` so we can easily identify what these links are about.

jesuino (changes requested):
Hello @ada333 

Thanks for your contribution!

I am adding a few suggestions of improvements, would you please take a look?

Thanks!

ederign:
@ada333 when you have a chance, can you please check @jesuino review?

ada333:
@jesuino thanks for review, I added the required changes

kale/rpc/kfp.py
```diff
@@ -47,6 +51,22 @@ def get_ui_host(request):
...
+
+    Returns a dict with 'upload' and 'run' keys. Values are the URL
+    patterns if set, or empty strings if not configured.
+
+    Placeholders:
+    - Upload: {pipeline_id}, {version_id}, {namespace}
+    - Run: {run_id}, {namespace}
+    """
+    return {
+        "upload": os.environ.get(KALE_UPLOAD_LINK_ENV, ""),
```
jesuino: I noticed what we have not a single validation for these URLs

Would you please validate the env vars before using it? A small check would be enough, for example, checking if the URL scheme is `http[s]://` by making sure it starts with `http://` or `https://`

labextension/src/widgets/deploys-progress/DeployProgress.tsx
```diff
@@ -290,7 +323,7 @@ export const DeployProgress: React.FunctionComponent<
         </a>
       </React.Fragment>
     );
-    logLinksHint(props.kfpUiHost || '');
+    logLinksHint(props.kfpUiHost || '', props.deployPanelCustomLinks);
```
jesuino: I wonder if we should call this for every run - I think this could be printed only once after Kale is started, no need to print this for every run!

jesuino (changes requested):
Hello @ada333 

I see that it is working as expected! I have one last suggestion, which is add at least some basic validation, otherwise I can set any URL, including malicious values.

```
$ export KALE_RUN_LINK=/my/run/link
$ export KALE_UPLOAD_LINK=/my/upload/link
```

Image

Image

I also confirmed that the placeholders for the URLs are working well:

Image

ada333:
@jesuino good suggestions thanks, they should be implemented now.

jesuino (approved):
LGTM!

PR #823, opened 2026-06-03
Removing Default Security Context

### LINKED ISSUE #822: [frontend] Default security context blocks users from setting it properly
### Environment

*  How did you deploy Kale?

`pip install kubeflow-kale`

Kale 2.1

### Steps to reproduce

* Create a new Notebook that has no metadata on it
* Modify the Security Context information using the User Settings
* Notice that the Security context is unmodified 

### Expected result
The user settings should be respected when there's no security context on the notebook

The problem is that we set a default security context on `LeftPanelTypes.ts::DefaultState`. Removing the default security context should be the solution since we check for existing security context on the notebook metadata before setting it on `useDeployment.ts`

jesuino:
The default security context is preventing the user settings to be applied correctly.
We didn't see this before because the examples have a default metatada, which means that the default from the labextension is not being used.

How to test it:

* Clear the current user settings (usually the settings are in file `~/.jupyter/lab/user-settings/jupyterlab-kubeflow
-kale/kale-settings.jupyterlab-settings`)
* Before starting KFP, use a value for some of the security context setting env var, such as: `export KALE_SECURITY_CONTEXT_RUN_AS_USER=123`
* Open Jupyter Lab and check the Kale settings, make it sure that the env vars you set are reflect on the default settings
* Create a new notebook
* Enable Kale and add at least a single step
*  COmpile the notebook and look for the security context, it should have the same values you saw in User settings
* Modify the user settings and add a new value for some of the Security context fields. Open the compiled DSL and confirm that it has the correct value:

image

closes #822

### DIFF
--- a/labextension/src/widgets/LeftPanelTypes.ts
+++ b/labextension/src/widgets/LeftPanelTypes.ts
@@ -60,11 +60,5 @@ export const DefaultState = {
     base_image: '',
     enable_caching: true,
     steps_defaults: [] as string[],
-    security_context: {
-      enabled: true,
-      run_as_user: 65534,
-      run_as_group: 0,
-      run_as_non_root: true,
-    },
   } as IKaleNotebookMetadata,
 };


hmtosi (commented):
@jesuino thank you for doing this debug.

I followed the testing steps as you described, and they all passed:
- clear the user settings as explained in the PR
- run `export KALE_SECURITY_CONTEXT_RUN_AS_USER=123`
- open JupyterLab, settings show `Run As User 123`
- create a notebook, enable kale, and add two pipeline steps
- compile and run the notebook
     - The pipeline runs successfully
     - The compiled DSL contains security context with Run As User = 123
- modify user settings to Run As User = 1234, then compile and run again
     - The pipeline runs successfully
     - The compiled DSL contains security context with Run As User = 1234

Before I approve, I want to make sure I understand the intended functionality. After I completed the test above, I did one more test:
- in the JupyterLab terminal, run `export KALE_SECURITY_CONTEXT_RUN_AS_USER=12345`
- refresh JupyterLab
- check the user settings the value is still 1234
- compile the pipeline - the value is still 1234

It seems that the environment variables can only be used before JupyterLab is started, otherwise the user settings override it (hence why we had to delete the user settings). If that is ok, then I will approve the PR!

jesuino:
Hello @hmtosi 

Thank you so much for your quick review.

> It seems that the environment variables can only be used before JupyterLab is started,

That's correct. On the original PR where we provided this features we had the same discussion and a good solution would be either change the order of priority (making the env var always override user setting) or do a major refactor to transform the settings, so we decided to make the env vars have effect only when there's no user setting.

hmtosi (approved):
LGTM

@jesuino thank you for explaining, this looks good to me

ederign (approved):
/approve
/lgtm

PR #842, opened 2026-06-22
chore(frontend): remove DeploysProgress wrapper and simplify to single DeployProgress

### LINKED ISSUE #820: chore(frontend) DeploysProgress does not make sense
There is a component called DeploysProgress.tsx that goes through multiple DeployProgress components and renders them.
Right now we allow only 1 DeployProgress component to render -> in the useDeployment hook after every run/compile all deployments are cleared:
https://github.com/kubeflow/kale/blob/main/labextension/src/widgets/hooks/useDeployment.ts#L257

The component DeploysProgress was meant for keeping a list of DeployProgress components and displaying them, but since we allow only 1 to be displayed it is not useful anymore and the code would be cleaner without this component.

---

Bhavd33p:
Fixes #820.

## What changed and why

`DeploysProgress` was a list container that rendered multiple `DeployProgress`
cards via a `{ [index: number]: DeployProgressState }` dictionary. However,
`activateRunDeployState` in `useDeployment.ts` calls `setDeploys({})` before
every new run/compile, meaning there is never more than one active deploy at a
time. The dictionary, index counter, and wrapper component were unnecessary.

## Changes

- Delete `DeploysProgress.tsx`
- Move shared types (`DeployProgressState`, `UploadPipelineResp`, `RunPipeline`)
  into `DeployProgress.tsx` where they belong
- Simplify `useDeployment` — replace the indexed `deploys` map with a single
  `DeployProgressState | null`
- Render `DeployProgress` directly in `LeftPanel` with a simple null-check
- Fix stale import in `Commands.ts` that still referenced the old path

## Testing

- Built successfully with `jlpm build:lib && jlpm build:labextension:dev`
- Manually tested in JupyterLab — compiled a notebook, progress card appears
  and updates correctly, close button dismisses it


ada333 (changes requested):
@Bhavd33p the renaming makes sense, thanks.
But what I meant is that this whole component is useless now -> we allow only one DeployProgress component to be displayed - in the useDeployment hook after every run/compile all deployments are cleared:
https://github.com/kubeflow/kale/blob/main/labextension/src/widgets/hooks/useDeployment.ts#L257

so DeploysProgress component could be deleted, the LeftPanel would hold only 1 DeployProgress... let me know if you want to work on that as well, if not I will approve this renaming and do the rest in some other PR.

Bhavd33p:
@ada333 Yes , I will work on that in this PR only will Implement and commit the required changes .

ada333:
@Bhavd33p thanks for the changes. The code does not compile though - `DeploysProgress` is still used in `Commands.ts` , but that should be an easy fix. Otherwise the changes look good. Please before every change check that the code compiles (you can do `make dev` command f.e.) and try the change out manually in JupyterLab or in KFP, thanks! Also if you could change the title of this PR and its description.

Bhavd33p:
Hi @ada333 ,All changes are done as discussed please review when you get a chance, thanks.

ada333 (approved):
@Bhavd33p LGTM, thanks for you first PR! Lets wait what the OGs (@jesuino or @ederign) will say 😄

ederign (approved):
@ada333 ran it locally.

Issue #843, opened 2026-06-24
feature: add a way to mount volumes in pipeline steps


Fruchix:
### Feature Area

/area frontend
/area backend

### What feature would you like to see?

I would like to be able to mount Kubernetes volume(s) using Kale. It is currently only possible to do so inside of the generated pipeline files (either .py or .yaml).

The feature should add notebook metadatas and add components in Kale JupyterLab extension to be able to point and click to mount volumes. The user should be able to mount 0 to N volumes, not only 1.

### What is the use case or pain point?

When working on AI/ML, people tend to store their data in high performance storages. Those storages are then made available in Kubernetes as StorageClasses and PersistentVolumeClaims.

### Is there a workaround currently?

One workaround is to generate a python pipeline file from a Jupyter notebook, THEN to mount the pvc on each task that requires it:

```py
from kfp import kubernetes, dsl

@dsl.component
def task1() -> str:
    ...

@dsl.component
def task2() -> str:
    ...

@dsl.pipeline
def my_pipeline():
    t1 = task1()
    kubernetes.mount_pvc(t1, pvc_name="...", mount_path="...")
    
    t2 = task2()
    kubernetes.mount_pvc(t2, pvc_name="...", mount_path="...")
```

---

Love this idea? Give it a 👍.

ederign:
@Fruchix that is a valid feature. Do you think folks wants to mount pvc by pipeline right? So it will be in the sidebar similar to Pipeline Metadata. . Something like:

1. A "Volumes" header with an "Add Volume" button
2. Each volume entry would need just two fields: **PVC Name** and **Mount Path** (the essential ones from `VolumeConfig`)
3. A remove button per entry

Fruchix:
Hi @ederign, after a bit of thinking I believe the feature that I am requesting might only be a partial answer.

There might be two different things to implement:
- when converting a notebook to a kubeflow pipeline, **automatically mount the volumes that have been mounted to the notebook itself**
- in general, have the **possibility to mount any volume** (which is the original feature I requested, and I think the approach you suggested is perfect)

Why?  I think people that would want volumes mounted in pipelines would probably fall into one of the two following cases:

[...]

PR #847, opened 2026-06-25
docs: Add KEP-0812 Composable Kale Notebooks proposal

ederign:
## Summary

Adds the design proposal for KEP-0812: Composable Kale Notebooks.

This KEP is the result of collaborative work between @ederign, @StefanoFioravanzo, and @Ya-shh as part of the [GSoC 2026 program](https://www.kubeflow.org/events/upcoming-events/gsoc-2026/#project-11-composable-kale-notebooks-with-visual-pipeline-editor). Yash delivered a solid POC that validated the core technical approach — compiling each notebook as a KFP sub-pipeline (GraphComponent) with automatic boundary variable detection via AST/PyFlakes analysis. This KEP builds on those findings to define the full design.

### Key design decisions

- **Compilation**: Each notebook becomes a KFP sub-pipeline via GraphComponent, preserving cell-level step visibility in the KFP UI (validated in Yash's POC)
- **Interface declaration**: Automatic inference with zero friction — no new tags needed for MVP. Name collisions raise errors.
- **Composition format**: A new `notebook` cell type in Kale's tag language. Users add `notebook:train` cells alongside `step:` cells in the same notebook. The `.ipynb` file is the composition format — no external config files.
- **Output marshaling**: Non-issue with the sub-pipeline approach — Kale's existing marshal system handles intra-notebook data, KFP artifacts handle inter-notebook data.

### What's in the proposal

- Motivation, goals, user stories
- Background on Kale's current architecture
- Detailed evaluation of all design options with rationale for chosen/rejected
- Notes on constraints (cycle detection, merge chain, name collisions)
- Implementation plan, test plan, migration, consequences
- Open questions for future work (nesting depth, React Flow integration)

Ref: https://github.com/kubeflow/kale/issues/812

### DIFF
--- a/docs/proposals/0812-composable-notebooks/README.md
+++ b/docs/proposals/0812-composable-notebooks/README.md
@@ -248,6 +248,8 @@ Key finding: `@dsl.pipeline` creates a `GraphComponent` (`kfp/dsl/graph_componen
 
 **Chosen because**: Preserves cell-level visibility in the KFP UI. Reuses 100% of existing Kale infrastructure (NotebookProcessor, Compiler, marshal system). Makes Decision 4 (output marshaling) a non-issue. The composition layer is purely additive. This approach was validated by Yash's GSoC POC, which successfully compiled multi-notebook compositions into KFP-valid pipeline YAML using this pattern.
 
+**DSL file structure**: Rather than generating a single monolithic DSL file, each notebook should produce its own independent `.py` DSL file. The orchestrating DSL file imports the sub-notebook DSL files as dependencies. This mirrors the recursive nature of the processing — each notebook is an independent module that can be imported without circular references — and gives each notebook natural ownership of its own configuration (base image, cache, accelerators).
+
 #### Option C: Kale-Owned Notebook-as-Component — Rejected
 
 Generate a regular `@dsl.component` that embeds and executes the notebook using Kale's own marshal infrastructure for I/O (no Papermill).
--- a/docs/proposals/0812-composable-notebooks/README.md
+++ b/docs/proposals/0812-composable-notebooks/README.md
@@ -376,7 +376,7 @@ If Option A (`@dsl.notebook_component`) had been chosen, this would have been th
 
 4. **Type inference uses name heuristics.** KFP artifact types are inferred from variable names using Kale's existing type map (`model`→Model, `dataset`→Dataset, `metrics`→Metrics, etc.). Future work may add explicit type annotations.
 
-5. **`notebook` cells break the code merge chain.** In Kale, untagged cells merge into the previous step. A `notebook:` cell is a reference to another notebook, not executable code — subsequent untagged cells must NOT be merged into it. The `notebook` type must be treated as a boundary (like `imports`, `skip`, `pipeline-parameters`) that stops the merge. Untagged cells after a `notebook:` cell should belong to the next explicit `step:` cell, not to the notebook reference.
+5. **`notebook` cells break the code merge chain.** In Kale, untagged cells merge into the previous step. A `notebook:` cell is a reference to another notebook, not executable code — subsequent untagged cells must NOT be merged into it. This requires a new merge rule: the parser must reset the merge target so that following untagged cells belong to the next explicit `step:` cell, not to the notebook reference or to any previous step.
 
 ### Risks and Mitigations
 
--- a/docs/proposals/0812-composable-notebooks/README.md
+++ b/docs/proposals/0812-composable-notebooks/README.md
@@ -1,0 +1,508 @@
+# KEP-0812: Composable Kale Notebooks
+
+|                     |                                                                  |
+| ------------------- | ---------------------------------------------------------------- |
+| **Authors**         | [ederign](https://github.com/ederign), collaborating with [StefanoFioravanzo](https://github.com/StefanoFioravanzo) and [Ya-shh](https://github.com/Ya-shh) |
+| **Created**         | 2026-05-27                                                       |
+| **Updated**         | 2026-06-25                                                       |
+| **Status**          | Provisional                                                      |
+| **Google Doc (draft)** | https://docs.google.com/document/d/1SBgVL91Z_2mz1R0WH2UXIX0V4WZECK2KvUl4LE016Cw/edit |
+| **Relevant Issues** | https://github.com/kubeflow/kale/issues/812                      |
+| **GSoC 2026**       | [Project #11: Composable Kale Notebooks](https://www.kubeflow.org/events/upcoming-events/gsoc-2026/#project-11-composable-kale-notebooks-with-visual-pipeline-editor) |
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [User Stories](#user-stories)
+- [Background: How Kale Works Today](#background-how-kale-works-today)
+- [Proposal](#proposal)
+  - [Architecture](#architecture)
+  - [The `notebook` Cell Type](#the-notebook-cell-type)
+  - [Boundary Variable Detection](#boundary-variable-detection)
+  - [Compiled Output](#compiled-output)
+- [Design Details](#design-details)
+  - [Decision 1: Compilation Approach](#decision-1-compilation-approach)
+  - [Decision 2: Notebook Interface Declaration](#decision-2-notebook-interface-declaration)
+  - [Decision 3: Composition Definition Format](#decision-3-composition-definition-format)
+  - [Decision 4: Output Marshaling](#decision-4-output-marshaling)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Test Plan](#test-plan)
+- [Implementation Plan](#implementation-plan)
+- [Migration](#migration)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives Considered](#alternatives-considered)
+- [Consequences](#consequences)
+- [Open Questions](#open-questions)
+
+## Summary
+
+Extend Kale to support composition of multiple notebooks into a single Kubeflow Pipeline. A new `notebook` cell type lets users reference other notebooks as sub-pipelines directly inside their notebook, alongside regular `step:` cells. Each referenced notebook is compiled through the existing NotebookProcessor and Compiler, producing a KFP sub-pipeline (GraphComponent) with full cell-level visibility in the KFP UI. Boundary variables between notebooks and steps are detected automatically via the same AST/PyFlakes mechanism Kale uses for cell-to-cell dependencies.
+
+This KEP is being developed as part of the [Google Summer of Code 2026](https://www.kubeflow.org/events/upcoming-events/gsoc-2026/#project-11-composable-kale-notebooks-with-visual-pipeline-editor) program under the Kubeflow organization.
+
+## Motivation
+
+Notebook workflows are commonly split across multiple files: one for data preprocessing, another for training, another for evaluation. Today, Kale only converts a single notebook into a pipeline. There is no way to compose multiple notebooks into a coordinated workflow without manually writing KFP pipeline code.
+
+This forces users to either cram everything into one notebook (hurting modularity and reuse) or leave the notebook environment to wire things together (losing the notebook-native experience that Kale provides).
+
+### Goals
+
+1. Compose multiple notebooks into a single KFP pipeline with validated data flow
+2. Preserve cell-level step visibility for each notebook in the KFP UI
+3. Detect data flow between notebooks automatically — zero friction for the user
+4. Allow mixing of regular `step:` cells and `notebook:` cells in the same notebook
+5. Composed notebooks compile and run through the existing `kale --nb` CLI path — no separate command
+6. Preserve compatibility with the existing single-notebook workflow
+7. Keep the composition layer additive — the existing NotebookProcessor and Compiler are not modified
+
+### Non-Goals
+
+- Visual pipeline editor (future work using React Flow, depends on this KEP)
+- Runtime notebook orchestration outside of KFP
+- Changing the behavior of existing cell types or the single-notebook compilation path
+- Supporting non-notebook components in the composition (e.g., container components, custom scripts)
+
+---
+
+## User Stories
+
+### Story 1: Data Scientist Composes a Training Workflow
+
+A data scientist has three notebooks — `preprocess.ipynb`, `train.ipynb`, and `evaluate.ipynb` — that they run manually in sequence. They create a `main.ipynb` with `notebook:` cells referencing each one:
+
+```
+┌─────────────────────────────────────────────────────────┐
+│ [notebook:preprocess 📓]                                │
+│  (references ./preprocess.ipynb)                        │
+├─────────────────────────────────────────────────────────┤
+│ [notebook:train 📓]                                     │
+│  (references ./train.ipynb)                             │
+├─────────────────────────────────────────────────────────┤
+│ [notebook:evaluate 📓]                                  │
+│  (references ./evaluate.ipynb)                          │
+└─────────────────────────────────────────────────────────┘
+```
+
+Kale detects that `preprocess` produces `dataset`, `train` consumes it, `train` produces `model`, `evaluate` consumes it. One click on "Compile and Run" produces a single KFP pipeline with three sub-pipelines, each showing its internal cell-level steps.
+
+### Story 2: Mixing Steps and Sub-Notebooks
+
+An ML engineer has a reusable `train.ipynb` from a colleague. They want to add their own preprocessing and postprocessing code around it:
+
+```
+┌─────────────────────────────────────────────────────────┐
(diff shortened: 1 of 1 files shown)


docs/proposals/0812-composable-notebooks/README.md
```diff
@@ -0,0 +1,508 @@
...
+
+3. **Variable name collisions are an error.** When two notebooks/steps define the same variable name, the compiler must raise an error rather than silently picking one. Ambiguous data flow must be resolved by the user (rename the variable in one of the notebooks). Note: the current POC has a bug here — it uses `setdefault()` which silently keeps the first notebook and drops the second. This must be fixed to raise an explicit error.
+
+4. **Type inference uses name heuristics.** KFP artifact types are inferred from variable names using Kale's existing type map (`model`→Model, `dataset`→Dataset, `metrics`→Metrics, etc.). Future work may add explicit type annotations.
+
+5. **`notebook` cells break the code merge chain.** In Kale, untagged cells merge into the previous step. A `notebook:` cell is a reference to another notebook, not executable code — subsequent untagged cells must NOT be merged into it. The `notebook` type must be treated as a boundary (like `imports`, `skip`, `pipeline-parameters`) that stops the merge. Untagged cells after a `notebook:` cell should belong to the next explicit `step:` cell, not to the notebook reference.
```
StefanoFioravanzo: Are we sure that `imports` and `pipeline-parameters` act as a boundary? `skip` definitely does, but I don't remember if the others do too. Asking just to make sure

docs/proposals/0812-composable-notebooks/README.md
```diff
@@ -0,0 +1,508 @@
...
+│ [notebook:train 📓]                                     │
+│  (references ./train.ipynb)                             │
+├─────────────────────────────────────────────────────────┤
+│ [step:evaluate ●●]                                      │
+│  score = evaluate_model(model, test_data)               │
+│  print(f"Accuracy: {score}")                            │
+└─────────────────────────────────────────────────────────┘
+```
+
+Kale detects that `train.ipynb` uses `dataset` and `features` (defined by `step:preprocess`) and that it produces `model` and `test_data` (used by `step:evaluate`). The compiled pipeline has `preprocess_step` (a component), `train_pipeline` (a sub-pipeline with its own internal steps), and `evaluate_step` (a component). Data flows automatically between all three — variables cross the step/notebook boundary the same way they cross step/step boundaries.
```
StefanoFioravanzo, line 115: The `train` notebook cannot be standalone, since it "needs" to depend on `dataset` and `features` variables for it to be acceptable for this pattern. I wonder: how will someone work on and test this notebook? 
This is partially mentioned below, but I feel like we haven't really made explicit what is the desired user experience in this case. It might be acceptable for a v1 to indeed require a sub-notebook to not define some of its variables. For a v2, I feel like we need to overcome this limitation

Thinking out loud: If I am developing `train` and I know someone will attach to it, I will probably define stub variables for me to test the notebook, with dummy values. These stub variables can indeed become part of the signature of the notebook. Below, in `Decision 2`, we mention the possibility of using `notebook-outputs` as mechanism to define outputs. Doing something similar for inputs might be interesting, or extending the `pipeline-parameters` mechanism.

Anyway - I think this part is still up for debate. I think we should add some specific considerations about this in the KEP

StefanoFioravanzo (commented):
I love this KEP, overall it is aligned with what we discussed already and provides a comprehensive overview of what we want to build.
I feel like we need to better explain just a couple of aspects, which I commented above

docs/proposals/0812-composable-notebooks/README.md
Ya-shh: @StefanoFioravanzo , I have checked the parser, and the three don't behave the same. After a `skip `cell, a following untagged cell merges into the step before it (skip is transparent, it doesn't change the active step). After` imports `or `pipeline-parameters`, a following untagged cell goes into that block, not a step. So none of them send a following cell to the next `step:`, which is what we want for `notebook`:. So `notebook:` is its own case, not the same as any of those three.  We would need to fix the wording to describe its own rule , @ederign please have a look

docs/proposals/0812-composable-notebooks/README.md
Ya-shh: @StefanoFioravanzo yeah, this is real gap,  right now a sub-notebook has to leave some of its variables undefined for us to pick them up as inputs, which is exactly why it can't run standalone. And if we add stubs to test it, those variables become defined, so we stop picking them up, the real upstream values never get passed in, and it silently runs on the dummy values. So we can't get both standalone-testable and composable unless the stubs are a declared signature, like you said

` pipeline-parameters` fits scalar inputs, but `dataset/features` are artifacts, so those would need their own declaration, a `notebook-inputs `alongside the `notebook-outputs` idea. Probably both, by type.

So v1 keeps the current behavior  which is sub-notebook leaves some variables undefined , and v2 adds explicit input/output declarations for the reusable case.  We could add a short section on this @ederign , please have a look

docs/proposals/0812-composable-notebooks/README.md
```diff
@@ -0,0 +1,508 @@
...
+- The cell body is empty for MVP — the metadata carries all the information
+
+### Boundary Variable Detection
+
+Data flow between notebooks and steps is inferred automatically using the same mechanism Kale uses for cell-to-cell dependencies, lifted one level:
+
+1. For each notebook, scan all step code for **defined variables** (AST: top-level assignments) and **needed variables** (PyFlakes: undefined names)
+2. If notebook/step B needs variable `X` and notebook/step A defines it, create a data edge A→B
+3. Topologically sort the DAG. Raise an error on cycles.
+4. For each boundary variable, identify which specific step within each notebook defines/consumes it. Wire KFP artifacts between those steps.
```
jesuino, line 201: I do believe that the data transfer between notebooks may be related to usually bigger artifacts (model pre trained moving to a post-training notebook, for exampe. I know that there's a feature to [mount volumes for steps](https://github.com/kubeflow/kale/issues/843), would we need something like this to make sure that the data flowing between notebooks are done correctly or is the built-in data transfer mechanism enough for this?

docs/proposals/0812-composable-notebooks/README.md
```diff
@@ -0,0 +1,508 @@
...
+  --> Composition Layer (match boundary variables across steps and sub-pipelines)
+  --> Compiler (Jinja2 templates → KFP DSL with sub-pipelines)
+  --> kfp.compiler.Compiler() → pipeline.yaml
+```
+
+### The `notebook` Cell Type
+
+A new `notebook` cell type is added to Kale's tag language, following the same conventions as the existing `step` type:
+
+- Tag format: `notebook:<name>` (analogous to `step:<name>`)
```
jesuino, line 189: Will the tag validate for existing notebooks when the user enters the notebook and/or will we have a way for users to either drag their notebook into the parent notebook cells and automatically make it a notebook tag into that cell?

docs/proposals/0812-composable-notebooks/README.md
```diff
@@ -0,0 +1,508 @@
...
+- [Migration](#migration)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives Considered](#alternatives-considered)
+- [Consequences](#consequences)
+- [Open Questions](#open-questions)
+
+## Summary
+
+Extend Kale to support composition of multiple notebooks into a single Kubeflow Pipeline. A new `notebook` cell type lets users reference other notebooks as sub-pipelines directly inside their notebook, alongside regular `step:` cells. Each referenced notebook is compiled through the existing NotebookProcessor and Compiler, producing a KFP sub-pipeline (GraphComponent) with full cell-level visibility in the KFP UI. Boundary variables between notebooks and steps are detected automatically via the same AST/PyFlakes mechanism Kale uses for cell-to-cell dependencies.
```
jesuino, line 44: One question I have is what is in practice a sub-pipeline? I don't see this exact component on Kubeflow Pipelines. Notice that if we say "KFP sub-pipeline", users will look exactly for this term on KFP Pipelines docs.

Another question I have is if we call a notebook that has no configuration (base image, cache, accelerators), will the sub-notebook inherent the settings from the parent?  Is it even possible to setup the same parameters as a step using this sub-component? If we refer something from KFP, users will know what is possible and even be able to modify the generated DSL.

docs/proposals/0812-composable-notebooks/README.md
```diff
@@ -0,0 +1,508 @@
...
+```
+
+### The `notebook` Cell Type
+
+A new `notebook` cell type is added to Kale's tag language, following the same conventions as the existing `step` type:
+
+- Tag format: `notebook:<name>` (analogous to `step:<name>`)
+- The referenced notebook path is stored in cell metadata
+- Ordering uses the same `prev:` mechanism as steps, though automatic inference handles most cases
+- The cell body is empty for MVP — the metadata carries all the information
```
jesuino, line 192: How will we determine the called notebook base parameters such as base image, accelerator and cache? Will it support the same settings as a `step`?

jesuino (changes requested):
Thanks for this feature.

I am adding a few questions I have. Let me know if it makes sense for you.

Thanks!

docs/proposals/0812-composable-notebooks/README.md
ederign: Good catch from both of you. Ya-shh is right, I just updated with:

> 5. **`notebook` cells break the code merge chain.** In Kale, untagged cells merge into the previous step. A `notebook:` cell is a reference to another notebook, not executable code — subsequent untagged cells must NOT be merged into it. This requires a new merge rule: the parser must reset the merge target so that following untagged cells belong to the next explicit `step:` cell, not to the notebook reference or to any previous step.

docs/proposals/0812-composable-notebooks/README.md
ederign: Thanks folks, for v1 my suggstion is we accept this limitation. Added caveat #6 to the KEP documenting this explicitly.
> 6. **Sub-notebooks cannot run standalone in v1.** Automatic inference detects a sub-notebook's inputs as undefined variables (via PyFlakes). If the sub-notebook defines those variables — even as stubs for testing — they are no longer undefined, inference stops detecting them as inputs, and upstream values silently never arrive. This means sub-notebooks used in a composition cannot also run independently. This is an accepted limitation of the zero-friction automatic inference approach (Decision 2, Option D). Post-MVP, explicit input/output declarations (see Decision 2, Option A) could resolve this by letting sub-notebooks define stub values that the compiler replaces with upstream artifacts when composed.

For a possible follow-up, sketched a `notebook-inputs` tag (paired with `notebook-outputs`) in Decision 2 Option A — sub-notebooks define stubs for local testing, the compiler replaces them with upstream artifacts when composed. See 46f4758.
>**Deferred to post-MVP**: Explicit declarations add friction without benefit when the composition is authored by the same person who wrote the notebooks. Needed later for the reusable-component story where different people author and compose notebooks. One possible follow-up is a `notebook-inputs` tag (paired with `notebook-outputs`) that would let sub-notebooks define stub values for local testing that the compiler replaces with upstream artifacts when composed, resolving the standalone vs. composable tension described in the caveats section.

I hope @Ya-shh has time to fix those :-)

docs/proposals/0812-composable-notebooks/README.md
ederign: Thanks for the review William!

With the sub-pipeline approach we chose, data flows at two levels , within each notebook (Kale's existing one) and between notebooks (KFP artifacts declared as `Input[Type]`/`Output[Type]` on sub-pipeline signatures) works as expected for this use case. @Ya-shh can you add this to your todo list to test? (I assume it works)

In theory this should handle large artifacts like pre-trained models since KFP artifact storage is designed for that. Volume mounts (#843) would be complementary for cases where you want to skip serialization overhead, but shouldn't be required for correctness. @jesuino we need to design here, how those mounts will flow in those use cases as well.

That said, we haven't validated this with large artifacts yet,  it's something we need to test during implementation to make sure it holds up in practice.

@jesuino makes sense?

docs/proposals/0812-composable-notebooks/README.md
ederign: For now it's manual , the user picks the notebook path in the cell metadata editor. At compile time, we'll validate that the file exists, is a valid Kale notebook, has no circular refs, etc.

Drag-and-drop from the file browser to auto-create a `notebook:` tag would be really cool though (but I'm not sure how this works in Jupyter). let's look into that when we get to the frontend phase (Phase 5). @Ya-shh can you add this for our wish list?

docs/proposals/0812-composable-notebooks/README.md
ederign: Fair point on the terminology. "Sub-pipeline" is my wrong shorthand, what we're actually doing is using a `@dsl.pipeline`-decorated function called as a task inside another pipeline, which internally creates a `GraphComponent` (extends `BaseComponent`). We should  probalby clarify this in the KEP so people can map it to KFP concepts.

On configuration inheritance, I've added to open questions. Each sub-notebook gets processed independently with its own PipelineConfig, so right now it would just use defaults if nothing is configured. Whether it should inherit the parent's base image, cache, accelerators — we need to define that. I'll add it to the open questions.

additions:
> 6. **KFP terminology**: The KEP uses "sub-pipeline" as shorthand. The underlying KFP mechanism is a `@dsl.pipeline`-decorated function called as a task inside another pipeline, which creates a `GraphComponent`. We should clarify this mapping so users can find the relevant KFP documentation.

7. **Configuration inheritance**: When a sub-notebook has no configuration (base image, cache, accelerators), should it inherit settings from the parent notebook's PipelineConfig, or use defaults? Should sub-notebooks support the same configuration options as steps?

docs/proposals/0812-composable-notebooks/README.md
ederign: Good question — this is the same concern I replied to above. Added it as an open question in the KEP, we need to figure this out during implementation.

ederign:
@jesuino @Ya-shh @StefanoFioravanzo thanks for the review! If you are happy with the answers, please resolve the conversation, otherwise keep the questions coming!

docs/proposals/0812-composable-notebooks/README.md
```diff
@@ -248,6 +248,8 @@ Key finding: `@dsl.pipeline` creates a `GraphComponent` (`kfp/dsl/graph_componen
 
 **Chosen because**: Preserves cell-level visibility in the KFP UI. Reuses 100% of existing Kale infrastructure (NotebookProcessor, Compiler, marshal system). Makes Decision 4 (output marshaling) a non-issue. The composition layer is purely additive. This approach was validated by Yash's GSoC POC, which successfully compiled multi-notebook compositions into KFP-valid pipeline YAML using this pattern.
 
+**DSL file structure**: Rather than generating a single monolithic DSL file, each notebook should produce its own independent `.py` DSL file. The orchestrating DSL file imports the sub-notebook DSL files as dependencies. This mirrors the recursive nature of the processing — each notebook is an independent module that can be imported without circular references — and gives each notebook natural ownership of its own configuration (base image, cache, accelerators).
```
ederign, line 251: @Ya-shh after @jesuino comments, I think this will be a better design for our compilation artifact because will allow us further configs. @StefanoFioravanzo also raised this point a few days ago!. wdyt?

ada333:
@ederign @Ya-shh to add my two cents here - how will the imports be handled? If f.e. train.ipynb imports numpy and outputs numpy array object, in evaluate.ipynb (that uses train.ipynb as nested notebook) do we need to import numpy there as well or will it automatically share imports with all nested notebooks?

Ya-shh:
> @ederign @Ya-shh to add my two cents here - how will the imports be handled? If f.e. train.ipynb imports numpy and outputs numpy array object, in evaluate.ipynb (that uses train.ipynb as nested notebook) do we need to import numpy there as well or will it automatically share imports with all nested notebooks?

hey @ada333 
imports are per notebook by design, each notebook's imports/functions cells are prepended to that notebook's own steps only, and nothing leaks across notebook boundaries, so each notebook is an independent unit that owns its imports (same reasoning as the one DSL module per notebook structure),
for your numpy example the array itself crosses the boundary through Kale's marshal layer, whose numpy backend serializes/deserializes it without any user level import. So evaluate.ipynb does not need import numpy just to receive the array, but numpy must be installed in its image, and if its code references np. by name it declares its own import numpy in its imports cell, exactly as it would standalone

docs/proposals/0812-composable-notebooks/README.md
Ya-shh: Yes I strongly agree here @ederign  , and I  have made changes to my PR

jesuino (approved):
Thanks!

ederign:
Merging this because it's already WIP. We can always amend later if necessary!

docs/proposals/0812-composable-notebooks/README.md
StefanoFioravanzo: Strongly agree :) This is a good direction!

docs/proposals/0812-composable-notebooks/README.md
StefanoFioravanzo: besides #843 which you linked to, we should also refer to https://github.com/kubeflow/kale/issues/856 They are both related although in different ways

PR #704, opened 2026-03-17
docs: improve documentation for utils and pipeline modules

Kaibalya-Mohanty:
## Documentation Improvements

- Improved docstrings for several utility functions:
  - random_string
  - abs_working_dir
  - shorten_long_string
  - dedent
  - remove_ansi_color_sequences

- - Added module-level documentation explaining pipeline architecture and execution flow:
  - Enhanced `add_step` method docstring
  - Added clearer explanations for pipeline structure and usage

These changes aim to improve clarity and onboarding for new contributors.

Looking forward to feedback!

PR #864, opened 2026-06-30
fix: improve ANSI regex, add short-string guard, add debug logging in rm_r

Kaibalya-Mohanty:
This PR separates the behavioral changes from PR #704 (docs-only).

Changes:
- remove_ansi_color_sequences: Broadened regex to catch more ANSI 
  escape sequences, added isinstance check to handle non-string input safely.
- shorten_long_string: Added early-return guard for strings already 
  shorter than chars * 2 to avoid unnecessary slicing.
- rm_r: Added log.debug() call to trace path removal attempts for 
  easier debugging.

### DIFF
--- a/kale/common/utils.py
+++ b/kale/common/utils.py
@@ -88,6 +88,7 @@ def rm_r(path, ignore_missing=True, silent=False):
         raise e
 
     log.info("Removing path `%s'", path)
+    log.debug("Attempting to remove path: %s", path)
 
     try:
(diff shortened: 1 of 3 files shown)


ada333:
hi @Kaibalya-Mohanty this PR is the same as https://github.com/kubeflow/kale/pull/704/changes , please close this one.

ederign:
This PR don't bring too much value for Kale so I'm closing it. feel free to check something in our roadmap.

PR #867, opened 2026-07-02
feat: add notebook: cell type for multi-notebook composition.

Ya-shh:
Backend and labextension for the composable-notebooks KEP ([KEP-0812](https://github.com/kubeflow/kale/pull/847)). Compose multiple Kale notebooks into one Kubeflow pipeline through the existing `kale --nb` path and the Kale panel, with no new command or flag.

Demo of the composition flow on YouTube: https://youtu.be/rJEnyU6t3S8?si=LTkmlLsmMPCjdKlh

### What it does

A `notebook:` cell type lets a notebook reference other notebooks. Two shapes are supported:

- **Sequence of notebooks (Story 1):** the root notebook holds only `notebook:` references. Each referenced notebook becomes a KFP sub-pipeline (GraphComponent), so cell-level step visibility is preserved. Data flow between notebooks is inferred by matching variable names, the same AST and PyFlakes mechanism Kale already uses between cells, applied one level up.

- **Mixed steps and notebooks (Story 2):** the root notebook can also carry its own `step:`, `imports` and `functions` cells alongside the references. Each such step becomes a top-level component, and ordering and data cross the step/notebook boundary the same way they cross a step/step one. A root step runs after every unit above it and before every unit below it (reading order), while independent references between steps still run in parallel.

Whether a notebook compiles as a plain pipeline, a sequence, or a mixed composition is decided entirely by the cells it contains.

### How it works (reworked after review)

The first version of this PR put composition in a separate `processors/workflow.py` behind a branch in `cli.py`. Two review findings retired that design:

- @StefanoFioravanzo pointed out the branch read as a codepath bolted onto the primary flow, and that the recursion belongs inside `NotebookProcessor`, the way a compiler recurses.
- @ederign found that the Kale panel's compile RPC never reached `workflow.py` at all, so composing from the UI silently produced a pipeline with the referenced notebooks missing.

Composition now lives inside `NotebookProcessor`, and `workflow.py` and the `cli.py` branch are deleted. Every notebook goes through one compile path, which is what fixes the UI case:

1. **Parse.** The processor reads cells in order. On a `notebook:` cell it recurses into a child `NotebookProcessor` and attaches the resulting pipeline as a `SubPipeline` node, a peer of `Step` in the same graph.
2. **Link.** `dependencies_detection` resolves only along edges that already exist, and a composition has no `prev:` tags to create them. So the edges are inferred first: data flow between units, plus reading order for the root notebook's own steps.
3. **Resolve.** The existing `dependencies_detection` then runs unchanged, so a variable crossing a notebook boundary is resolved by exactly the code that resolves one crossing a cell boundary.
4. **Emit.** `Compiler` walks the mixed graph and writes one DSL module per referenced notebook plus the orchestrator.

### DSL file structure (per review feedback)

Each notebook produces its own independent DSL module (`.kale/kale_notebook_<name>.py`) and the orchestrating DSL file imports those modules as dependencies. This mirrors the recursive shape of the processing: each notebook is an independent module with no circular references, each owns its own configuration, and each is compilable standalone. In a mixed composition the root notebook's own steps live in the orchestrator, which is that notebook's own module. The `kale_notebook_` prefix means a notebook named `json.ipynb`, `class.ipynb` or `2train.ipynb` compiles rather than erroring.

### What is in the PR

**Backend**

`processors/nbprocessor.py`: the `notebook:<name>` tag (referenced path in cell metadata) added to the tag language; recursive parsing of references into `SubPipeline` nodes; `link_composition_units()` for edge inference; `propagate_subpipeline_boundaries()` to hand each boundary variable to the inner step that produces or consumes it. A `notebook:` cell breaks the code merge chain (KEP-0812 caveat 5): untagged cells after a reference attach to the next `step:` cell, and orphaned code raises rather than being dropped.

`step.py`: `SubPipeline(Step)`, a referenced notebook as a node of the same graph, carrying the same `name`, `ins`, `outs` and `config` so dependency detection and code generation treat a reference and a step uniformly.

`compiler.py`: `generate_composition()` and the per-notebook module emission. Each referenced notebook's module is generated by a compiler bound to that notebook's own pipeline, so its pipeline parameters and imports are the ones its steps are built with, exactly as when it is compiled on its own.

`templates/subpipeline_template.jinja2` and `templates/composition_template.jinja2`: the per-notebook module (components, typed sub-pipeline, standalone compile hook) and the orchestrator that imports the modules, holds any root-step components, and wires everything.

`common/kfputils.py`: `compile_pipeline` carries the DSL script's sibling modules into its temp dir so the orchestrator's imports resolve.

**Frontend (labextension)**

A `Notebook` cell type in the Kale cell editor: pick it, enter the notebook path, and it writes the `notebook:` tag plus the `notebook_path` metadata. Reference cells render an inline `notebook: <name>` chip and are excluded from `prev:` dependency choices and the merge-chain notice.

**Examples and tests**

`examples/composition/` holds `main.ipynb` (three references, dataset to model to prediction) and `main_mixed.ipynb` (references plus a root `step:` cell). `test_composition.py` covers reference-to-node conversion, inferred order and data, variables crossing the step/notebook boundary both ways, reading order, preserved parallelism, the merge rule, orphan code, missing paths, reference cycles, name collisions, per-module output, safe module naming, parameter survival, nested-reference rejection, reported step config, and two real `kfp.compiler` compilations.

The single-notebook path is untouched: a notebook with no `notebook:` cells compiles exactly as before, and no existing cell type or API changes.

### Testing

- `kale --nb examples/composition/main.ipynb` compiles to a root DAG of three sub-DAGs, each holding its notebook's own cell-level steps; `main_mixed.ipynb` adds the root's step as a top-level component wired in reading order.
- The Kale panel's compile RPC produces the same result, which is the case that previously dropped references.
- Both submit and run green against a KFP cluster.
- 297 backend tests passing, ruff clean, labextension tsc, eslint and prettier clean.

### Known limitations

- **Nested references.** A `notebook:` cell inside a referenced notebook raises rather than compiling to something that flattens a level of nesting. Nested composition is not part of the KEP stories, so this is a deliberate rejection rather than a partial implementation. Happy to change it if reviewers would prefer it supported.
- **Per-step configuration.** The composition templates do not yet carry over per-step `limits`, `annotations`, `labels`, retries or timeouts. Compiling reports exactly which step and which field is affected rather than dropping it quietly. Restoring this is the remaining step of the agreed rework plan.
- Explicit typed input/output declarations remain the v2 discussion from the KEP review.


Ya-shh:
You can watch the full demo of this feature on YouTube
https://youtu.be/rJEnyU6t3S8?si=LTkmlLsmMPCjdKlh

Ya-shh:
moved the kfp import above the sys.path mutation so no generated module can shadow it, and the composer now rejects notebook names that collide with kfp/kale/stdlib. Kept insert(0) rather than append deliberately: the generated modules must win resolution of their own names, otherwise a same named module elsewhere on sys.path would silently shadow them

Ya-shh:
New updated DSL structure 
https://youtu.be/AFvds4Jwdnw?si=YMoywpODG25WNArl

thank you @jesuino for the suggestion!

kale/processors/nbprocessor.py
```diff
@@ -465,6 +512,13 @@ def parse_cell_metadata(self, metadata):
                 parsed_tags["step_names"] = [t]
                 return parsed_tags
 
+            # A notebook reference is matched on the full tag (either form),
+            # before the split below could mistake `step:notebook:x` for a step.
```
StefanoFioravanzo: Maybe we could simplify and avoid risking this mistake by only support the form `"notebook:<NAME>"` and NOT support `"step:notebook:<NAME>"`? In this way you would have the `tag_part = "notebook"` and easily identify the notebook name, without collisions. 
Also, introducing two different ways to define this tag is error prone. unless I am missing the reason why you want to also support the `step:notebook:<NAME>` form

StefanoFioravanzo: if indeed we support just the simple `notebook:<NAME>" form you can move this parsing logic below, aligning with the rest of the tag parsing loop.

kale/processors/nbprocessor.py
StefanoFioravanzo: You introduced the concept of "merge_forward", I like it because it makes it explicit what the code wants to do (" merge the next cell, if the tag does not change"). I feel like we can simplify this entire loop by removing this elif branches, and relying only on the merge forward flag, wdyt?

If you flag `merge_forward` to `true` on every cell tag that allows merging the next cell (basically any tag, besides `notebook` and `skip`), then you can just verify if `merge_forward` is true, thus merge.

I might be forgetting edge cases here, but I think it's worth considering simplifying all this.

StefanoFioravanzo: Another idea - you could also have a list of "merge_forward_tags". This list would tell which tags are allowed to merge foward. In this way you don't need to keep the status in a flag, you just check whether the current cell does not have any tag defined, and then check the previous cell tag agains the `merge_forward_tags` list

kale/processors/workflow.py
```diff
@@ -0,0 +1,537 @@
...
+Each notebook is compiled into its own KFP sub-pipeline by reusing Kale's
+per-notebook processing and component generation. The sub-pipelines are wired
+together by matching the variables one notebook defines to those another
+consumes, the same name-based detection Kale uses between cells, one level up.
+The single-notebook path is untouched.
+
+Each notebook produces its own independent DSL module (``.kale/<name>.py``,
+importable and independently compilable), and the orchestrating DSL file
+imports those modules as dependencies. This mirrors the recursive shape of the
+processing and gives every notebook ownership of its own configuration.
```
StefanoFioravanzo: It would be nice to add some ASCII diagram and an simple example to illustrate the concept here

kale/processors/workflow.py
```diff
@@ -0,0 +1,537 @@
...
+    variable. This matches how Kale detects cell-to-cell dependencies.
+    """
+    names = set()
+    for node in ast.parse(code).body:
+        if isinstance(node, ast.Assign):
+            for target in node.targets:
+                names |= _names_in_target(target)
+        elif isinstance(node, (ast.AugAssign, ast.AnnAssign)) and isinstance(node.target, ast.Name):
+            names.add(node.target.id)
+    return names
```
StefanoFioravanzo: I suggest adding all the ast-based utils to common/astutils.py. This is a general comment, so if there are other functions inside this python module that can be moved to reusable common modules, let's do that

kale/cli.py
```diff
@@ -123,6 +178,11 @@ def main():
         if args.devpi_simple_url:
             os.environ["KALE_DEVPI_SIMPLE_URL"] = args.devpi_simple_url
 
+    # Notebooks with `notebook:` cells compose sub-pipelines; otherwise fall
+    # through to the single-notebook path below.
+    if _compile_composition(args):
+        return
+
```
StefanoFioravanzo: I feel like this is a weird pattern. Creating a function that does some heavy lifting processing as a special case of the primary flow, and in that case just "returns", feels like a codepath that was added as an afterthought. We have the opportunity to refactor code so that it all feels natural.

I would expect to see the recursive notebook compilation embedded inside the `NotebookProcessor` call down below. NotebookProcessor should be able to identify if it is compiling a single notebook, or if it needs to start a recursive process.

StefanoFioravanzo: I think this means that a lot of what you implemented in workflow.py should probably be moved inside `NotebookProcessor`, but I haven't looked deeply at `workflow.py` yet

kale/cli.py
```diff
@@ -44,6 +44,61 @@
...
+    # A ValueError here is a problem in how the user tagged or wired the
+    # notebooks (missing notebook path, code in a reference cell, orphaned code,
+    # name collision, cycle), so show the message alone instead of a traceback.
+    try:
+        notebook_refs = extract_notebook_references(args.nb)
+        if not notebook_refs:
+            return False
+
+        nb_meta = nbformat.read(args.nb, 4).metadata.get("kubeflow_notebook", {})
+        pipeline_name = nb_meta.get("pipeline_name") or args.pipeline_name
```
StefanoFioravanzo: This is part of what I mentioned below - a compiler work recursively, so it would be interesting to figure out if it's possible to embed a completely recursive logic inside NotebookProcessor. The processor would start parsing the notebook sequentially, when it finds a notebook reference, it start a recursive processing loop, using the context that is has gathered so far.

StefanoFioravanzo: This is how code compilers work in general, although I am no compiler expert :) so others feel free to chime in

kale/processors/workflow.py
```diff
@@ -0,0 +1,537 @@
...
+        if not name.isidentifier() or keyword.iskeyword(name):
+            raise ValueError(
+                f"Notebook name '{name}' is not a valid Python module name. Rename "
+                f"the notebook: each notebook becomes its own importable DSL module."
+            )
+        if name in sys.stdlib_module_names or name in ("kfp", "kale"):
+            raise ValueError(
+                f"Notebook name '{name}' collides with a reserved module name "
+                f"(Python standard-library, 'kfp' or 'kale'). Rename the notebook: "
+                f"each notebook becomes its own importable DSL module."
```
StefanoFioravanzo: We could automatically prepend the notebooks' modules with a standard prefix, like `kale_notebook_*`, so that we don't throw collision errors that would be frustrating.  This could also be something configurable by the user, form the UI, eventually

StefanoFioravanzo (commented):
@Ya-shh I went through the code. For of all, congrats for submitting this first version. It's a lot of work and I really appreciate the lengths you went to understand the code deeply and implement a sophisticated mechanism.

I didn't run it myself, but I saw the demo and I know the approach work, practically. So we are in a very good position, because you proved this can be done. 

I added some comments inline, but to me the bottom line is that we should try to rethink a bit how you put together the whole logic inside `workflow.py`. When thinking about an elegant, fluid, and well architected solution, I imagine a fully recursive implementation inside `NotebookProcessor`, instead of having to build a separate dedicated logic inside a separate module (`workflow.py`). The function `compose_notebooks_as_subpipelines` is very long and a bit hard to follow as well. 

I'd like to hear what @ederign thinkgs as well, because what I am proposing would mean changing this code A LOT, but at the same time would allows us to have code that anyone can understand and feel like it was built for multi-notebook orchestration from day 1. As it is now, it still feels like it was built afterwards, with bits and pieces conneting sometimes in an awkward way

Also - mind you - I know it's hard to build a fully recursive loop, but that's also why it will be so much more elegant and beautiful :)

ederign (changes requested):
Reviewed at `0f7e2d5`.  I'll share in slack the reproducers that I've used.
 
@StefanoFioravanzo I agree, let's go with the recursive approach.

@Ya-shh, great work first of all, you proved the hard part works. GraphComponents nest, we keep cell-level visibility, the inference works, and it compiles end to end (R2). That was the risky part of the KEP and it's settled.

What decided it for me is that Stefano's point already cost us a real bug.
 I think composition in its current shape can't be compiled from the Kale UI. See screenshot:
Image



The labextension only ever calls one compile RPC — `nb.compile_notebook` (`Commands.ts:235` → `rpc/nb.py:75`) — which goes straight to `NotebookProcessor` + `Compiler` and never reaches `workflow.py`. This PR doesn't touch `kale/rpc/` at all, and all seven labextension files it changes are cell-metadata authoring. So Story 2 compiles a one-step pipeline and throws away both referenced notebooks; Story 1 compiles a pipeline with zero components. Neither warns (R1).

@Ya-shh the demo obviously works, so I'm assuming you compiled with `kale --nb`, which is what the Testing section of the description says, and the UI part was authoring the `notebook:` cells. If you did compile from the Kale panel and it worked, please let me know, then I've got this wrong and I'd want to know.

Either way that's exactly your `cli.py:185` comment, Stefano, the branch is in `cli.main()`, so only the CLI gets it. Move it into `NotebookProcessor` and the RPC works for free, and the React Flow editor later too.

Two things I'd add to your proposal, Stefano:

1. **Pin down what "recursive" means.** `dependencies_detection()` is graph-wide, so the DAG has to exist before resolution. It's two phases: parse recursively (a `notebook:` cell spawns a child processor and attaches a node), then resolve over the graph as we do today.
2. **The missing piece is a `SubPipeline` node next to `Step`**, both with `.name/.ins/.outs/.config`. That's what makes the boundary-variable and dropped-config problems impossible by construction, one definition scanner and one template. Without it, recursion just moves the branching into `nbprocessor.py`, which is already 983 lines.

Suggested order: @Ya-shh please don't start over:

1. Architecture-independent fixes: reuse `get_marshal_candidates`, make the three fallbacks raise, add a real `kfp.compiler` test per story. These can land on their own.
2. `SubPipeline` node.
3. Recursion into `NotebookProcessor`, delete the `cli.py` branch — R1 should then pass with no extra code.
4. Unified emission, which restores the dropped step config.

Your demo, tests and examples stay and act as the safety net for each step.

I'll also share in slack the reproducer for you. Keep hacking!

Ya-shh:
> @Ya-shh I went through the code. For of all, congrats for submitting this first version. It's a lot of work and I really appreciate the lengths you went to understand the code deeply and implement a sophisticated mechanism.
> 
> I didn't run it myself, but I saw the demo and I know the approach work, practically. So we are in a very good position, because you proved this can be done.
> 
> I added some comments inline, but to me the bottom line is that we should try to rethink a bit how you put together the whole logic inside `workflow.py`. When thinking about an elegant, fluid, and well architected solution, I imagine a fully recursive implementation inside `NotebookProcessor`, instead of having to build a separate dedicated logic inside a separate module (`workflow.py`). The function `compose_notebooks_as_subpipelines` is very long and a bit hard to follow as well.
> 
> I'd like to hear what @ederign thinkgs as well, because what I am proposing would mean changing this code A LOT, but at the same time would allows us to have code that anyone can understand and feel like it was built for multi-notebook orchestration from day 1. As it is now, it still feels like it was built afterwards, with bits and pieces conneting sometimes in an awkward way
> 
> Also - mind you - I know it's hard to build a fully recursive loop, but that's also why it will be so much more elegant and beautiful :)

@StefanoFioravanzo  thank you for  appreciating my work  and thanks a lot for  the deep insights and honestly its correct to say that `workflow.py` grew around `NotebookProcessor` instead of inside it, and the `cli.py` return-True branch is exactly the afterthought pattern you called out.  @ederign  UI finding proved the cost of that better than any style argument could, any entry point that doesn't know about `workflow.py` silently gets the wrong pipeline

So I'm on board with the recursive approach. On your inline points, agreed on dropping the `step:notebook:` form ,it existed because the old cell editor could only write `step: tags`, but this PR's Notebook cell type writes `notebook:` directly, so it lost its reason to exist, agreed on the merge_forward_tags list (nicer than my flag juggling), agreed on the kale_notebook_* prefix instead of collision errors, and I'll move what survives of the AST helpers into astutils and add the ASCII diagram. The long compose_notebooks_as_subpipelines mostly dissolves once the logic lives in the processor, which I think is the real fix for its length

I'll reply on each inline thread as the corresponding change lands so they're easy to resolve :)

Ya-shh:
> Reviewed at `0f7e2d5`. I'll share in slack the reproducers that I've used.
> 
> @StefanoFioravanzo I agree, let's go with the recursive approach.
> 
> @Ya-shh, great work first of all, you proved the hard part works. GraphComponents nest, we keep cell-level visibility, the inference works, and it compiles end to end (R2). That was the risky part of the KEP and it's settled.
> 
> What decided it for me is that Stefano's point already cost us a real bug. I think composition in its current shape can't be compiled from the Kale UI. See screenshot: Image
> 
> The labextension only ever calls one compile RPC — `nb.compile_notebook` (`Commands.ts:235` → `rpc/nb.py:75`) — which goes straight to `NotebookProcessor` + `Compiler` and never reaches `workflow.py`. This PR doesn't touch `kale/rpc/` at all, and all seven labextension files it changes are cell-metadata authoring. So Story 2 compiles a one-step pipeline and throws away both referenced notebooks; Story 1 compiles a pipeline with zero components. Neither warns (R1).
> 
> @Ya-shh the demo obviously works, so I'm assuming you compiled with `kale --nb`, which is what the Testing section of the description says, and the UI part was authoring the `notebook:` cells. If you did compile from the Kale panel and it worked, please let me know, then I've got this wrong and I'd want to know.
> 
> Either way that's exactly your `cli.py:185` comment, Stefano, the branch is in `cli.main()`, so only the CLI gets it. Move it into `NotebookProcessor` and the RPC works for free, and the React Flow editor later too.
> 
> Two things I'd add to your proposal, Stefano:
> 
> 1. **Pin down what "recursive" means.** `dependencies_detection()` is graph-wide, so the DAG has to exist before resolution. It's two phases: parse recursively (a `notebook:` cell spawns a child processor and attaches a node), then resolve over the graph as we do today.
> 2. **The missing piece is a `SubPipeline` node next to `Step`**, both with `.name/.ins/.outs/.config`. That's what makes the boundary-variable and dropped-config problems impossible by construction, one definition scanner and one template. Without it, recursion just moves the branching into `nbprocessor.py`, which is already 983 lines.
> 
> Suggested order: @Ya-shh please don't start over:
> 
> 1. Architecture-independent fixes: reuse `get_marshal_candidates`, make the three fallbacks raise, add a real `kfp.compiler` test per story. These can land on their own.
> 2. `SubPipeline` node.
> 3. Recursion into `NotebookProcessor`, delete the `cli.py` branch — R1 should then pass with no extra code.
> 4. Unified emission, which restores the dropped step config.
> 
> Your demo, tests and examples stay and act as the safety net for each step.
> 
> I'll also share in slack the reproducer for you. Keep hacking!

@ederign thank you  for appreciating my work and thanks for  the reproducers, and for pinning down R1 so precisely. To confirm your assumption: yes, the demo compiled via kale --nb, the panel was only used for authoring the notebook: cells. I ran both your scripts verbatim against the PR head and got identical output: R2 composes producer - consumer, and through the RPC path Story 1 gives zero components and Story 2 one, both silent, you had it exactly right :)

I also verified the pre existing guard issue from your README: Pipeline.steps is a property returning a generator, so not self.pipeline.steps at compiler.py:112 can never be True. Since that's independent of this PR I'll send it as a separate small fix

Now the Plan, following your order with R1 first as you suggested:

1. Kill R1's silence now: the steps-guard fix plus a clear error from NotebookProcessor when it meets notebook: cells before it's composition aware, and your R1/R2 adapted into regression tests. R1's test asserts the error for now and flips to asserting full composition once step 3 lands, so the same test becomes the acceptance criterion for the recursion. Alongside: reuse get_marshal_candidates, make the three fallbacks raise, plus @StefanoFioravanzo 's tag-form, prefix, and merge simplifications

2.SubPipeline node next to Step

3.Recursive parse in NotebookProcessor, delete the cli.py branch

4.Unified emission, restoring the per-step config

Ya-shh:
One process question for you both @StefanoFioravanzo  and @ederign , please let me know your thoughts ?
Should I do sequential commits on this PR, or stacked PRs per step? Happy either way, whichever makes the re-review easiest for you :)

ederign:
@Ya-shh do all commits on this PR! I'm fine reviewing it.

kale/processors/nbprocessor.py
Ya-shh: > Maybe we could simplify and avoid risking this mistake by only support the form `"notebook:<NAME>"` and NOT support `"step:notebook:<NAME>"`? In this way you would have the `tag_part = "notebook"` and easily identify the notebook name, without collisions. Also, introducing two different ways to define this tag is error prone. unless I am missing the reason why you want to also support the `step:notebook:<NAME>` form

You are right, and the reason for the second form is gone. It existed because the old cell editor could only write step: tags, but this PR adds a proper Notebook cell type that writes notebook:<name> directly, so the dual form was vestigial. Dropped in the commit `3e3cea0`: NOTEBOOK_TAG is now just notebook:<name>, and step:notebook:x fails with the standard "Unrecognized tag" error. No notebook in the repo (or in Eder's reproducers) had to change

Ya-shh: > if indeed we support just the simple `notebook:" form you can move this parsing logic below, aligning with the rest of the tag parsing loop.

Done in the same commit, with one tag form there's no ambiguity, so the special-case above the split is gone and the reference is parsed inside the normal loop alongside step/prev/image/cache, using tag_parts exactly like the others

kale/processors/nbprocessor.py
Ya-shh: Took yours second idea in the commit `3677ba5`. There's now a `merge_targets` mapping of tag to the block it absorbs into, which removes both branch ladders: the four if step_name == "imports" cases and the four elif prev_step_name ==  cases become one lookup each. Eight branches, two lookups

One naming note: I renamed the flag to awaiting_step, your merge_forward_tags describes a tag absorbing the cells that follow it, while my flag meant "hold these cells for the next step" after a notebook: reference, the mapping now carries the concept you liked :)

Verified behavior-preserving: I generated the DSL for a composition and a plain single-notebook pipeline, stashed the change, regenerated with the old parser, and diffed

kale/processors/workflow.py
Ya-shh: thanks for the suggestion ! I have added in` 3677ba5`: the module docstring now shows a root notebook with step:prepare / notebook:train / step:report and the pipeline it compiles to (component - sub-DAG - component), plus a line on how dataset and model cross the step/notebook boundary as artifacts

kale/processors/workflow.py
Ya-shh: Better than moving them, they are deleted. `5e396ff` replaces our hand-rolled scanner with Kale's own astutils.get_marshal_candidates, the same one dependencies_detection uses between cells, so composition and cell-level detection now share one definition of "what can be marshalled out"

One adaptation was needed: get_marshal_candidates returns every top-level name including ones only read, so used directly every consumer would look like a producer of the variable it consumes, 

subtracting the unit's own PyFlakes-missing names is what orients the edge, there's a test for exactly that
Reusing Kale's scanner also removed the mismatch that had forced a fallback in _toplevel_ref, so that's gone too, along with the three silent fallbacks Eder flagged, which now raise

kale/cli.py
Ya-shh: Agreed and @ederign  R1 finding showed the cost of the pattern , the end state is that workflow.py goes away entirely, 

the unit abstraction becomes a SubPipeline node beside Step, and the emission moves into Compiler and the templates. Phase 1 landed the pieces that transfer as-is Kale's own marshal-candidate scanner, 
no silent fallbacks, single tag form, the cli.py branch and this module go with the recursion in step 3

---

**Update:** this is done in `5bc0b6e`. workflow.py and the `_compile_composition` branch in cli.py are both deleted, so there is no special-case codepath left 
A `notebook:` cell now spawns a child NotebookProcessor and attaches the result as a SubPipeline node beside the Steps, so every notebook goes through the same compile path whether it references others or not

kale/cli.py
Ya-shh: This is the plan, with @ederign  refinement that it's two phases: a recursive parse a notebook: cell spawns a child processor and attaches a node followed by the existing graph-wide resolve, since dependencies_detection needs the whole DAG before it can resolve
 The SubPipeline node beside Step comes first so the graph can hold both kinds, then the recursion

---

**Update:** implemented in `5bc0b6e`. The processor parses sequentially and, on a `notebook:` cell, recurses into a child processor whose pipeline becomes a SubPipeline node

One thing worth calling out, since it shaped the design: `dependencies_detection` resolves only along edges that already exist, and a composition has no `prev:` tags to create them
So the edges are inferred first (data flow between units, plus reading order for the root notebook's own steps) and then the existing resolution runs unchanged ,that keeps one definition of how a variable crosses a boundary, whether that boundary is two cells or two notebooks

kale/processors/workflow.py
Ya-shh: Done in `f67e190`,  generated modules are kale_notebook_<name>.py, and all three collision guards are deleted: a notebook named json.ipynb, kfp.ipynb, class.ipynb or 2train.ipynb now compiles instead of erroring

 Only the genuine duplicate-name case still raises, since two notebooks resolving to one name would overwrite each other's module

One thing the test caught: the prefix has to drive every generated identifier, not just the filename, task variables were still built from the raw name, so 2train.ipynb produced 2train_task , a syntax error  the rule is now module name for all generated Python identifiers, plain name for the logical identity, display name for the KFP UI
 There's a test that compiles a json + 2kfp pipeline through the real KFP compiler

Ya-shh:
Please review and let me know
  
Phase 1 from @ederign's order is landed `ff590ab, 5e396ff, 3e3cea0, f67e190, 03a016f` the architecture independent fixes, and  @StefanoFioravanzo's tag-form, prefix and merge-loop simplifications ,now  there's a real kfp.compiler compile test per story 

 R1 now fails loudly through the RPC path instead of silently compiling the wrong pipeline
 
 298 backend tests green
 
  Starting the SubPipeline node and the recursion next,  replies on each thread above :)

Ya-shh:
@StefanoFioravanzo @ederign step 3 of the rework is landed in `5bc0b6e`

 Composition now happens inside `NotebookProcessor`, and `processors/workflow.py` plus the `_compile_composition` branch in `cli.py` are deleted
  A `notebook:` cell spawns a child processor and attaches the result as a `SubPipeline` node beside the Steps, so every notebook goes through one compile path
   Edges are inferred before resolution, because `dependencies_detection` only resolves along edges that already exist and a composition has no `prev:` tags to create them
   The resolution itself is unchanged, so a variable crossing a notebook boundary is handled by exactly the code that handles one crossing a cell boundary

R1 is fixed,  The panel's compile RPC was the path that silently dropped references, It now produces a root DAG where each referenced notebook is a nested sub-DAG holding its own cell-level steps, and a mixed root adds its own steps as top-level components, verified against the reproducers in the gist

**Two things I want to flag :**

1. A referenced notebook's `pipeline-parameters` were being dropped, It compiled clean but would have failed at runtime with a `NameError`, since the component signature lost the name its body still used ,fixed in the same commit: each referenced notebook's module is now generated by a compiler bound to that notebook's own pipeline, which is also closer to the per-module design @ederign asked for
2. Per-step config the composition templates do not carry over yet (`limits`, `annotations`, `labels`, retries, timeouts) is now logged by step and field instead of disappearing quietly, carrying it over is step 4, the remaining piece

A `notebook:` cell inside a referenced notebook currently raises ,It used to flatten that level of nesting silently, which I did not think was acceptable, but nested composition is not in the four steps or in the KEP stories either
 Is rejecting it the behaviour you want for now, or should it compile?

Ya-shh:
Step 4 is landed too, in `d1fa92b`, so the four step plan is complete !!

The per-task config block is now a single shared macro imported by every emission path, so a step's `limit:`, `label:`, `annotation:` and `cache:` tags have the same effect whether its notebook is composed or compiled on its own
 previously a referenced notebook's GPU limit was dropped, which meant a step tagged for a GPU silently ran on CPU once composed
  The warning I mentioned above is removed, since the gap it reported is now closed

Single notebook output is unchanged, verified byte for byte against the golden DSL fixtures in `test_e2e.py`

One correction to my note above:
`retry_*` and `timeout` turned out not to be emitted by any Kale template, so they are unused `StepConfig` fields across the project rather than something composition drops
 probably worth a separate look :)

Ya-shh:
@StefanoFioravanzo 

Template: done `subpipeline_template` and `composition_template` are both
deleted
 root, referenced and composing notebooks all render from
`pipeline_template` now
 Two checks that it is the right seam: the compiled KFP IR is byte identical before and after for five pipelines, compositions included, so nothing about what runs moved and it then survived merging #907, where PVC mounting slotted into the shared macro in four lines and composed notebooks got PVC mounts for free

What is genuinely sub-pipeline specific turned out to be four things, not the
three I said
 I found the fourth by breaking it: KFP raises "Platform-specific
features can only be set on primitive components" if you set pod labels or a
security context on a nested pipeline task
 so a sub-pipeline call takes caching and nothing else, and that is KFP's constraint rather than ours

`SubPipeline` I would still keep and that error is part of why, It is not a
pipeline, it is a use of one: reference the same notebook twice and you get two
nodes over two pipelines with their own `ins`, `outs` and config, and only one
of the two kinds of node can carry platform config at all
 Definition vs call site, eager to know your view on this :)

separately, @ederign's second round of reproducers is at six of seven passing

StefanoFioravanzo:
@Ya-shh  congrats on this amazing work, and thanks for sticking with us through all the rounds of reviews! 
I gave another coarse look at the code and architecture, we are definitely in a good spot now. I didn't manage to test the feature flag, I was in a hurry and I won't have the laptop with me for the next few days. But if @ederign or others cross check this and confirm that the feature flag works as expected, I am all on for merging this PR! This is a great milestone and paves the way to the second phase - creating a full fledges low-code notebook orchestration UI :D

ederign:
@Ya-shh merging this. Congrats, it's a big one for Kale. It will open a sea of new possibilities. 


  The part I want to call out isn't the feature, it's that you took "restructure most of it" without starting over. Exactly as we proposed in the beginning.

  Follow-ups, none blocking:

  1. #945 - root pipeline-parameters don't reach steps inside a referenced notebook. Compiles clean, NameError at runtime. Same family as the  four you fixed, and it's the mirror of the one you caught yourself in 5bc0b6e. This is the one I'd take first.
  2. #942 - the kale_notebook_* prefix keys on the pipeline name, and kale --nb defaults that to kale-pipeline, so two compositions in one
  dir still overwrite each other on the CLI path. Panel path is fine.
  3. #944 - docs.

Thanks for taking the feedback well, and the error messages and comments explaining why rather than what is not a common thing to see.

Keep pushing!!

ederign (approved):

PR #868, opened 2026-07-08
fix(frontend): break shared references to DefaultState metadata

### LINKED ISSUE #643: [frontend] DefaultState global object is mutated at runtime, causing shared state between notebooks
### Description                                                                                                                                                                             
   
  `DefaultState` in `LeftPanel.tsx` [reference](https://github.com/kubeflow/kale/blob/9d52f720320fc466783ac3105f7fdf5fecfe009a/labextension/src/widgets/LeftPanel.tsx#L85) is an exported `const` object used as initial state. But inside `setNotebookPanel()` [reference](https://github.com/kubeflow/kale/blob/9d52f720320fc466783ac3105f7fdf5fecfe009a/labextension/src/widgets/LeftPanel.tsx#L263), the code directly mutates it:                  
                                                                                            
  ```typescript
  DefaultState.metadata.base_image = baseImage;
```
  Since the component initializes with state = DefaultState [reference](https://github.com/kubeflow/kale/blob/9d52f720320fc466783ac3105f7fdf5fecfe009a/labextension/src/widgets/LeftPanel.tsx#L110), the component state and the global share the same object reference. Mutating the global also silently mutates
  this.state bypassing React.

  Problems

  1. React state bypassed: Direct mutation without setState, no re-render triggered
  2. Reset doesn't reset: resetState() spreads the mutated global, so "defaults" carry previous values
  3. State leaks between notebooks: Notebooks with no metadata inherit mutated values from previously opened notebooks

  Suggested fix

  Store the detected base image in component state instead of mutating the global. Break the shared reference on initialization.

  ---

ai-naymul:
Fixes #643

The direct mutation from the original report was removed in #754, but the shared references remained: the left panel state was initialized with the same object as `DefaultState.metadata`, and its `steps_defaults` array was shared into every notebook's state. Any in-place write would still leak across notebooks and get persisted into their .ipynb files.

- add `createDefaultMetadata()` so every consumer gets a fresh copy
- freeze `DefaultState` so an accidental in-place write throws instead of silently leaking
- add regression tests

### DIFF
--- a/labextension/src/__tests__/leftPanelTypes.spec.ts
+++ b/labextension/src/__tests__/leftPanelTypes.spec.ts
@@ -1,0 +1,65 @@
+// Copyright 2026 The Kubeflow Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { createDefaultMetadata, DefaultState } from '../widgets/LeftPanelTypes';
+
+// Regression tests for https://github.com/kubeflow/kale/issues/643:
+// notebook state must never share object references with the DefaultState
+// global, otherwise an in-place write leaks into every open notebook.
+describe('createDefaultMetadata', () => {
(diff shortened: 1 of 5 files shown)


jesuino:
Hello @ai-naymul 

Thanks for your contribution! Would you please make the change suggested by copilot and update your branch so we can see the CI output?

Thanks!

jesuino:
Hello @ai-naymul - Please let me know if you had a chance to look into this. Please notice that you also need to resolve conflicts. Thanks!

jesuino:
Hello @ai-naymul 

There's no update since July, hence I am going ahead and closing this. Feel free to re-open when you are back.

Thanks!

ederign:
Closing due to lack of activity. Feel free to reopen as @jesuino mentioned.

PR #874, opened 2026-07-09
feat(backend): auto-promote trailing cell variable to output artifact

### LINKED ISSUE #783: feature: Auto include ouput parameters on a step if the last parameter is a variable
### Feature Area

/area backend

I would like that variables returned by a cell is transformed into output artifacts automatically.

Let's say my notebook results into a variable that I would like to make it accessible to users. Currently I have to add a task just to print that variable, so Kale creates an output artifact which I can download. See this example:

Image

I had to create a new step, but what I wanted was to access the chroma_db object. It would be great if I have a way to make it as an output parameter automatically or using a new property in pipeline which we can use identify the "result artifact" without the need of create a new cell.

Create a cell just to make it sure that Kale will serialize and pass the result variable as an artifact and we will be able to download it.

---

Love this idea? Give it a 👍.

harshhh817:
When the last logical line of a cell is a bare variable (the idiomatic Jupyter way of surfacing a value, e.g. a trailing `chroma_db`), Kale now promotes it to a terminal output artifact even when no downstream step consumes it. Previously such 'result' variables were silently dropped because outs are only marshalled when a descendant step reads them.

- astutils.get_trailing_variable(): detect a trailing bare ast.Name
- nbprocessor.dependencies_detection(): promote it via add_artifact(), reusing KFP_ARTIFACT_TYPE_MAP for type inference
- unit tests for positive/negative cases

Closes #783

### DIFF
--- a/examples/rag/portable_rag_langchain.ipynb
+++ b/examples/rag/portable_rag_langchain.ipynb
@@ -170,7 +170,9 @@
    },
    "source": [
     "### Check the vector DB\n",
-    "On this step we check the produced database and Kale will have to create the chroma_db output artifact to be passed to this, so we can download it from KFP"
+    "On this step we check the produced database and Kale will have to create the chroma_db output artifact to be passed to this, so we can download it from KFP.\n",
+    "\n",
+    "The step ends with a bare `md5_hash` on its last line: Kale automatically promotes a trailing variable like this to a KFP output artifact, so the hash can be inspected and downloaded from the KFP UI even though no later step consumes it."
    ]
   },
   {
@@ -186,7 +188,9 @@
    "outputs": [],
    "source": [
     "md5_hash = hashlib.md5(chroma_db).hexdigest()\n",
-    "print(f'Vector DB created sucessfully. MD5 Hash: {md5_hash}')"
+    "print(f'Vector DB created sucessfully. MD5 Hash: {md5_hash}')\n",
+    "# A bare variable on the last line becomes a KFP output artifact\n",
+    "md5_hash"
    ]
   },
   {
--- a/examples/rag/portable_rag_langchain.ipynb
+++ b/examples/rag/portable_rag_langchain.ipynb
@@ -134,7 +134,16 @@
    },
    "source": [
     "### Create Vector DB\n",
-    "We use the embeddings and the documents from the github repo to build the vector database, zip and store the zip file contents in a variable"
+    "\n",
+    "We use the embeddings and the documents from the GitHub repo to build the\n",
+    "Chroma vector database, zip it, and read the archive into the `chroma_db`\n",
+    "variable. We also compute and print an MD5 hash so the run logs record which\n",
+    "database was produced.\n",
+    "\n",
+    "The cell ends with a bare `chroma_db` on its last line. Because nothing\n",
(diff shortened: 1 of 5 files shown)


ada333 (changes requested):
@harshhh817 great feature, thanks for doing this! Code looks good to me. I tried it out on some basic notebook and seems to be working well. Since this feature can be useful to users I would describe it in the documentation with some easy examples of how it works (something like the parameters on the added test). Also we could add it to some example notebook, so users are aware of this feature? 
@jesuino what do you think?

harshhh817:
Thanks for trying it out @ada333! Added an "Output artifacts" section to `annotating-notebooks.md` with the promotion rules, the type-inference table, and counter-examples matching the test cases — and the RAG example notebook's final step now ends with a bare `md5_hash` to demonstrate the feature in a real pipeline.

ada333 (approved):
LGTM! let's wait what the OG @jesuino thinks

examples/rag/portable_rag_langchain.ipynb
```diff
@@ -186,7 +188,9 @@
    "outputs": [],
    "source": [
     "md5_hash = hashlib.md5(chroma_db).hexdigest()\n",
-    "print(f'Vector DB created sucessfully. MD5 Hash: {md5_hash}')"
+    "print(f'Vector DB created sucessfully. MD5 Hash: {md5_hash}')\n",
```
jesuino, line 173: The MD5 sum being a separated cell is actually an excuse to print the `chroma_db` variable because initially I planned to do it on the `create_vector_database` but I would not have a reason to have access to the `chroma_db` file, so moving the md5_sum to a new step was the temp solution. But this is not a good solution because it will leave to more resource consumption and a more complex pipeline.

What you should do is actually remove the markdown explaining the MD5Sum and the step `check_vector_db`, elaborate more the `create_vector_database` markdown, edit the cell to add the md5sum print and add a last line to the cell which would be the `chroma_db` variable:

Image

jesuino (changes requested):
Hello,

Thank you so much for this contribution. The changes seems to be working as expected and I also don't have too much to say about the code because it looks good!

I only have a comment about the example update. Please take a look and let me know WDYT!

harshhh817:
Thanks @jesuino, that makes sense! Restructured the RAG example: folded the md5sum print into `create_vector_database`, ended that cell with a bare `chroma_db` so it's promoted to an output artifact, and removed the now-redundant `check_vector_db` step and its markdown. Updated the docs reference to match.

jesuino:
Thanks for making the changes @harshhh817 . Please update your branch (sync with main) so the CI will trigger and we can see the results. Thanks!

ederign:
@jesuino I just rebased and retrigger CI, please review it!

jesuino (changes requested):
@harshhh817 

We are almost there - linting is failing, would you please fix it on your end? Please also remember to update the branch correctly.

The lint error:

```
Run make lint-backend
Linting backend...
uv run ruff check kale
All checks passed!
uv run ruff format --check kale
Would reformat: kale/tests/unit_tests/test_ast.py
1 file would be reformatted, 57 files already formatted
make: *** [Makefile:94: lint-backend] Error 1
Error: Process completed with exit code 2.
```

jesuino (approved):
LGTM!

PR #875, opened 2026-07-10
feat(backend): Support KALE_KFP_NAMESPACE for configurable KFP server namespace

### LINKED ISSUE #798: feature(backend): Support KALE_KFP_NAMESPACE for configurable KFP server namespace in KFPServerConfig
Add support for a `KALE_KFP_NAMESPACE` environment variable (or equivalent) so the Kubernetes namespace used for KFP API operations is configurable without hardcoding it in Kale’s server config defaults.
Today, the KFP connection namespace is fixed in code:
- `KFPServerConfig` in `kale/config/kfp_server_config.py`:
  - `namespace = Field(type=str, default="kubeflow-pipelines")` (or `default="kubeflow"` in some versions/branches)
- `get_kfp_client()` in `kale/common/kfp_client_factory.py` repeats a hardcoded fallback:
  - `namespace = namespace or config.namespace or "kubeflow-pipelines"` (or `"kubeflow"`)
Proposed behavior:
- Read the default namespace from `KALE_KFP_NAMESPACE` when building or loading `KFPServerConfig`.
- Keep the existing JSON config file (`~/.config/kale/kfp_server_config.json`, overridable via `KALE_CONFIG_PATH`) as the next source of truth.
- Preserve explicit `namespace=` arguments to `get_kfp_client()` as the highest priority override.
- Use a single canonical default in code (e.g. `kubeflow-pipelines` or document the chosen default) and remove duplicate hardcoded fallbacks in `kfp_client_factory.py`.
- Document the variable next to `KALE_CONFIG_PATH` and other Kale env settings.
- Add unit tests for: unset env (default), env override, config file override, and explicit parameter override.
Example:
```bash
export KALE_KFP_NAMESPACE="kubeflow"
# or for a custom install:
export KALE_KFP_NAMESPACE="ml-pipelines"
```
Kale would then create KFP clients against the correct namespace for that cluster without per-user edits to the config file.

Kubeflow Pipelines is not always installed in the same namespace across environments. Some clusters use `kubeflow`, others `kubeflow-pipelines`, and some teams use custom namespaces for the KFP API server and related resources.

Kale’s Jupyter extension, RPC layer, and CLI paths that call `get_kfp_client()` rely on `KFPServerConfig` defaults when users have not saved a personal `kfp_server_config.json`. With a hardcoded namespace, pipeline upload/list/run operations target the wrong namespace: 404s, permission errors, or empty experiment lists even when auth and host are correct.

We deploy Kale in shared notebook and CI images used across multiple clusters. An environment variable (set via `PodDefault`, deployment env, or CI secrets) is much easier to standardize than baking a different config file per cluster or asking every user to configure Kale manually.

Note: Kale’s `--kubernetes-namespace` CLI flag controls where pipeline run workloads are scheduled; it is separate from the KFP API client namespace in KFPServerConfig. Both may need to differ in multi-user setups, which makes a dedicated, documented env var for the KFP client namespace especially important.

Partial workarounds exist:

1. `~/.config/kale/kfp_server_config.json` — Users can set "namespace": "<your-namespace>" and point to it with `KALE_CONFIG_PATH`. This works but is manual, easy to miss in notebooks, and hard to enforce fleet-wide.
2. Explicit `namespace` in code — Callers that invoke `get_kfp_client(namespace="...")` can override the default, but extension/RPC entry points do not expose this to end users.
3. Fork or patch Kale — Change the `Field` default and factory fallback. Fragile on upgrades and duplicates logic already spread across two files.
4. Wrong namespace guess — Teams on `kubeflow` vs `kubeflow-pipelines` see confusing failures until someone discovers the mismatch; switching branches/versions can change the baked-in default without any env-based escape hatch.

A `KALE_KFP_NAMESPACE` (or similar) would align with how Kale already configures paths (`KALE_CONFIG_PATH`) and pip indexes (`KALE_PYPI_PROD_URL`, etc.), and make multi-cluster deployments predictable.

---

Love this idea? Give it a 👍.

harshhh817:
## What this does

Adds support for a `KALE_KFP_NAMESPACE` environment variable so the Kubernetes namespace used for KFP API operations is configurable without hardcoding it in Kale's server config or asking every user to hand-edit `kfp_server_config.json`.

Closes #798

## Why

KFP is not installed in the same namespace across environments (`kubeflow`, `kubeflow-pipelines`, or custom). With the namespace hardcoded, pipeline upload/list/run operations target the wrong namespace and users hit 404s, permission errors, or empty experiment lists even when host and auth are correct. An env var is much easier to standardize fleet-wide (via `PodDefault`, deployment env, or CI secrets) than baking a per-cluster config file.

## Changes

- **`kale/config/kfp_server_config.py`**
  - Added a single canonical default `DEFAULT_NAMESPACE` and a `get_default_namespace()` helper that reads `KALE_KFP_NAMESPACE`.
  - Changed the `namespace` `Field` default to `None` so `load_config()` can distinguish "user didn't set a namespace" from "user explicitly chose the default", then resolves it via `_apply_default_namespace()`.
- **`kale/common/kfp_client_factory.py`**
  - Removed the duplicate hardcoded `"kubeflow"` fallback. `load_config()` now returns an already-resolved namespace; an explicit `namespace=` argument still wins.
- **`kale/tests/unit_tests/test_kfp_server_config.py`**
  - Added tests for: unset env (default), env override, config-file override, config file without `namespace` falling back to the env var, and explicit-parameter override.

## Resolution order (highest priority first)

1. explicit `namespace=` argument to `get_kfp_client()`
2. `"namespace"` in the JSON config file (`KALE_CONFIG_PATH`)
3. `KALE_KFP_NAMESPACE` environment variable
4. canonical in-code default (`DEFAULT_NAMESPACE`)

This mirrors how Kale already configures paths (`KALE_CONFIG_PATH`).

## Testing

```
uv run pytest kale/tests/unit_tests/test_kfp_server_config.py -q
# 32 passed (24 existing + 8 new)
uv run ruff check   # clean
```

## Note

The issue mentions some branches default to `kubeflow-pipelines`, but `main` currently defaults to `kubeflow`, so I kept `kubeflow` as the canonical default to avoid changing existing behavior. Happy to switch the default if maintainers prefer.


ada333 (changes requested):
hi @harshhh817 from quick look the code looks good - please also add the env var to documentation (https://github.com/kubeflow/kale/blob/main/docs/source/user-guide/running-pipelines.md?plain=1#L104)

harshhh817:
Thanks for the review @ada333! Added `KALE_KFP_NAMESPACE` to the environment variables table in `running-pipelines.md`.

ada333 (approved):
LGTM - I tried setting up the env var and running Kale to see if the target namespace changes, thanks for working on this @harshhh817

kale/config/kfp_server_config.py
```diff
@@ -53,25 +84,40 @@ def get_config_path() -> str:
...
+        KFPServerConfig instance. If no config file exists, returns default
+        config. In all cases the ``namespace`` is resolved to a concrete value
+        using (in priority order) the config file, the ``KALE_KFP_NAMESPACE``
+        environment variable, then the canonical in-code default.
     """
     config_path = get_config_path()
     if not os.path.exists(config_path):
         log.debug("No KFP server config found at %s, using defaults", config_path)
-        return KFPServerConfig()
+        return _apply_default_namespace(KFPServerConfig())
```
jesuino: I see multiple calls to this method, could this be done in the _postprocess method? learn more about it in kale/config/config.py:

```
    Validators are useful for single-field checks. To performs post
    initialization/validation checks, over multiple fields, or postprocess the
    validated fields, override the `_validate` and `_postprocess` functions,
    respectively.
```
Here's a suggested solution (not tested):
```
    def _postprocess(self):
        """Fill in the default namespace if not explicitly set."""
        if self.namespace is None:
            self.namespace = get_default_namespace()
```

jesuino (changes requested):
Hello,

The changes looks good and it seems to work well. I have a possible suggestion to avoid code duplication, would you please check it?

Thanks!

harshhh817:
Thanks @jesuino! Moved the default-namespace resolution into `KFPServerConfig._postprocess`, which removes the `_apply_default_namespace` helper and its three call sites in `load_config`. All config unit tests pass.

ederign:
@jesuino I just fire up the CI!

ederign:
@jesuino can you please take a look?

ederign:
@harshhh817 you need to fix lint for this pR!

jesuino (approved):
LGTM! It is working as expected

jesuino:
@harshhh817 you need to update your branch and make the lint so we can merge this, would you please do that?

harshhh817:
Thanks both — rebased onto current main (23 commits) and the lint format issue is fixed. Suite passes (276) and `ruff check`/`ruff format --check` are both clean. Ready for another look.

PR #888, opened 2026-07-20
docs: Add KEP-843 for mounting PVCs

ada333:
Adds KEP-843: Mount PersistentVolumeClaims in Pipeline Steps.

Please read through the document and share your thoughts, also check out the open questions, thanks!

@Fruchix @ederign @StefanoFioravanzo

### DIFF
--- a/docs/proposals/0843-mount-volumes/README.md
+++ b/docs/proposals/0843-mount-volumes/README.md
@@ -1,0 +1,558 @@
+# KEP-0843: Mount PersistentVolumeClaims in Pipeline Steps
+
+|                     |                                                    |
+| ------------------- | -------------------------------------------------- |
+| **Authors**         | [ada333](https://github.com/ada333)                |
+| **Created**         | 2026-07-20                                         |
+| **Status**          | Provisional                                        |
+| **Relevant Issues** | https://github.com/kubeflow/kale/issues/843  https://github.com/kubeflow/kale/issues/856      |
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [User Stories](#user-stories)
+- [Background: How Kale Handles Data Today](#background-how-kale-handles-data-today)
+- [Proposal](#proposal)
+  - [Architecture](#architecture)
+- [Design Details](#design-details)
+  - [Frontend — Left Panel](#frontend--left-panel)
+  - [Backend — Compiler](#backend--compiler)
+  - [Backend — RPC](#backend--rpc)
+  - [CLI](#cli)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation Plan](#implementation-plan)
+- [Migration](#migration)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives Considered](#alternatives-considered)
+- [Consequences](#consequences)
+
+---
+
+## Summary
+
+Add support for mounting Kubernetes PersistentVolumeClaims (PVCs) into Kale
+pipeline step Pods via notebook metadata and the JupyterLab extension UI. Users
+will be able to configure 0–N volumes per pipeline in the Left Panel sidebar.
+The Kale compiler will add a `kfp.kubernetes.mount_pvc(...)` call for every
+volume in notebook metadata on every pipeline step.
+
+---
+
+## Motivation
+
+When running AI/ML workloads, large datasets and model artifacts are often
+stored on high-performance storage exposed in Kubernetes as PVCs. Kale's
+current KFP v2 code path (introduced in 2.x) does not support mounting PVCs
+into pipeline step Pods. Users are forced to either:
+
+1. Hand-edit the generated DSL to add `kubernetes.mount_pvc(...)` for every
+   step after compilation, or
+2. Route large data through Kale marshal + the KFP artifact store, which is
+   inefficient for files that are already on a mounted filesystem.
+
+Beyond inefficiency, pickle-based marshal has hard failure modes that a shared
+PVC would avoid entirely:
+
+- **Large datasets and models** — serializing and deserializing gigabyte-scale
+  objects between steps is slow and often runs out of memory or artifact store
+  quota.
+- **Non-serializable Python objects** — database connections, file handles, and
+  certain framework objects (e.g. some model classes) cannot be pickled at all;
+  passing them via a shared path sidesteps the problem.
+- **Ephemeral shared state** — scratch space written by one step and consumed
+  by the next (e.g. a temporary data store or checkpoint directory) has no
+  natural representation as a KFP artifact.
+
+Kale 1.x had a volumes feature using KFP v1 APIs (`VolumeOp`, `PipelineVolume`,
+`add_pvolumes`). All of those APIs were removed in KFP v2. This KEP revives the
+volumes feature using the current `kfp-kubernetes` extension library.
+
+### Goals
+
+1. Let users configure 0–N volumes (PVC name + mount path) per pipeline in the
+   JupyterLab Left Panel sidebar.
+2. Persist that configuration in notebook metadata.
+3. Have the Kale compiler emit `kubernetes.mount_pvc(...)` for every step in
+   the generated KFP v2 pipeline.
+4. Optionally pre-fill volumes from the PVCs already mounted on the running
+   notebook Pod ("use this notebook's volumes").
+5. Show a combobox of available PVCs in the namespace (with suggestions from the
+   cluster) alongside free-text input, always accessible regardless of API availability.
+
+### Non-Goals
+
+- Per-step volume selection — all steps receive the same set of volumes in v1.
+- Automatically ordering parallel steps that share files on a PVC.
+- Creating new PVCs from the Kale UI (`new_pvc`, `VolumeOp`) — mount existing
+  PVCs only.
+- Using PVCs as the Kale marshal transport — marshal continues to use KFP
+  artifacts.
+- Making PVC existence a hard compiler requirement — offline/CI compile must
+  still work.
+
+---
+
+## User Stories
+
+### Story 1: Notebook volume available in pipeline
+
+A user creates a notebook in Kubeflow and mounts a PVC named `my-data` at
+`/data` via the Kubeflow UI. Code in the notebook reads from `/data` fine.
+When they compile a Kale pipeline, the step Pods do not have `/data` mounted
+and the pipeline fails.
+
+With this feature, the user clicks "Use notebook volumes" in the Volumes panel.
+Kale calls the existing `nb.list_volumes` RPC, detects that `my-data` is
+mounted at `/data`, and pre-populates a volume entry. All pipeline steps then
+get `my-data` mounted at `/data`.
+
+### Story 2: Mount an arbitrary PVC
+
+A user knows their training data lives on PVC `training-data`. Their notebook
+does not have it mounted. They open the Volumes panel, click "Add Volume",
+select `training-data` from the PVC dropdown (or type it manually), set mount
+path `/data`, and compile. All pipeline steps get `training-data` mounted at
+`/data`.
+
+### Story 3: Multiple volumes on all steps
+
+A user needs datasets at `/data` (PVC `datasets`) and wants to write model
+checkpoints to `/models` (PVC `model-store`). They add two volume entries. Both
+PVCs are mounted at both paths on every step.
+
+```python
+# Inside a pipeline step — just normal file I/O, no Kale API needed
+df = pd.read_csv("/data/train.csv")       # reads from datasets PVC
+model.save("/models/checkpoint.pt")       # writes to model-store PVC
+```
+
+---
+
+## Background: How Kale Handles Data Today
+
+### Two data paths
+
+Today Kale has one mechanism for moving data between pipeline steps. This KEP
+introduces a second. They serve different purposes:
+
+| Mechanism | For what | How data moves |
+|-----------|----------|----------------|
+| **Kale marshal + KFP artifacts** (existing) | Python variables between steps (`ins`/`outs`) | serialize → object store → deserialize |
+| **PVC mounts (this KEP)** | Files on a shared filesystem (`/data/...`) | mounted volume; code uses paths directly |
+
+These two paths are complementary, not competing. Once this KEP is implemented,
+a step can simultaneously read `/data/train.csv` from a PVC and pass a
+processed `DataFrame` to the next step via Kale's normal marshal mechanism.
+
+### Why PVCs?
+
+The KFP artifact store round-trip (serialize → upload → download → deserialize)
+is negligible for small objects but expensive for large datasets, model
+checkpoints, or files that are already on disk in the right format. A PVC mount
+avoids that cost entirely: step B opens the same file step A wrote, with no
+intermediate copy.
+
+### Kale 1.x volumes (historical reference)
+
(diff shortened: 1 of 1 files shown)


docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+   JupyterLab Left Panel sidebar.
+2. Persist that configuration in notebook metadata.
+3. Have the Kale compiler emit `kubernetes.mount_pvc(...)` for every step in
+   the generated KFP v2 pipeline.
+4. Optionally pre-fill volumes from the PVCs already mounted on the running
+   notebook Pod ("use this notebook's volumes").
+5. Show a combobox of available PVCs in the namespace (with suggestions from the
+   cluster) alongside free-text input, always accessible regardless of API availability.
+
+### Non-Goals
```
hmtosi, line 96: Are "Non Goals" things you're trying to avoid, or are they something you want to implement in the future?

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+processed `DataFrame` to the next step via Kale's normal marshal mechanism.
+
+### Why PVCs?
+
+The KFP artifact store round-trip (serialize → upload → download → deserialize)
+is negligible for small objects but expensive for large datasets, model
+checkpoints, or files that are already on disk in the right format. A PVC mount
+avoids that cost entirely: step B opens the same file step A wrote, with no
+intermediate copy.
+
```
hmtosi, line 197: I remember that the marshaling only worked on certain object types, and we were going to have to custom build some to accommodate new data types. Do you know if the PVCs will be object agnostic? In other words, will this help us get around the marshaling issue, or will certain objects still take extra code?

hmtosi (commented):
@ada333 this looks great, thank you for making it!

I just have a couple clarify questions that I am curious about.

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+Beyond inefficiency, pickle-based marshal has hard failure modes that a shared
+PVC would avoid entirely:
+
+- **Large datasets and models** — serializing and deserializing gigabyte-scale
+  objects between steps is slow and often runs out of memory or artifact store
+  quota.
+- **Non-serializable Python objects** — database connections, file handles, and
+  certain framework objects (e.g. some model classes) cannot be pickled at all;
+  passing them via a shared path sidesteps the problem.
```
StefanoFioravanzo: Don't you need to serialize these objects into the PVC just like you would with an artifact?

docs/proposals/0843-mount-volumes/README.md
StefanoFioravanzo: I hope it's the latter ahah

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+A user creates a notebook in Kubeflow and mounts a PVC named `my-data` at
+`/data` via the Kubeflow UI. Code in the notebook reads from `/data` fine.
+When they compile a Kale pipeline, the step Pods do not have `/data` mounted
+and the pipeline fails.
+
+With this feature, the user clicks "Use notebook volumes" in the Volumes panel.
+Kale calls the existing `nb.list_volumes` RPC, detects that `my-data` is
+mounted at `/data`, and pre-populates a volume entry. All pipeline steps then
+get `my-data` mounted at `/data`.
```
StefanoFioravanzo: I would add the ability to select which volumes are persisted in the pipeline. We could add a dialong at compile time that allows me to select all, none, or some of the notebook's volumes.

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+### Story 2: Mount an arbitrary PVC
+
+A user knows their training data lives on PVC `training-data`. Their notebook
+does not have it mounted. They open the Volumes panel, click "Add Volume",
+select `training-data` from the PVC dropdown (or type it manually), set mount
```
StefanoFioravanzo, line 138: I think the "dropdown" part of essential for a nice UX. Good that you mentioned it, I just wanted to highlight it

StefanoFioravanzo: when it comes to the mountpath - what about using env vars to generalize the mounting. We could offer two modes: 

1. You define you own mount path, you hardcoded it in your notebook
2. You use a predefined KALE env var in your notebook. Kale populated the env var at runtime with the mountpath we select. This, on one hand, avoids collisions with existing paths, and second, it allows you to create a reusable notebook, also working with multiple mountpaths or local paths between testing and prod.

docs/proposals/0843-mount-volumes/README.md
StefanoFioravanzo: the logic to serialize an object into an artifact or into a pvc file is exactly the same. So I don't feel like that adding PVCs solves the problem of un-serializable objects or to have to implement specialized backends

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+### Non-Goals
+
+- Per-step volume selection — all steps receive the same set of volumes in v1.
+- Automatically ordering parallel steps that share files on a PVC.
+- Creating new PVCs from the Kale UI (`new_pvc`, `VolumeOp`) — mount existing
+  PVCs only.
+- Using PVCs as the Kale marshal transport — marshal continues to use KFP
+  artifacts.
+- Making PVC existence a hard compiler requirement — offline/CI compile must
+  still work.
```
StefanoFioravanzo, line 105: something that might be a non-goal for this v1, but that we should definitely track in the Volumes story, is support for native K8s snapshots. If a storage provides supports the snapshot API, it would be great if Kale gave you the option to mount a **new** PVC from a snapshot of the local volumes, or another PVC entirely. This allows you to create a lineage of your work, storage-wise. This feature opens a whole new set of potential features around reproducibility

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+   variable edge between them), the user must either introduce a variable
+   dependency or manually add `.after()` in the generated DSL. This is an edge
+   case — in tagged notebooks, most ordering is already implied by variable
+   flow.
+
+4. **RWO volumes and parallel steps.** A `ReadWriteOnce` (RWO) PVC can
+   typically be mounted read-write by only one node at a time. If the scheduler
+   places parallel steps on different nodes, the second Pod may fail to attach
+   the volume. Users with large datasets should prefer `ReadWriteMany` (RWM)
+   storage classes for shared pipeline data.
```
StefanoFioravanzo, line 434: I suggest we add implementation notes so that we add at least a warning about it in the left panel. Or we even add a detection mechanism so that if a selected pvc is RWO, we add a big fat warning. Lauinching a pipeline that mounts a RWO should probably be couple with the option to add a node affinity rule to the pipeline steps, unless nowadays K8s handles thins transparently (dunno, I haven't looked at this stuff in a while)
I had my fair share of pipelines hanging there while waiting for unmountable PVCs in the past :)

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+---
+
+## Open Questions
+
+1. **Should deprecated volume-creation flags be deleted?**
+   The `--storage-class-name` and `--volume-access-mode` flags are currently
+   deferred as out of scope for v1 (they only apply to PVC creation). Should
+   they be removed from the codebase now to avoid confusion, or left in place
+   in case PVC creation support is added in a future version?
```
StefanoFioravanzo: I feel like this is an opportunity to clean up old/dead code related to volume management. I +1 deleting

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+   Currently a missing or wrong-namespace PVC would only surface as a Kubernetes
+   mount error when the step Pod is scheduled — the error message is not
+   user-friendly. A runtime check could give a clearer error earlier. Two
+   approaches are possible:
+   - A **dedicated first step** in the generated pipeline that calls the
+     Kubernetes API to verify all configured PVCs exist before any compute
+     steps run.
+   - An **init container** injected into every step Pod via `kfp-kubernetes`
+     that checks its own mounts before the main container starts.
+   Is the improved error message worth the added complexity and latency?
```
StefanoFioravanzo: I feel like this MIGHT be useful in certain scenarios, but not worth it in the v1. Also, there might be some many different use cases.. also this would need configuration from the Kale UI, and I think it might not be universal and rather an advanced behaviour that would need to be toggled on

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+3. **Should Kale warn users about RWO volumes and parallel steps?**
```
StefanoFioravanzo: YES, and even more, and I commented above. I feel this could be source of major frustration if not handled well. Many folks might not understand deeply the nuances of how volumes work in kubernetes

StefanoFioravanzo:
@ada333 I added something comments, thank you for creating this KEP, it's great!

docs/proposals/0843-mount-volumes/README.md
Fruchix: From how database connections and file handles work, from https://kale.kubeflow.org/en/latest/concepts/data-passing.html#non-serializable-objects, and from the fact that each pipeline step is running in its own pod, I think the serialization of those objects is never possible, a PVC wont change that.
Each step should recreate its own DB connections, file handles, etc.

docs/proposals/0843-mount-volumes/README.md
Fruchix: Completely agree with the big fat warning, that is the type of error that might happen even after reading the docs completely and not understanding why it fails.
The thing is, I am not sure where the detection should be implemented, as if there are RBAC or cluster access problems then the information would not be available in the UI directly.
It would be great to have 2 levels of warning, one in the UI if RBAC and cluster access work, and a well written warning within the pipeline at runtime.

docs/proposals/0843-mount-volumes/README.md
Fruchix: I really like this env var idea

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+   cluster) alongside free-text input, always accessible regardless of API availability.
+
+### Non-Goals
+
+- Per-step volume selection — all steps receive the same set of volumes in v1.
+- Automatically ordering parallel steps that share files on a PVC.
+- Creating new PVCs from the Kale UI (`new_pvc`, `VolumeOp`) — mount existing
+  PVCs only.
+- Using PVCs as the Kale marshal transport — marshal continues to use KFP
+  artifacts.
```
Fruchix, line 103: From all the non-goals, I think this one would be the most important in the future.
Some framework objects could be huge when doing model training (for example when loading the datasets), and the default marshal serialization might slow down the pipeline.

docs/proposals/0843-mount-volumes/README.md
Fruchix: Is it also the case if multiple steps use the same object?
For example if a pipeline step loads `train.csv` into a DataFrame, and N pipeline steps use this DataFrame, does that require multiple uploads or only once to everybody?
With either marshall or PVC the serialization, download and deserialization stages always need to be done, but maybe we can save time during the upload stage?

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+**Interactions:**
+
+| Action | Behaviour |
+|--------|-----------|
+| Add Volume | Appends a blank `{ name: '', mount_point: '' }` entry to `metadata.volumes` |
+| Remove (×) | Drops that entry from the list |
+| PVC field | Combobox — free-text input with suggestions populated from `list_pvcs` RPC; suggestions are shown when available but typing a name manually always works |
+| Mount path | Free-text; must be unique across entries |
+| "Use notebook volumes" button | Calls the existing `nb.list_volumes` RPC and pre-fills entries from the notebook Pod's own PVC mounts; logs each added volume (name + mount point) to the browser console so the user can see exactly what was detected |
```
Fruchix: This button could open a sub menu to support the volume selection (see comments from Stefano at lines R111 to R114).
Maybe: 
```
Kale Deployment Panel
├── Pipeline Metadata  (experiment, name, description, docker image)
├── Volumes
│   ├── [ + Add Volume ]
│   │   ├ [ "Select from notebook" button ]
│   │   │   └ Dropdown menu of notebook volumes, selecting one prefills the two fields below
│   │   ├ [ PVC dropdown/input (can be prefilled using above button) ] [ PVC mount point ]
│   │   └ [ "Cancel" button ] [ "Add volume" button ]
│   ├── Volume 1: [ PVC dropdown/input ] [ /data   ] [×]
│   └── Volume 2: [ PVC dropdown/input ] [ /models ] [×]
└── Deploy button / progress
```

docs/proposals/0843-mount-volumes/README.md
```diff
@@ -0,0 +1,558 @@
...
+   typically be mounted read-write by only one node at a time. If the scheduler
+   places parallel steps on different nodes, the second Pod may fail to attach
+   the volume. Users with large datasets should prefer `ReadWriteMany` (RWM)
+   storage classes for shared pipeline data.
+
+5. **PVC existence is not checked at compile time.** A missing or
+   wrong-namespace PVC causes the step Pod to fail at runtime with a standard
+   Kubernetes mount error surfaced in the KFP UI. This is intentional: offline
+   and CI compile must work without cluster access. No runtime pre-check is
+   performed today either — see Open Questions.
```
Fruchix, line 440: In the future, it could be interesting to add a missing PVC warning (not an error), simply to be able to detect those without failing offline/CI compile

Fruchix (commented):
I really like all your propositions and the caveats/edge cases identification

docs/proposals/0843-mount-volumes/README.md
ada333: Yes, it is meant for later implementation, not necessary in v1 :)

docs/proposals/0843-mount-volumes/README.md
ada333: yes, agreed.

docs/proposals/0843-mount-volumes/README.md
ada333: if the N pipeline steps use it, only 1 upload is made to MinIO storage and N downloads are made. If we will use PVCs as marshal transport , you'd eliminate both — 1 write to the shared volume, N local reads.

docs/proposals/0843-mount-volumes/README.md
ada333: added to non-goals

docs/proposals/0843-mount-volumes/README.md
ada333: I like it too, I added it to KEP

docs/proposals/0843-mount-volumes/README.md
ada333: agreed, added a UI and runtime warning to goals

ada333:
@hmtosi @Fruchix @StefanoFioravanzo thank you for your comments! I tried to answer/implement all of them. Let me know if anything else needs to be done.

ederign:
@ada333 as @StefanoFioravanzo engaged with this one, if he approves it's good on my side :)

hmtosi (approved):
/lgtm
@ada333 thank you for this great work! The changes look good

Fruchix (approved):
Everything is fine with me. Previous comments have been appropriately answered.
The KEP is good and well defined, thank you @ada333!

jesuino (approved):

PR #889, opened 2026-07-20
feat: Allow users to hide the HTML Report

jesuino:
## Allow Users to Hide the HTML Report
This branch adds a new feature that allows users to disable HTML report generation on a per-step basis in Kale pipelines.

### Backend (Python):

* Added generate_html_report field to StepConfig in step.py
* Added report:enabled|disabled tag parsing in nbprocessor.py
* Modified compiler.py to conditionally include the HTML report output parameter
* Updated Jinja2 template nb_function_template.jinja2 to skip HTML artifact generation when disabled

### Frontend (TypeScript/React):

* Added new HtmlReportDialog component in dialogs/HtmlReportDialog.tsx
* Added "REPORT" button to CellMetadataEditor to open the dialog
* Extended TagsUtils to parse and serialize report: tags
* Updated useCellTags hook with updateHtmlReport function
* Added report: to known tag prefixes in constants

### Usage

Users can now add a report:disabled tag to step cells to skip HTML report generation, reducing pipeline artifact overhead when the rendered notebook output isn't needed.

Example, the following notebook:
image

Would generate an HTML report for both the first cell and the second, but the first HTML report would be generated but without any content:

image

Disabling the HTML output now generates a different pipeline with no HTML output for the first cell
image

image

### DIFF
--- a/labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
+++ b/labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
@@ -260,7 +260,7 @@ export const CellMetadataEditor: React.FC<IProps> = props => {
                     onClick={() => setHtmlReportDialogOpen(true)}
                     style={{ width: '5%' }}
                   >
-                    REPORT
+                    HTML REPORT
                   </Button>
                 </div>
               </>
--- a/labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
+++ b/labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
@@ -27,6 +27,7 @@ import { SelectMulti } from '../../components/SelectMulti';
 import { GpuDialog } from './dialogs/GpuDialog';
 import { BaseImageDialog } from './dialogs/BaseImageDialog';
 import { CacheDialog } from './dialogs/CacheDialog';
+import { HtmlReportDialog } from './dialogs/HtmlReportDialog';
 import { useUpdateCellTags } from './hooks/useCellTags';
 import { useEditorPosition } from './hooks/useEditorPosition';
 import {
@@ -48,6 +49,7 @@ export interface ICellEditorData {
   limits?: { [id: string]: string };
   baseImage?: string;
   enableCaching?: boolean;
+  generateHtmlReport?: boolean;
 }
 
 export interface IProps extends ICellEditorData {
@@ -62,6 +64,7 @@ export const CellMetadataEditor: React.FC<IProps> = props => {
     limits = {},
     baseImage,
     enableCaching,
+    generateHtmlReport,
     resolvedDefaultBaseImage,
   } = props;
 
@@ -77,6 +80,7 @@ export const CellMetadataEditor: React.FC<IProps> = props => {
     limits,
     baseImage,
     enableCaching,
+    generateHtmlReport,
   });
 
   useEditorPosition(editorRef, notebook);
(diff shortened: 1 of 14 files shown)


labextension/src/widgets/cell-metadata/dialogs/HtmlReportDialog.tsx
```diff
@@ -0,0 +1,89 @@
...
+    }
+  }, [open, generateHtmlReport]);
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const val = e.target.value as HtmlReportValue;
+    setReportValue(val);
+    onUpdateHtmlReport(val === 'disabled' ? false : undefined);
+  };
+
+  return (
```
ada333, line 58: The HTML Dialog is very similar to the cache dialog, we could create new class that these two would use, would be nice - but it is not necessary.

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
```diff
@@ -244,6 +249,20 @@ export const CellMetadataEditor: React.FC<IProps> = props => {
...
+                  <Button
+                    disabled={!hasStepName}
+                    color="primary"
+                    variant="contained"
+                    size="small"
+                    title="HTML Report"
+                    onClick={() => setHtmlReportDialogOpen(true)}
+                    style={{ width: '5%' }}
+                  >
+                    REPORT
```
ada333: I feel like REPORT is not very clear for first time users, but there are not many options on how to call it, since it needs to be short - for the button to look like the other buttons on CellMetadata. Maybe HTML?

ada333 (approved):
LGTM, I tried it out and seems to be working great. I have two comments but both are optional.

labextension/src/widgets/cell-metadata/dialogs/HtmlReportDialog.tsx
jesuino: I think we should do some UX for this editor because it is getting too overwhelming for the user to see all the changes. I did create a RFE:

https://github.com/kubeflow/kale/issues/894

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
jesuino: I do agree that Report may be confusing, could not think of any other better name, but HTML is even more confusing IMO. Perhaps something like: "Generate Output" or something like that

labextension/src/widgets/cell-metadata/dialogs/HtmlReportDialog.tsx
ada333: good point, I agree

labextension/src/widgets/cell-metadata/CellMetadataEditor.tsx
ederign: Why not HTML report?

ederign (changes requested):
William, two things: 

- No tests: The PR adds no unit tests for the new tag parsing, compiler conditional, or template rendering. Given there are existing tests in `kale/tests/unit_tests/test_parser.py` and `test_rpc_compile.py`, could you please address it?

2. Template formatting nit: The Jinja2 diff produces `{%- endif %}{%- if step.outs|length > 0 %}` on a single line, functionally fine but reduces readability.

jesuino:
thank you for your review, @ederign  - would you please check my changes? I added a few tests, modified the template and the name of the button to "HTML Report"

Thanks!

ederign (approved):
Thanks @jesuino

PR #892, opened 2026-07-21
fix(backend): handle async functions in AST parsing utilities

harshhh817:
## What this does

Treats `ast.AsyncFunctionDef` the same as `ast.FunctionDef` in `kale/common/astutils.py`, so `async def` helpers in a `functions`/`step` cell are handled correctly by:

- `parse_functions` — now extracts async functions (previously dropped them).
- `get_function_and_class_names` — now registers async function names.
- `get_marshal_candidates` — now stops at async function bodies, so it no longer leaks their local variables and correctly adds the function name.

## Why

`astutils` only checked `ast.FunctionDef`, so an `async def` was mishandled in three ways. Reproducer on `main`:

```python
from kale.common import astutils

astutils.parse_functions("async def foo():\n    return 1")
# {}                      -> expected {"foo": ...}

astutils.get_function_and_class_names("async def foo():\n    pass")
# set()                   -> expected {"foo"}

astutils.get_marshal_candidates("async def foo():\n    local_x = 1\n    return local_x")
# ['local_x']             -> expected ['foo']
```

The `get_marshal_candidates` case is the most harmful: it walked into the async function body and returned the function's **local** variable as a marshal candidate while missing the function name. In a pipeline this means an `async def` helper would not be injected into steps, would be treated as a missing dependency, and its internals could be marshalled between steps.

## Testing

Extended `kale/tests/unit_tests/test_ast.py` with an async snippet and a dedicated `parse_functions` async test. All three new assertions fail on `main` and pass with this change; the full `test_ast.py` suite (47 tests) passes.

I came across this while working on the output-artifact/AST code in #874. Happy to open a tracking issue if maintainers prefer.

### DIFF
--- a/kale/common/astutils.py
+++ b/kale/common/astutils.py
@@ -128,6 +128,7 @@ def get_marshal_candidates(code):
     #  Context manager (just the alias)
     contexts = (
         ast.FunctionDef,
+        ast.AsyncFunctionDef,
         ast.ClassDef,
     )
     tree = ast.parse(commented_code)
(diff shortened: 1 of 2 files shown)


harshhh817:
Friendly ping — small backend fix: `astutils` only checked `ast.FunctionDef`, so an `async def` helper was dropped by `parse_functions`, its name was missing from `get_function_and_class_names`, and `get_marshal_candidates` walked into its body and returned the function's *local* variables as marshal candidates.

Regression tests included; still green. Happy to rebase onto current main if that helps.

PR #866, opened 2026-07-01
feat(backend): support KALE_PYPI_PROD_URL for configurable production PyPI index

Bhavd33p:
## Summary
- `compute_pip_index_urls()` hardcoded the production pip index fallback to `https://pypi.org/simple`, appended unconditionally regardless of `KALE_PIP_INDEX_URLS`/`KALE_DEV_MODE` overrides
- Add `KALE_PYPI_PROD_URL` env var to make this fallback configurable, defaulting to `https://pypi.org/simple` for backward compatibility
- Documented alongside the other `KALE_PIP_*`/`KALE_DEV_MODE` vars in `docs/source/user-guide/running-pipelines.md` and `docs/source/api/cli.md`

Fixes #797

## Test plan
- [x] `pytest kale/tests/unit_tests/test_utils.py -k compute_pip_index_urls -v` — 8 passed (5 existing + 3 new)
- [x] Full unit suite: `pytest kale/tests/unit_tests/` — 247 passed
- [x] Manual check: `KALE_PYPI_PROD_URL` set → used in place of `pypi.org`; unset → falls back to `https://pypi.org/simple`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #893, opened 2026-07-21
fix(backend): avoid duplicate PyPI index url in compute_pip_index_urls

harshhh817:
## What this does

Fixes `compute_pip_index_urls` so the production PyPI index (`https://pypi.org/simple`) is added exactly once.

## Why

The tail of the function was:

```python
urls.append(pypi_prod_url)              # unconditional append

# remove duplicates while keeping order
if pypi_prod_url not in urls:           # always False here — dead code
    urls.append(pypi_prod_url)
return urls
```

Because `pypi_prod_url` was appended unconditionally just above, the `if pypi_prod_url not in urls` guard was always false and never deduplicated anything. When a user's `KALE_PIP_INDEX_URLS` already included `https://pypi.org/simple`, it ended up in the list twice:

```python
KALE_PIP_INDEX_URLS="https://pypi.org/simple,http://myindex/simple"
# before: ['https://pypi.org/simple', 'http://myindex/simple', 'https://pypi.org/simple']
# after:  ['https://pypi.org/simple', 'http://myindex/simple']
```

The fix drops the unconditional append and keeps only the guarded one, so PyPI is appended only when the user hasn't already listed it, preserving their chosen position.

## Testing

Added `test_compute_pip_index_urls_override_includes_pypi` covering the override-includes-PyPI case. It fails on `main` and passes with this change; the existing `compute_pip_index_urls` tests continue to pass.

Found while reading `kale/common/utils.py`. Happy to open a tracking issue if maintainers prefer.

### DIFF
--- a/kale/common/utils.py
+++ b/kale/common/utils.py
@@ -250,11 +250,9 @@ def compute_pip_index_urls() -> list[str]:
                 )
             )
 
-    # important to keep the prod url at the end to preserve package
-    # resolution order.
-    urls.append(pypi_prod_url)
-
-    # remove duplicates while keeping order
+    # Keep the prod url at the end to preserve package resolution order, but
+    # only add it when the user hasn't already included it, so it is not
+    # duplicated in the resulting list.
     if pypi_prod_url not in urls:
         urls.append(pypi_prod_url)
     return urls
--- a/kale/tests/unit_tests/test_utils.py
+++ b/kale/tests/unit_tests/test_utils.py
@@ -128,6 +128,21 @@ def test_compute_pip_index_urls_override_beats_dev_mode(monkeypatch):
     ]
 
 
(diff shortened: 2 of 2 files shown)


harshhh817:
Closing this — it's been superseded by #866, which fixed the same duplication while making the prod index URL configurable.

Main now does:

```python
urls = [u for u in urls if u != pypi_prod_url]
urls.append(pypi_prod_url)
```

which is better than what I proposed here: my version only skipped the append when the URL was already present, so a user-supplied `https://pypi.org/simple` would have stayed wherever it appeared in the list instead of being moved last. #866 both de-duplicates and preserves the resolution order. Verified on current main that `KALE_PIP_INDEX_URLS` containing the prod URL yields no duplicate and keeps it last.

Thanks!

PR #908, opened 2026-07-28
feat: consolidate per-step config buttons into a single Configure dialog

### LINKED ISSUE #894: [feature] Unified Step Configuration Panel
### Feature Area

 /area frontend

Currently we have multiple buttons on top of a step and each button opens a different dialog. 

Image

The feature is to have a single configuration button when clicked open a single dialog with all the options, something like:

Image

* It is hard to maintain different classes for the dialogs in the UI
* Users may get confused with all the different options
* Provides sustainable growth of options because the cell pane is running out of space

n/a

---

Love this idea? Give it a 👍.

cordeirops:
## Summary

Closes #894 - consolidates the per-step config buttons (Base Image,
GPU, Caching) into a single "Configure step" gear-icon button and
dialog, instead of one button + one dialog per option.

This was flagged during review of #891 (Secrets): with #889 (Report)
and #891 both adding their own button/dialog, the cell toolbar was
about to have five separate config buttons. This PR gives both of
those a single tabbed dialog to plug into instead.

Note: #891 depends on this PR and is blocked until it merges.

## Changes

- New `StepConfigDialog.tsx`: one Dialog with tabs (Base Image / GPU /
  Caching), replacing `BaseImageDialog.tsx`, `GpuDialog.tsx`, and
  `CacheDialog.tsx` (all deleted). Each tab's content is unchanged from
  its original dialog, just extracted into a section component.
- Tabs are driven by a `{label, render}[]` array that `Tabs`/
  `DialogContent` map over. Adding a new option going forward (e.g.
  Report from #889, Secrets from #891) means appending one entry to
  that array, not adding another button/dialog pair or a hardcoded tab
  index.
- `CellMetadataEditor.tsx`: the three buttons + three open-state
  variables became one gear-icon `IconButton` (`aria-label="Configure
  step"`) + one open-state, matching the existing icon-button style
  already used for close/clear-metadata in the same toolbar.
- Updated the Playwright e2e assertion that checked for the three
  separate button titles to check for the new button's aria-label.

No backend or notebook-tag-format changes - this is a pure UI
reorganization.

image
image
image

## Fixed during self-review

Caught and fixed before opening this PR:
- The Caching tab's local state was resynced via a `useEffect` that
  fired on remount (since inactive tabs unmount), causing a one-frame
  flash of the wrong radio selection when switching tabs. Removed the
  local state entirely - the value is now derived directly from the
  `enableCaching` prop.
- "Reset to Default" on the Base Image tab had lost the `onClose()`
  call the original dialog made alongside the reset.
- The dialog didn't reset to the first tab when reopened, so it could
  reopen on whichever tab was last viewed (e.g. for a different step).
- The unified dialog had dropped the original GPU dialog's
  `aria-labelledby`/`aria-describedby` accessibility attributes.

## Test plan

- [x] `make test-labextension` (typecheck, eslint, jest)
- [x] Manual test in a running JupyterLab: opened the dialog, verified
      all three tabs render pre-populated from existing tags
      (`limit:`, `cache:`, `image:`), verified each tab writes the
      correct tag on change, verified Reset to Default closes the
      dialog, verified the dialog reopens on Base Image after being
      left on another tab, verified no flash on the Caching tab, and
      verified `aria-labelledby`/`aria-describedby` are present.

🤖 Generated with [Claude Code](https://claude.com/claude-code), reviewed by @Cordeirops


labextension/src/widgets/cell-metadata/dialogs/StepConfigDialog.tsx
```diff
@@ -0,0 +1,357 @@
...
+      label="Custom Base Image"
+      value={baseImage || ''}
+      updateValue={(v: string) => onUpdateBaseImage(v)}
+      placeholder={resolvedDefaultBaseImage}
+      style={{ width: '100%', marginTop: '8px' }}
+    />
+    <Button
+      onClick={() => {
+        onUpdateBaseImage('');
+        onClose();
```
ada333: let's not close the whole config dialog when Image is reset to default

labextension/src/widgets/cell-metadata/dialogs/StepConfigDialog.tsx
```diff
@@ -0,0 +1,357 @@
...
+  onUpdateBaseImage: (value: string) => void;
+  // GPU / limits
+  limits: { [id: string]: string };
+  updateLimits: (actions: ILimitAction[]) => void;
+  // Caching
+  enableCaching?: boolean;
+  onUpdateCaching: (value: boolean | undefined) => void;
+}
+
+const BaseImageSection: React.FC<
```
ada333: I would move these sections in separate files, not necessity I just like having 1:1 component:file :)
So whatever you decide.

labextension/src/widgets/cell-metadata/dialogs/StepConfigDialog.tsx
```diff
@@ -0,0 +1,357 @@
...
+      open={props.open}
+      onClose={props.onClose}
+      fullWidth
+      maxWidth="sm"
+      scroll="paper"
+      aria-labelledby="step-config-dialog-title"
+      aria-describedby="step-config-dialog-description"
+    >
+      <DialogTitle id="step-config-dialog-title">Configure Step</DialogTitle>
+      <Tabs
```
ada333, line 119: Image

The tabs default text color (GPU, CACHING) in jupyter dark mode is almost invisble, could you set it to something with better visibility in both light and dark mode?

ada333 (changes requested):
hi @cordeirops, I really like this feature, the implementation is already very strong, I just left few nitpick comments.

cordeirops:
Hey @ada333, thanks for the review. Addressed all three:

1. Reset to Default no longer closes the dialog, it just resets the base image now.
2. Split BaseImageSection, GpuSection, and CacheSection into their own files under dialogs/sections/, one file per component like you suggested.
3. Fixed the Tab label contrast in dark mode, they use the same --jp-ui-font-color* variables as the rest of the dialog now.

While debugging the tab contrast I also found two bugs in the same area, fixed them too:
- Input.tsx was silently dropping any sx prop passed in whenever the field had no error, that's why GpuSection's disabled-field styling never worked.
- GpuSection's disabled-field style was hardcoded to a dark-theme-only white, changed it to a --jp-* variable and extended it to cover the label and border colors too.

Nothing outside this dialog changed, Input/Select/theme are untouched. Pushed to the same branch.

image

jesuino:
Hello @cordeirops 

My PR was merged and it led to conflicts in yours. Would you please solve the conflict?

Thanks!

cordeirops:
Hello @jesuino

I'd resolved the conflicts!

ada333 (approved):
@cordeirops looks great thanks! I did not find anything wrong in the code and tested the feature JL - confirmed that the tags get into DSL.

jesuino (approved):
LGTM - Great work!

ederign (approved):

Issue #879, opened 2026-07-13
chore: Refactor nbprocessor.py


ada333:
## Chore description
nbprocessor.py has currently around 1000 lines . It has many responsibilities. It is hard to read and it would be painful to scale with new features. I would suggest splitting into more files - and I think along the way some redundancy will be found. 

NOTE: If you want to work on this, let's first talk about strategy on how the file should be split.   

 /area backend

---

Love this idea? Give it a 👍.

harshhh817:
I'd like to work on this, and per your note here's a strategy proposal first.

I mapped the current file (983 lines) by responsibility:

| Responsibility | Members | ~lines |
|---|---|---|
| **Tag grammar + cell metadata parsing** | the `*_TAG` regexes, `get_annotation_or_label_from_tag`, `get_limit_from_tag`, `parse_cell_metadata` (140) | ~200 |
| **Dependency / dataflow analysis** | `dependencies_detection` (125), `_propagate_free_vars_from_function` (68), `_detect_fns_free_variables` (54), `_detect_in_dependencies` (23), `_ensure_fns_free_variables` | ~280 |
| **Metrics** | `METRICS_TEMPLATE`, `get_pipeline_metrics_source` (42), `assign_metrics` (58) | ~105 |
| **Reserved-cell source extraction** | `_get_reserved_tag_source` (43), `get_imports_and_functions`, `get_pipeline_parameters_source` | ~60 |
| **Orchestration** (stays) | `NotebookConfig`, `NotebookProcessor.__init__/run/to_pipeline/_post_pipeline/_configure_poddefaults/_apply_steps_defaults`, `parse_notebook` (127) | ~230 |

Proposed split under `kale/processors/`:

- `nbtags.py` — the tag language: regex constants + tag parsing helpers + `parse_cell_metadata`. This is the piece with the cleanest boundary: it maps a cell's metadata to step config and touches no notebook I/O or pipeline state, so it becomes directly unit-testable.
- `dependencies.py` — the free-variable / marshalling analysis. Pure-ish over `(pipeline graph, source)`; the biggest single win in reduced file size.
- `metrics.py` — metrics template + extraction + `assign_metrics`.
- `nbprocessor.py` — keeps `NotebookConfig`, `NotebookProcessor` and `parse_notebook` as the orchestrator that calls into the above.

That leaves the entry point around ~250 lines with the same public API (`NotebookProcessor(...).run()`), so nothing changes for callers.

[...]

PR #909, opened 2026-07-30
chore(backend): move nbprocessor constants into a separate module

harshhh817:
## What this does

Moves the constants out of `nbprocessor.py` into a new `kale/processors/constants.py`, and replaces the inline magic strings with those constants.

Part of #879 — this is the "start small" step @ada333 asked for.

## Why

`nbprocessor.py` had module-level constants at the top *and* a lot of repeated string literals inline: `"step_names"` (10x), `"tags"` (11x), `"limits"` (5x), `"prev_steps"` (4x), `"labels"` (4x), `"enable_caching"` (3x), `"base_image"` (3x), `"fns_free_variables"` (3x), `"code"` (3x), and others. A typo in any of them fails silently — a wrong dict key reads as "not set", and a wrong `getattr` name returns `None` — rather than raising.

## How it's grouped

Rather than one flat list, the new module groups by what each string actually *is*, since they aren't interchangeable:

- **Notebook JSON keys** — read from the `.ipynb` structure (`KALE_NB_METADATA_KEY`, `CELL_METADATA_TAGS`, `CELL_TYPE_CODE`).
- **Cell tag language** — the user-facing tag regexes and the two tag lists.
- **Parsed-tag dictionary keys** — `STEP_NAMES`, `PREV_STEPS`, `LIMITS`, `LABELS`, `ANNOTATIONS`, `BASE_IMAGE`, `ENABLE_CACHING`. These are the contract between `parse_cell_metadata` (which produces the dict) and `parse_notebook` (which consumes it), so both sides now reference the same names.
- **Dynamically accessed `Step` attributes** — `FNS_FREE_VARIABLES`, `PARAMETERS`, looked up via `getattr`/`hasattr`.
- **Templates and type maps** — `METRICS_TEMPLATE`, `KFP_ARTIFACT_TYPE_MAP`, plus `DEFAULT_VARIABLE_TYPE` / `DEFAULT_ARTIFACT_TYPE` for the `"str"` / `"Artifact"` fallbacks.

## Testing

- `pytest kale/tests/unit_tests` → **244 passed**
- `ruff check kale` and `ruff format --check kale` → clean
- Verified as a pure move: I compared every constant defined on `main` against the new module programmatically — all **21 are identical** in value. No behaviour change.

## Two judgment calls, happy to change either

1. **`_TAGS_LANGUAGE` → `TAGS_LANGUAGE`** (and `_STEPS_DEFAULTS_LANGUAGE` → `STEPS_DEFAULTS_LANGUAGE`). The leading underscore marked them module-private; now that they're imported across modules that's no longer accurate. Say the word if you'd rather keep the underscore.
2. **Two `.format()` calls became f-strings.** Not gratuitous: with the quoted dict keys replaced by constants, ruff's `UP032` became applicable and `ruff check` fails without it. They're the only non-mechanical lines in the diff.

Follow-ups from #879 (extracting the tag parsing, dependency analysis and metrics into their own modules) are deliberately left out of this PR.

### DIFF
--- a/kale/processors/constants.py
+++ b/kale/processors/constants.py
@@ -1,0 +1,152 @@
+# Copyright 2026 The Kubeflow Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Constants used when processing a notebook into a pipeline.
+
+Grouped by what each string actually *is*, since they are not
+interchangeable: keys read from the notebook JSON, the cell tag language,
+the keys of the parsed-tag dictionary shared between
+``parse_cell_metadata`` and ``parse_notebook``, names of
+:class:`~kale.step.Step` attributes looked up dynamically, and code
+templates / type maps.
+"""
+
+# --------------------------------------------------------------------------
+# Notebook JSON keys and the cell tag language
+# --------------------------------------------------------------------------
+
+# fixme: Change the name of this key to `kale_metadata`
+KALE_NB_METADATA_KEY = "kubeflow_notebook"
+
+SKIP_TAG = r"^skip$"
+IMPORT_TAG = r"^imports$"
+FUNCTIONS_TAG = r"^functions$"
+PREV_TAG = r"^prev:[_a-z]([_a-z0-9]*)?$"
+STEP_TAG = r"^step:([_a-z]([_a-z0-9]*)?)?$"
+# A `notebook:<name>` cell references another notebook as a sub-pipeline. The
+# referenced path is carried in the cell metadata (`notebook_path`), not the tag.
+NOTEBOOK_TAG = r"^notebook:([_a-z]([_a-z0-9]*)?)?$"
+PIPELINE_PARAMETERS_TAG = r"^pipeline-parameters$"
+PIPELINE_METRICS_TAG = r"^pipeline-metrics$"
+# Annotations map to actual pod annotations that can be set via KFP SDK
+_segment = "[a-zA-Z0-9]+([a-zA-Z0-9-_.]*[a-zA-Z0-9])?"
+K8S_ANNOTATION_KEY = f"{_segment}([/]{_segment})?"
+ANNOTATION_TAG = rf"^annotation:{K8S_ANNOTATION_KEY}:(.*)$"
+LABEL_TAG = rf"^label:{K8S_ANNOTATION_KEY}:(.*)$"
+# Limits map to K8s limits, like CPU, Mem, GPU, ...
+# E.g.: limit:nvidia.com/gpu:2
+LIMITS_TAG = r"^limit:([_a-z-\.\/]+):([_a-zA-Z0-9\.]+)$"
+# Image tag for per-step Base image selection
+# E.g.: image:python:3.11-slim
+IMAGE_TAG = r"^image:(.+)$"
+# Cache tag for per-step caching control
+# E.g.: cache:enabled or cache:disabled
+CACHE_ENABLED = "enabled"
+CACHE_DISABLED = "disabled"
+CACHE_TAG = rf"^cache:({CACHE_ENABLED}|{CACHE_DISABLED})$"
+# Report tag for per-step HTML report generation control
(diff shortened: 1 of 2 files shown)


kale/processors/constants.py
```diff
@@ -0,0 +1,144 @@
...
+    "dataset": "Dataset",  # if "dataset" in var_name.lower()-> kfp.dsl.Dataset
+    "data": "Dataset",  # if "data" in var_name.lower()-> kfp.dsl.Dataset
+    "metrics": "Metrics",  # if "metrics" in var_name.lower()-> kfp.dsl.Metrics
+    "classification": "ClassificationMetrics",
+    r"a-zA-Z0-9_": "Artifact",  # default for any other variable
+}
+
+
+# Separator between a tag's parts, e.g. `limit:nvidia.com/gpu:2`.
+TAG_SEPARATOR = ":"
```
ada333, line 103: not used anywhere

kale/processors/nbprocessor.py
```diff
@@ -647,24 +606,27 @@ def _get_reserved_tag_source(self, search_tag):
...
             # parse only source code cells
-            if c.cell_type != "code":
+            if c.cell_type != CELL_TYPE_CODE:
                 continue
             # in case the previous cell was a `search_tag` cell and this
             # cell is not any other tag of the tag language:
             if detected and (
-                ("tags" not in c.metadata or len(c.metadata["tags"]) == 0)
-                or all(not any(re.match(tag, t) for t in c.metadata["tags"]) for tag in language)
+                ("tags" not in c.metadata or len(c.metadata[CELL_METADATA_TAGS]) == 0)
```
ada333: here use the tags constant

kale/processors/constants.py
```diff
@@ -0,0 +1,144 @@
...
+
+# --------------------------------------------------------------------------
+# Dynamically accessed Step attributes
+#
+# Looked up with getattr/hasattr, so a typo silently degrades to None
+# instead of raising.
+# --------------------------------------------------------------------------
+
+FNS_FREE_VARIABLES = "fns_free_variables"
+PARAMETERS = "parameters"
```
ada333, line 144: not used anywhere

ada333 (changes requested):
hey @harshhh817 looking mostly good, I have few comments

harshhh817:
Thanks @ada333 — all three addressed in ddb8fdb.

- **`tags` at nbprocessor.py:619** — missed literal, now uses `CELL_METADATA_TAGS`. There are no `"tags"` literals left in the file.
- **`TAG_SEPARATOR` / `PARAMETERS` unused** — you're right, I defined both without applying them. Rather than delete them I wired them up, since both had real call sites: `TAG_SEPARATOR` at the four tag `split`/`join` spots, and `PARAMETERS` at its `getattr`/`hasattr` sites (matching how `FNS_FREE_VARIABLES` is used). Happy to drop them instead if you'd rather keep this PR to a pure move.

Suite still passes (268) and ruff is clean.

harshhh817:
@ada333 gentle nudge on this one whenever you have a moment — the three points from your review are addressed in ddb8fdb (the missed `tags` literal, plus `TAG_SEPARATOR` and `PARAMETERS` now applied at their call sites rather than sitting unused). CI is green and it's rebased on current main.

No rush — mainly flagging it because the follow-up extractions from #879 build on this one, so it's the piece gating the rest of that series.

harshhh817:
Rebased onto current main to clear the conflict with #867 (multi-notebook composition). `NOTEBOOK_TAG` is folded into `constants.py` and added to `TAGS_LANGUAGE` in the same position it holds upstream (right after `STEP_TAG`).

Verification: all 25 constants compare byte-identical to upstream/main (checked programmatically, including the `TAGS_LANGUAGE` ordering), 321 tests pass, `ruff check` and `ruff format --check` clean. Still zero behaviour change.

Two judgement calls I'd rather flag than assume:

1. **Scope.** #867 introduced `"notebook_names"` (4 uses) and `"notebook_path"` (3 uses) — the same class of literal this PR extracts. I converted them too, since leaving them behind would make the file half-converted. Happy to drop that and keep the diff to what existed when the PR opened — which would you prefer?

2. **`notebook_path` spans two namespaces.** It is both a cell-metadata key and a parsed-tag key, so the line reads `parsed_tags[NOTEBOOK_PATH] = metadata.get(NOTEBOOK_PATH)`. I used one constant with a comment noting it serves both, rather than two constants with identical values. Easy to split if you'd rather keep the groups strictly separate.

ada333 (approved):
lgtm!

harshhh817:
@ederign @jesuino — this one has @ada333's review and the `lgtm` label but still needs `approved` before tide will take it. Could one of you take a look when you get a moment?

Same situation on #962, if it's easier to do both in one pass. This one is the constants extraction @ada333 scoped out of #879, so the follow-up splits build on top of it.

jesuino (approved):

ederign (approved):

PR #934, opened 2026-08-12
fix(frontend): block invalid step names instead of only warning

### LINKED ISSUE #882: [bug] notebook cells (pipeline components) do not allow hyphens in the name
**Problem Description**
I need to confirm with the other Kale maintainers, but I believe that notebook cells (which compile to pipeline components) should follow DNS naming conventions and allow hyphens (`-`) in the name. Currently, the notebook cells are allowed to have underscore (`_`), but not hyphen. Furthermore, pipelines themselves are allowed to have hyphens.

**Current Behavior**
When a user enters a cell name with a hyphen (ie. `add-strings`), they get a warning that `-` is not allowed. However, they are still allowed to save the pipeline with that cell name, and assign it as a dependency in the following steps. However, the pipeline will fail to compile, with an error saying the cell name is unrecognized.

**Acceptance Criteria**
- [ ] Determine the criteria for DNS compliance of pipeline names and pipeline component names
- [ ] If the current behavior matches DNS compliance (ie. `pipeline-name` with hyphen but `pipeline_component_name` with underscore), then no further action is needed
- [ ] However, if the current behavior does not match DNS compliance, we must adjust the code to allow the proper characters, and adjust the warning if an improper character is entered.

**Example of Current Behavior**
Image
Image
Image

Bhavd33p:
## Summary
- `Input.tsx`'s `handleChange` called `updateValue()` unconditionally, regardless of whether the typed value failed the regex/maxLength/`onBeforeUpdate` validation. For the pipeline-step-name field this meant a name like `add-strings` got saved into the notebook cell metadata (and could be selected as a dependency by other steps) even though it's invalid — it would only fail later, at pipeline compile time, with a confusing error, since step names are spliced into the generated code as Python function/variable identifiers (so hyphens can't be allowed there, unlike pipeline names which follow DNS-1123 naming and do allow hyphens).
- `Input` now tracks the field's live text in local state, separate from the committed value passed via `updateValue`. Invalid text still renders in the field with the existing red error styling/message (so users can see and fix it), but it's never propagated to `updateValue`/notebook metadata. This also incidentally fixes the same class of bug for duplicate step names (the `onBeforeUpdate` check), which had the identical problem.
- Reworded `STEP_NAME_ERROR_MSG` to explain *why* hyphens aren't allowed for step names (unlike pipeline names), instead of just stating the rule.

Fixes #882

## Test plan
- [x] `tsc --noEmit`, `eslint:check`, `prettier:check` all pass.
- [x] `jest` passes (no labextension component tests exist yet in this repo to extend).
- [x] Manually verified in a running JupyterLab + Kale sidebar: typing a hyphenated step name (e.g. `add-strings`) now shows the reworded error message live and is never committed (no tag chip appears above the cell, confirmed against the notebook's actual cell metadata). Fixing it to a valid name (e.g. `add_strings`) commits normally and shows the step chip.

### DIFF
--- a/labextension/src/components/Input.tsx
+++ b/labextension/src/components/Input.tsx
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import TextField, { TextFieldProps } from '@mui/material/TextField';
 import { styled } from '@mui/material/styles';
 
@@ -70,6 +70,13 @@ export const Input: React.FunctionComponent<IInputProps> = props => {
   } = props;
 
   const [beforeUpdateError, setBeforeUpdateError] = useState(false);
+  // Tracks exactly what's in the text field, which may briefly diverge from
+  // `propsValue` while its content fails validation (see handleChange).
+  const [localValue, setLocalValue] = useState(String(propsValue));
+
+  useEffect(() => {
+    setLocalValue(String(propsValue));
+  }, [propsValue]);
 
   const getRegex = (): string | RegExp | undefined => {
     if (regex) {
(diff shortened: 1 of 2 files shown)


jesuino (changes requested):
Please notice that your change is bringing a regression, which is allow users to submit pipelines with bad name. It happens because you are validating all inputs and this validation is not triggering the behavior to prevent users from submitting bad names as the pipeline name. In main the compile, upload and run button is disabled when the pipeline name is not valid

Image

PR #935, opened 2026-08-12
chore: share constants between backend and frontend via JSON

### LINKED ISSUE #925: chore: Have shared json file for constants used both in backend and frontend code
## Chore description

I noticed that there are quite a lot of same constants defined twice (in backend and frontend code). This might be a problem when we change the constant in one place and forget to change the second definition.
It could be solved by creating a shared json file with these constants.

Image

### Labels

---

Love this idea? Give it a 👍.

Bhavd33p:
## Summary
- `kale/rpc/errors.py`'s `Code` enum and `labextension/src/lib/RPCUtils.tsx`'s `RPC_CALL_STATUS` enum independently hardcoded the same seven name→int error-code mappings. Nothing would catch it if one side changed a value (or added/removed a code) without updating the other — exactly the kind of drift risk described in the issue.
- Added `kale/rpc/error_codes.json` as the single source of truth for these values.
  - Backend: `Code`'s member values are now read from the JSON at import time (kept as an explicit `enum.Enum` class body, rather than building the enum dynamically from the dict, so `Code.SERVICE_UNAVAILABLE` etc. stay fully statically type-checkable).
  - Frontend: `RPC_CALL_STATUS`'s member values are now read from the same JSON via a relative import (`../../../kale/rpc/error_codes.json`). This only needs the file to exist in the repo at *build* time — webpack/tsc bake the JSON's contents into the compiled output, so there's no runtime dependency between the built labextension and the Python package's file layout (verified: `tsc --noEmit`, `tsc` with emit, and the full `jlpm build` webpack bundle all resolve it cleanly, and the compiled bundle contains the baked-in values).
- Added a backend test that asserts `Code`'s values match the JSON, so a future hardcoded edit that skips the JSON gets caught.

I picked this as the first (smallest, self-contained) case to tackle out of several duplicated-constant instances I found while looking into this — happy to follow up with others (e.g. the cell-tag vocabulary in `kale/processors/nbprocessor.py` vs `labextension/src/widgets/cell-metadata/constants.ts`) in separate PRs, since that one overlaps with the in-flight #909 (nbprocessor constants extraction) and deserves its own focused pass.

Fixes #925

## Test plan
- [x] `pytest kale/tests/unit_tests` — 269 passed (268 + new test).
- [x] `ruff check kale` / `ruff format --check kale` pass.
- [x] `tsc --noEmit`, `eslint:check`, `prettier:check` pass.
- [x] `jest` passes.
- [x] Full `jlpm build` (webpack) succeeds; confirmed the compiled bundle contains the JSON's values (`grep SERVICE_UNAVAILABLE` on the built `lib_index_js.*.js`).


ada333 (commented):
Hi @Bhavd33p this is great start, in the issue I meant string constants like cell tags, tags prefixes, default base image... that are both in FE and BE. Are you planning to also implement it in this PR?

Issue #937, opened 2026-08-14
[feature] Event tracking subsystem for Kale


adrielparedes:
### Feature Area

/area frontend
/area backend

### What feature would you like to see?

An event tracking subsystem for Kale that captures user actions (compile, run,
configure, etc.) and forwards them to a pluggable analytics backend. The system
uses a facade pattern with a Noop default tracker, so deployments that do not
want analytics pay zero cost and require zero configuration. Core Kale ships
only the EventTracker protocol and NoopTracker — all backend implementations
live in separate repositories, discovered at runtime via Python entry points.

### What is the use case or pain point?

Today there is no visibility into how data scientists use Kale — which features
are adopted, where users drop off, or what configuration patterns are common.
This lack of data makes it impossible to prioritize feature work based on actual
usage or to detect friction points before users report them.

PR #938, opened 2026-08-14
docs: Add Event Tracking KEP

adrielparedes:
## Summary
- Add KEP proposal for issue #937: a pluggable event tracking subsystem for Kale
- Introduces an `EventTracker` protocol with a facade pattern — Noop default, zero cost when disabled
- All backend implementations (Amplitude, Segment, etc.) live in separate repositories, discovered at runtime via Python entry points
- Covers architecture, event schema, JupyterLab UI event routing, configuration, and anonymous identity
- Documents alternatives considered (OpenTelemetry, per-entry-point tracking, in-tree backends)
### Key design decisions
- **Facade pattern**: A single `EventTracker` protocol that all entry points (JupyterLab, CLI, core engine) emit events through
- **Noop default**: No analytics unless explicitly configured — zero network calls, zero buffering, zero config
- **Entry point discovery**: Backends are separate packages registered via `kale.tracking.backends` entry points — core Kale has zero vendor dependencies
- **UI event routing**: All events route through the Python backend; pure UI events use a dedicated `POST /kale/api/track` endpoint
## Related
- Issue: #937

### DIFF
--- a/docs/proposals/0937-event-tracking/README.md
+++ b/docs/proposals/0937-event-tracking/README.md
@@ -1,0 +1,671 @@
+# KEP-0937: Event Tracking Facade for Kale
+
+
+|                     |                                                   |
+| ------------------- | ------------------------------------------------- |
+| **Authors**         | [adrielparedes](https://github.com/adrielparedes) |
+| **Created**         | 2026-08-14                                        |
+| **Updated**         | 2026-08-14                                        |
+| **Status**          | Draft                                             |
+| **Relevant Issues** | https://github.com/kubeflow/kale/issues/937       |
+
+
+## Table of Contents
+
+- [KEP-0937: Event Tracking Facade for Kale](#kep-0937-event-tracking-facade-for-kale)
+  - [Table of Contents](#table-of-contents)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [User Stories](#user-stories)
+    - [Story 1: Product Manager Analyzing Feature Adoption](#story-1-product-manager-analyzing-feature-adoption)
+    - [Story 2: Platform Team Deploying Kale Without Analytics](#story-2-platform-team-deploying-kale-without-analytics)
+    - [Story 3: Downstream Distributor Adding a Custom Backend](#story-3-downstream-distributor-adding-a-custom-backend)
+  - [Proposal](#proposal)
+    - [Architecture](#architecture)
+    - [Entry Point Discovery](#entry-point-discovery)
+    - [JupyterLab UI Event Routing](#jupyterlab-ui-event-routing)
+  - [Design Details](#design-details)
+    - [EventTracker Protocol](#eventtracker-protocol)
+    - [Noop Backend (Default)](#noop-backend-default)
+    - [Configuration](#configuration)
+    - [Anonymous Identity](#anonymous-identity)
+    - [Event Schema](#event-schema)
+    - [Proposed Events](#proposed-events)
+      - [Reserved Properties](#reserved-properties)
+    - [File Structure](#file-structure)
+    - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+    - [Risks and Mitigations](#risks-and-mitigations)
+    - [Test Plan](#test-plan)
+  - [Implementation Plan](#implementation-plan)
+  - [Migration](#migration)
+  - [Implementation History](#implementation-history)
+  - [Drawbacks](#drawbacks)
+  - [Alternatives Considered](#alternatives-considered)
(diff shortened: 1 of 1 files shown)


docs/proposals/0937-event-tracking/README.md
```diff
@@ -0,0 +1,671 @@
...
+
+
+### JupyterLab UI Event Routing
+
+Events triggered by backend operations (compile, run) are tracked inline in
+the server extension handlers that already exist — no new endpoints needed.
+The frontend simply passes the `trigger` value as part of the existing RPC
+payload.
+
+Pure UI events (`extension_opened`, `extension_action`) have no corresponding
```
jesuino, line 197: Couldn't we use the in-place RPC mechanism instead calling a new endpoint?

docs/proposals/0937-event-tracking/README.md
```diff
@@ -0,0 +1,671 @@
...
+- Notebook source code or cell contents
+- File paths beyond the working directory name
+- User names, emails, or identifiable information (unless opted in)
+
+
+
+### Proposed Events
+
+
+| Event               | Category  | When                                                      |
```
jesuino, line 414: There's also the upload event. Please also notice that upload and run events fires compile event, and run also fires upload. Will we fire all the events or a top level event should be enough? In another words if users click `run` will we also fire upload and compile or a single run event is enough?

jesuino (changes requested):
The KEP is looking good to me, thanks! I have two comments. One is a clarification and the other is a suggestion to avoid an endpoint and to use the existing RPC mechanism in place. Let me know WDYT.

docs/proposals/0937-event-tracking/README.md
```diff
@@ -0,0 +1,671 @@
...
+| ------------------- | --------- | --------------------------------------------------------- |
+| `compile_started`   | COMPILE   | User triggers compile                                     |
+| `compile_completed` | COMPILE   | Compile finishes (includes step count, duration)          |
+| `compile_failed`    | COMPILE   | Compile errors out (includes error category, not message) |
+| `run_started`       | RUN       | Pipeline submitted to KFP                                 |
+| `run_completed`     | RUN       | Pipeline run finishes (includes status)                   |
```
ada333, line 420: agree with Will, upload should be also there + run failed is missing I believe

docs/proposals/0937-event-tracking/README.md
```diff
@@ -0,0 +1,671 @@
...
+reserved with consistent semantics across all events that use them:
+
+
+| Property         | Type         | Description                                                               |
+| ---------------- | ------------ | ------------------------------------------------------------------------- |
+| `trigger`        | string       | UI surface that initiated the action (`sidebar`, `toolbar`, `cli`, `api`) |
+| `duration_ms`    | int          | Elapsed time for completed/failed operations                              |
+| `step_count`     | int          | Number of pipeline steps involved                                         |
+| `error_category` | string       | High-level error class (not the raw message)                              |
+| `fields_changed` | list[string] | Config field names that were modified                                     |
```
ada333, line 440: not necessary for v1, but for future work I would like to see Kale feature statistics (volume_count, uses_composed_notebooks, enables_caching, uses_custom_base_image ...). So it would be nice if another section for future work would be added

ada333 (commented):
@adrielparedes KEP looks good to me. I left 2 comments.

docs/proposals/0937-event-tracking/README.md
hmtosi, line 414: Going with Adam and Will's comments about the events, I think it would be better to consider it from the perspective of the button click. The user cannot separately trigger just the compile or just the run, and it would be helpful to know how they proceed through this workflow, as you mention above. I would change this table to the following, and also include events for the Kale settings.

Event | Category | When
--- | --- | ---
`extension_opened` | UI | Jupyterlab panel opened
`extension_action` | UI | Button clicked in the extension (* maybe includes the compile clicks)
`compile_and_save*` | UI | User clicks Compile and Save button
`compile_and_upload*` | UI | User clicks Compile and Upload button
`compile_and_run*` | UI | User clicks Compile and Run button
`compile_started` | COMPILE | Compile starts
`compile_completed` | COMPILE | Compile finishes (includes step count, duration, path to saved file)
`compile_failed` | COMPILE | Compile errors out (includes error category, not message)
`upload_completed` | UPLOAD | Pipeline uploaded to KFP
`upload_failed` | UPLOAD | Upload errors out (includes error category, not message)
`run_started` | RUN | Pipeline run started in KFP
`run_completed` | RUN | Pipeline run finishes (includes status)
`run_failed` | RUN | Pipeline run fails (includes error category)
`config_changed` | CONFIGURE | User modifies pipeline config (includes which fields, not values)
`settings_changed` | CONFIGURE | User modifies values in the Settings menu (includes which fields, not values)

hmtosi (changes requested):
@adrielparedes just a few comments about the specific actions, otherwise it looks good!

PR #939, opened 2026-08-15
feat: add searchable runtime image selection

CodeVishal-17:
## Summary

- Replace the Base Image text input with a searchable MUI Autocomplete
- Provide a curated list of common runtime images
- Keep freeSolo support for custom image names
- Preserve Reset to Default behavior

## Testing

- TypeScript compilation passed
- JupyterLab extension build passed
- Manually verified the searchable Base Image selector in JupyterLab
- Verified selecting `pytorch/pytorch:2.0-cuda12`


labextension/src/lib/runtimeImages.ts
```diff
@@ -0,0 +1,7 @@
+export const RUNTIME_IMAGES = [
+  'python:3.12',
```
ada333: Let's not hardcode the runtime images. Could you instead add them to `kale-settings.json` so users can configure the list? Then let's also mention it in docs for the users.

labextension/src/widgets/cell-metadata/dialogs/sections/BaseImageSection.tsx
```diff
@@ -26,25 +26,53 @@ export const BaseImageSection: React.FC<IBaseImageSectionProps> = ({
...
+    setInputValue(baseImage || '');
+  }, [baseImage]);
+
+  return (
+    <>
+      <p style={{ margin: '8px 0' }}>
+        Default: <strong>{resolvedDefaultBaseImage}</strong>
+      </p>
+
+      <Autocomplete
```
ada333, line 43: Right now this component looks different from other dropdowns in Kale UI (GPU Type, Experiment name dropdown). I wonder if we should unify them or leave this one to look different since it is the only one that is searchable.

ada333 (changes requested):
hi @CodeVishal-17 looks good - I left two comment -> to allow users configure the list of images. The second one I am not sure about, just thinking out loud

ada333 (changes requested):
@CodeVishal-17 add the documentation please
also I noticed that when I type in the image and don't press enter it is not applied to the step - this should be changed

ada333 (approved):
LGTM @CodeVishal-17 , thank you for working on this!

docs/source/user-guide/annotating-notebooks.md
```diff
@@ -95,6 +95,23 @@ Pipeline-level Kale settings live on the notebook (not on a cell) under the
...
 
+### Runtime images
+
+The `runtimeImages` setting controls the container images available in the
+**Base Image** selector when configuring a pipeline step. The list can be
+customized in **JupyterLab Settings → Kale → Runtime images**.
+
+The default list includes:
+
+- `python:3.12`
```
jesuino, line 106: I would remove some images for the sake of simplicity. I think, for example, `python:3.11` and `pytorch/pytorch:2.0-cuda12` could be removed

jesuino (changes requested):
It is looking great and working as expected! The only suggestion is to decrease the number of default images because it is polluting the setting with a lot of options. I would also move it to be the last item on the configuration - it is hard to go thought all the settings if we have a large list of images

Thanks!

jesuino (approved):
It is looking much better now, thanks!

Issue #945, opened 2026-08-24
[backend] Root pipeline parameters do not reach steps inside a referenced notebook — compiles clean, NameError at runtime


ederign:
### Environment

* How did you deploy Kale? Local source checkout (`uv sync`), compiled through the same
  `NotebookProcessor` + `Compiler` path the labextension's `nb.compile_notebook` RPC uses.
* Kale version: `2.2.1a1` (`main` @ `835916e`, plus #867)

### Steps to reproduce

A root notebook defines a pipeline parameter; a step **inside a referenced notebook** uses it.

`root.ipynb` (pipeline name `root`):

| cell tags | source |
| --- | --- |
| `pipeline-parameters` | `epochs = 7` |
| `notebook:child` (metadata `notebook_path: child.ipynb`) | *(empty)* |

`child.ipynb` (pipeline name `child`):

| cell tags | source |
| --- | --- |
| `step:train` | `print(f'child training for {epochs} epochs')` |

Compile `root.ipynb`. **It succeeds** — no error, no warning.

### Expected result

Either the parameter is threaded into the sub-pipeline, or compilation fails with a clear error
naming the unbound variable. Instead it compiles clean and the step fails at runtime.

### Actual result

The generated child module drops the parameter entirely:

```python
# .kale/kale_notebook_root_child.py
def train_step(train_html_report: Output[HTML]):        # <- no `epochs` parameter
    _kale_pipeline_parameters_block = f'''
    '''                                                  # <- empty

    _kale_block1 = '''
    print(f'child training for {epochs} epochs')         # <- `epochs` is unbound here
    '''
```

There is no data-loading marshal entry for `epochs` either, so nothing binds it at runtime. The
step raises `NameError: name 'epochs' is not defined` inside the pod, after the pipeline has been
accepted and scheduled.

The compiled IR confirms the parameter never crosses the boundary:

```
root pipeline parameters (IR)        ['epochs']
child sub-DAG input parameters (IR)  none
root call site                       kale_notebook_root_child_pipeline()
sub-pipeline signature               def kale_notebook_root_child_pipeline():
```

For contrast, the *same notebook compiled standalone* works — Kale injects pipeline parameters into
every step signature, e.g. `report_step(epochs=epochs, lr=lr)`. The behaviour only diverges once
the notebook is composed.

### Why this matters

This is the same class as the four "works standalone, silently different once composed" bugs fixed
[...]

PR #946, opened 2026-08-26
fix(backend): stop CLI defaults from overriding the notebook's pipeline name

### LINKED ISSUE #942: [backend] Generated sub-pipeline modules collide across notebooks on the `kale --nb` path
### Environment

* How did you deploy Kale? Local source checkout (`uv sync`), compiling via the `kale` CLI.
* Kale version: `2.2.1a1` (`main` @ `835916e`, plus #867)

### Steps to reproduce

Generated sub-pipeline modules are prefixed to avoid collisions, but the prefix is keyed on the
**pipeline name** rather than on the notebook. `kale --nb` defaults the pipeline name to
`kale-pipeline` for every notebook, so on the CLI path every composition in a directory generates
the same set of filenames and silently overwrites the previous one.

Using the examples shipped in `examples/composition/`:

```console
$ cp examples/composition/*.ipynb /tmp/repro/ && cd /tmp/repro

$ kale --nb main.ipynb
$ ls .kale/
kale-pipeline.kale.py
kale-pipeline.pipeline.yaml
kale_notebook_kale_pipeline_notebook_a.py
kale_notebook_kale_pipeline_notebook_b.py
kale_notebook_kale_pipeline_notebook_c.py

$ kale --nb main_mixed.ipynb
$ ls .kale/
kale-pipeline.kale.py                        # <- overwritten
kale-pipeline.pipeline.yaml                  # <- overwritten
kale_notebook_kale_pipeline_notebook_a.py    # <- overwritten
kale_notebook_kale_pipeline_notebook_b.py    # <- overwritten
kale_notebook_kale_pipeline_notebook_c.py    # <- overwritten
```

Both compositions land on the same five files. No warning is emitted, and the first pipeline's
generated code is gone.

The two notebooks do carry distinct names in their metadata (`notebook-sequence` and
`notebook-sequence-mixed`), but `kale --nb` ignores them, and the generated DSL reports
`name='kale-pipeline'` for both.

### Expected result

Compiling two different composition notebooks in the same directory should produce two distinct
sets of generated files, or fail loudly rather than silently overwriting.

### Materials and Reference

The prefix is built from the pipeline name:

https://github.com/kubeflow/kale/blob/main/kale/compiler.py#L91-L96

```python
def _module_name(root_name: str, reference_name: str) -> str:
    """Module a referenced notebook is generated into.

    The root's name is part of it because two notebooks in one directory can
    reference different notebooks under the same name, and their modules are
    written side by side.
    """
    return f"{MODULE_PREFIX}{root_name}_{reference_name}".replace("-", "_")
```

and the caller passes `pipeline_name`, not the notebook name:

```python
def _module_name(self, node):
    """Module the referenced notebook of ``node`` is generated into."""
    return _module_name(self.pipeline.config.pipeline_name, node.name)   # kale/compiler.py:447
```

while the CLI defaults that value identically for every notebook:

```python
metadata_group.add_argument(
    "--pipeline_name", type=str, default="kale-pipeline", help="Name of the deployed pipeline"
)                                                                        # kale/cli.py:92
```

Notes:

* **The labextension path is not affected.** The panel's `nb.compile_notebook` RPC passes the
  notebook's real pipeline name, so the prefix differs per notebook and the collision fix works as
  intended there. This is why the composition test suite does not catch it — every test supplies an
  explicit pipeline name.
* `kale/cli.py` predates #867; the `kale-pipeline` default is long-standing. Only the interaction
  is new: before #867 there were no generated sibling modules at all.
* Passing `--pipeline_name` explicitly is a working workaround today.

### Suggested fix

Key `_module_name` on the notebook filename instead of the pipeline name. The filename is unique
within a directory by construction, which is exactly the property the prefix needs, and it closes
the gap on both the CLI and the panel paths. The docstring's stated intent ("two notebooks in one
directory") already describes the notebook, not the pipeline.

Worth deciding separately whether `kale --nb` should keep defaulting `--pipeline_name` to a
constant, or fall back to the notebook's own metadata / filename.

Found while reviewing #867.

harshhh817:
## What this does

Stops `--pipeline_name` and `--experiment_name` from overriding the notebook's own metadata when they weren't actually passed.

Fixes #942.

## Why

Both arguments carried an argparse default, so they were always non-`None` and always survived the filter that builds the metadata-overrides dict:

```python
mt_overrides_group_dict = {
    a.dest: getattr(args, a.dest, None)
    for a in mt_overrides_group._group_actions
    if getattr(args, a.dest, None) is not None      # every other override is None when unset
}
```

Overrides are applied *on top of* the notebook's metadata, so every notebook compiled with `kale --nb` was named `kale-pipeline` no matter what it declared. They're the only two arguments in that group with defaults — the rest correctly resolve to `None`.

That's what produces the collision in #942: `_module_name()` prefixes generated sub-pipeline modules with the root pipeline name, so two composition notebooks in one directory generate identical filenames and the second run silently overwrites the first.

Worth noting the blast radius is a little wider than the issue title suggests — on the CLI path a notebook's `pipeline_name` and `experiment_name` were being ignored outright, so pipelines were also uploaded under the wrong name.

## The fix

Drop both argparse defaults and pass them to `NotebookProcessor` as kwargs instead. `NotebookConfig` resolves them *below* the notebook's metadata:

```python
self.config = NotebookConfig(**{**kwargs, **nb_metadata})
```

so precedence becomes explicit CLI argument → notebook metadata → CLI default, which is what the CLI's own help text describes ("If the same argument is provided both in the Notebook metadata and from CLI, the CLI parameter will take precedence" — for an argument the user actually provided).

## Testing

Reproduced the issue exactly with the shipped examples, then confirmed the fix. Before, both compositions land on the same five files:

```console
$ kale --nb main.ipynb && kale --nb main_mixed.ipynb
$ ls .kale/
kale-pipeline.kale.py                        # overwritten
kale_notebook_kale_pipeline_notebook_a.py    # overwritten
...
```

After, they coexist:

```console
$ ls .kale/
notebook-sequence.kale.py
notebook-sequence-mixed.kale.py
kale_notebook_notebook_sequence_notebook_a.py
kale_notebook_notebook_sequence_mixed_notebook_a.py
...
```

New `kale/tests/unit_tests/test_cli.py` covers the three precedence cases for both fields — metadata is kept when the CLI passes nothing, the default still applies when neither names one, and an explicit argument still wins — plus that two notebooks keep distinct names. Full suite: **328 passed**; `ruff check` and `ruff format --check` clean.

## Note on #945

While reproducing this I looked at #945 (root pipeline parameters not reaching steps inside a referenced notebook) — that's a separate problem in how sub-pipeline inputs are built and isn't affected by this change. Happy to take that one next if nobody else has picked it up.

### DIFF
--- a/kale/cli.py
+++ b/kale/cli.py
@@ -39,6 +39,10 @@ is provided both in the Notebook metadata and from CLI, the CLI parameter
 will take precedence.\n
 """
 
+# Used only when neither the notebook metadata nor the CLI names one.
+DEFAULT_PIPELINE_NAME = "kale-pipeline"
+DEFAULT_EXPERIMENT_NAME = "Kale-Pipeline-Experiment"
+
 METADATA_GROUP_DESC = """
 Override the arguments of the source Notebook's Kale metadata section
 """
@@ -85,12 +89,9 @@ def main():
     metadata_group.add_argument(
         "--experiment_name",
         type=str,
-        default="Kale-Pipeline-Experiment",
         help="Name of the created experiment",
     )
-    metadata_group.add_argument(
-        "--pipeline_name", type=str, default="kale-pipeline", help="Name of the deployed pipeline"
-    )
+    metadata_group.add_argument("--pipeline_name", type=str, help="Name of the deployed pipeline")
     metadata_group.add_argument(
         "--pipeline_description", type=str, help="Description of the deployed pipeline"
     )
@@ -127,7 +128,16 @@ def main():
         for a in mt_overrides_group._group_actions
         if getattr(args, a.dest, None) is not None
     }
-    processor = NotebookProcessor(args.nb, mt_overrides_group_dict)
+    # Passed as kwargs, not as metadata overrides: NotebookConfig resolves
+    # kwargs *below* the notebook's own metadata, so these act as a last-resort
+    # fallback for a notebook that names neither, while an explicit --pipeline_name
+    # still takes precedence (it lands in mt_overrides_group_dict).
+    processor = NotebookProcessor(
+        args.nb,
+        mt_overrides_group_dict,
+        pipeline_name=DEFAULT_PIPELINE_NAME,
+        experiment_name=DEFAULT_EXPERIMENT_NAME,
+    )
     pipeline = processor.run()
     imports_and_functions = processor.get_imports_and_functions()
     dsl_script_path = Compiler(pipeline, imports_and_functions).compile()
--- a/kale/tests/unit_tests/test_cli.py
+++ b/kale/tests/unit_tests/test_cli.py
@@ -1,0 +1,94 @@
+# Copyright 2026 The Kubeflow Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for how the CLI resolves notebook metadata against its own arguments."""
+
+import nbformat as nbf
+import pytest
+
+from kale.cli import DEFAULT_EXPERIMENT_NAME, DEFAULT_PIPELINE_NAME
+from kale.processors import NotebookProcessor
+
+
+def _write_nb(path, metadata):
+    """Write a single-step notebook whose Kale metadata is `metadata`."""
+    nb = nbf.v4.new_notebook()
+    nb.metadata["kubeflow_notebook"] = {"volumes": [], **metadata}
+    cell = nbf.v4.new_code_cell(source="x = 1")
+    cell.metadata["tags"] = ["step:one"]
+    nb.cells.append(cell)
+    nbf.write(nb, str(path))
+    return path
+
+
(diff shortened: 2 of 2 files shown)


kale/cli.py
```diff
@@ -127,7 +128,16 @@ def main():
...
-    processor = NotebookProcessor(args.nb, mt_overrides_group_dict)
+    # Passed as kwargs, not as metadata overrides: NotebookConfig resolves
+    # kwargs *below* the notebook's own metadata, so these act as a last-resort
+    # fallback for a notebook that names neither, while an explicit --pipeline_name
+    # still takes precedence (it lands in mt_overrides_group_dict).
+    processor = NotebookProcessor(
+        args.nb,
+        mt_overrides_group_dict,
+        pipeline_name=DEFAULT_PIPELINE_NAME,
+        experiment_name=DEFAULT_EXPERIMENT_NAME,
```
jesuino: The experiment name can't be empty and it seems that this approach won't handle cases when the experiment name is empty, leading to regression. For example, the candies-sharing example is failing because now it says that it has no experiment name. To reproduce this: after running `make dev` in your branch try to run the CLI:

`$ .venv/bin/python -m kale.cli --nb examples/base/candies_sharing.ipynb  --run_pipeline`

Leads to the following error:
```
HTTP response body: {"code":13,"message":"[ExperimentServer]: Failed to create a experiment due to conversion error: InternalServerError: Failed to convert API experiment to model experiment: Invalid input error: Experiment must have a non-empty name","details":[{"@type":"type.googleapis.com/google.rpc.Status","code":13,"message":"Internal Server Error"}]}
```

jesuino (changes requested):
Hello @harshhh817 thanks for this contribution

The fix looks good, but it seems that there's a regression, would you please see my comment?

thanks!

PR #949, opened 2026-08-29
fix(frontend): Add PVC mode in UI, UI RWO warning and documentation of volumes in Kale

ada333:
This PR is followup to mounting volumes in kale #907 
Adds docs for user
Fetches PVC mode and displays it in UI - left panel and in the dropdown:

Screenshot From 2026-09-01 12-53-48

Screenshot From 2026-09-01 12-53-55

Screenshot From 2026-09-01 12-54-01

Also adds warning before running the pipeline when some mounted PVC has RWO or RWOP mode:
Screenshot From 2026-09-01 13-00-48

### DIFF
--- a/docs/source/user-guide/volumes.md
+++ b/docs/source/user-guide/volumes.md
@@ -1,0 +1,145 @@
+# Mounting Volumes
+
+Kale pipelines can mount Kubernetes Persistent Volume Claims (PVCs) into
+pipeline steps. This is useful for:
+
+- Sharing large datasets across steps without marshalling overhead
+- Persisting outputs beyond the pipeline run
+- Accessing pre-existing data stored on cluster volumes
+
+## Adding Volumes via the JupyterLab Extension
+
+1. Open the Kale panel in the left sidebar
+2. Scroll to the **Volumes** section
+3. Click **Add Volume**
+4. Select a PVC from the dropdown (or type a name manually)
+5. Specify the **mount path** (e.g., `/data`)
+6. Optionally enable **Expose as env var** to create an environment variable
+   pointing to the mount path
+
+### Selecting from Notebook Pod
+
+If you're running JupyterLab inside a Kubeflow notebook server, click
+**Select from notebook** to see volumes already mounted on your notebook pod.
+This makes it easy to use the same volumes in your pipeline.
+
+### Access Mode Badges
+
+Each PVC in the dropdown shows its access mode:
+
+| Badge | Access Mode       | Description                                      |
+|-------|-------------------|--------------------------------------------------|
+| RWO   | ReadWriteOnce     | Can be mounted read-write by a single node       |
+| RWX   | ReadWriteMany     | Can be mounted read-write by multiple nodes      |
+| ROX   | ReadOnlyMany      | Can be mounted read-only by multiple nodes       |
+| RWOP  | ReadWriteOncePod  | Can be mounted read-write by a single pod        |
+
+## Volume Configuration in Notebook Metadata
+
+Volumes are stored in the notebook's Kale metadata under the `volumes` key:
+
+```json
+{
+  "kubeflow_notebook": {
+    "volumes": [
+      {
+        "name": "my-data-pvc",
+        "mount_point": "/data",
+        "type": "pvc",
+        "expose_as_env_var": true
+      }
+    ]
+  }
+}
+```
+
+### Fields
+
+| Field              | Required | Description                                           |
+|--------------------|----------|-------------------------------------------------------|
+| `name`             | Yes      | Name of the PVC to mount                              |
+| `mount_point`      | Yes      | Absolute path where the volume will be mounted        |
+| `type`             | Yes      | Always `"pvc"` for PVC volumes                        |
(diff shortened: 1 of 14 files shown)


docs/source/user-guide/volumes.md
```diff
@@ -0,0 +1,137 @@
...
+  to mount the volume.
+- **RWOP**: Only one step pod can mount the volume at a time, so parallel steps
+  will fail or hang waiting for the volume.
+
+### Solutions
+
+1. **Use ReadWriteMany (RWX) volumes** — These can be mounted by pods on any
+   node simultaneously. Most network file systems (NFS, CephFS, etc.) support RWX.
+
+2. **Ensure sequential execution** — If all steps depend on each other (no
```
jesuino, line 113: I am wondering if the user use the PVC from the Notebook pod and the pod is still holding the PVC, will it impact the pipeline execution? Will he have to stop the workbench to use the PVC? Hence why I had that suggestion to copy files to a PVC selected by the user, so it won't conflict with the PVC held by the workbench. Does it make sense?

labextension/src/widgets/hooks/useDeployment.ts
```diff
@@ -146,6 +147,43 @@ export function useDeployment({
       metadata.output_path = outputPath;
     }
 
+    // Check for ReadWriteOnce volumes and warn the user
+    if (metadata.volumes && metadata.volumes.length > 0) {
+      try {
+        const pvcs = await commands.listPvcs();
```
jesuino, line 153: My AI agent is telling me that this could lead to a problem when you switch notebooks - it could list PVCs for an outdated PVC? A possible solution would use the [PR from Harsh ](https://github.com/kubeflow/kale/pull/962) to get the active notebook

Does it make sense?

docs/source/user-guide/volumes.md
```diff
@@ -0,0 +1,137 @@
...
+2. Scroll to the **Volumes** section
+3. Click **Add Volume**
+4. Select a PVC from the dropdown (or type a name manually)
+5. Specify the **mount path** (e.g., `/data`)
+6. Optionally enable **Expose as env var** to create an environment variable
+   pointing to the mount path
+
+### Selecting from Notebook Pod
+
+If you're running JupyterLab inside a Kubeflow notebook server, click
```
jesuino, line 22: I am wondering if the user use the PVC from the Notebook pod and the pod is still holding the PVC, will it impact the pipeline execution? Will he have to stop the workbench to use the PVC? Hence why I had that suggestion to copy files to a PVC selected by the user, so it won't conflict with the PVC held by the workbench. That is not true if the PVC uses RWX. Does it make sense? 

NOTE: For RWO and the pipeline pods and the notebook being on the same node will also work, but I have not made a research of how a pipeline and a workbench pod relate to nodes - how do we guarantee that they are running on the same node? Hence why I believe that mentioning RWX or a note linking the [Kubernetes docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) may be enough and let the user decide

labextension/schema/kale-settings.json
```diff
@@ -22,6 +22,12 @@
       "description": "Show the 'Notebook' cell type, which references another notebook so it compiles as a nested sub-pipeline. Multi-notebook composition is still in development. Compiling a notebook that already contains a 'notebook:' cell works whether or not this is enabled.",
       "default": false
     },
+    "enableVolumes": {
+      "type": "boolean",
+      "title": "Enable volumes / PVC panel",
+      "description": "Show the Volumes section in the Kale sidebar, allowing PVCs to be mounted into pipeline steps. Disable to hide the panel when volume management is handled outside of Kale.",
```
jesuino, line 28: I would not say "when volume management is handled outside of Kale", I would just say " Disable to hide the PVC panel" because there may be the users that simply don't care about volumes and PVC

jesuino (changes requested):
Hello @ada333 

A few suggestions for changing, please let me know if it makes sense to you.

Thanks!

labextension/src/widgets/hooks/useDeployment.ts
ada333: I was not able to reproduce this. If it occurs I can create issue and fix it later

docs/source/user-guide/volumes.md
ada333: good point, I added warning (and in the docs) to the dialog when selecting PVCs from notebook pod:

Image

ada333:
@jesuino thanks for review, good points. I implemented them (except one which I could not replicate) + added volumes panel to be disabled by default.

labextension/src/widgets/hooks/useDeployment.ts
jesuino: I do believe this may happen only in very specific circumstances, for example, if the backend is slow and you switch notebooks while PVCs are still loading, the new notebook may receive PVC information from the older notebook. If we see this scenario in the future, then this may be the culprit!, for now we can leave it as it is.

jesuino (approved):
LGTM!

PR #950, opened 2026-09-01
fix(frontend): reset left panel metadata on notebook switch

### LINKED ISSUE #644: bug(frontend): Left panel shows stale metadata when switching between notebooks
### Description

When switching between notebooks, the left panel briefly shows metadata (experiment name, pipeline name) from the previous notebook before loading the new notebook's metadata.

This happens because `setNotebookPanel()` in `LeftPanel.tsx`https://github.com/kubeflow/kale/blob/36a24509df3dba60ab72ffa35a8397471122ab87/labextension/src/widgets/LeftPanel.tsx#L4 has multiple `setState` calls with `await`s between them. While waiting for async operations (RPC calls like `getBaseImage()`, `getExperiments()`), the UI still displays the old notebook's state.

You can see it in this video at the moment of switching from titanic to cifar10 the left panel still shows titanic's metadata on the cifar10 notebook:

https://www.loom.com/share/228199400bbb494c8d2b7106a46f822e

### Expected result

When switching notebooks, the left panel should immediately reset or show a loading state, not carry over metadata from the previous notebook.

---

Hariharanpugazh:
## What this does

Fixes stale metadata in the Kale left panel when switching between notebooks.

## Problem

When switching from one notebook to another, the left panel kept showing the
**previous** notebook's experiment, pipeline name, and description. The
notebook-load sequence in `useNotebookLoader.loadNotebookPanel()` awaits several
backend RPCs (`getKfpUiHost`, `getNamespace`, `getExperiments`) before it calls
`setMetadata`, and it never cleared state first. During that async window the
panel rendered stale values from the notebook we just left.

## Fix

- Reset the panel to fresh defaults **synchronously** at the start of every
  notebook switch, before any `await`, via a new `resetForNotebookSwitch()` in
  `useNotebookMetadata`. This also sets `gettingExperiments`, so the experiment
  field shows its existing "Loading..." state during the load instead of stale text.
- Guard the async load with an `isStale()` check after each await: if the user
  switches notebooks again while a load is in flight, the late-resolving load
  bails out instead of clobbering the now-active notebook's metadata.
- Clear the loading flag at the end of `loadNotebookPanel` so it resets on the
  non-backend path too.

Reset uses a fresh default metadata object rather than the shared
`DefaultState.metadata` reference, so it stays compatible with the
`DefaultState` freezing proposed in #868 (#643).

## Testing

- `jlpm eslint:check` — clean
- `prettier --check` — clean
- `tsc --noEmit` — clean
- `jest` — passing

Manually verified by opening two notebooks with different metadata and switching
between them: the panel now resets/loads instead of showing the previous
notebook's values.

## Related

- Fixes #644
- Touches the same files as #868 (#643); reset avoids mutating `DefaultState`
  so the two changes shouldn't conflict.


ada333 (commented):
hi @Hariharanpugazh I could not replicate the bug that is described in this PR - could you please share a video, or detailed instructions?

ada333 (commented):
hi @Hariharanpugazh - I still was not able to reproduce either problem. If you see it please share a video with the bug.
However when trying to reproduce this issue I discovered a new bug described here https://github.com/kubeflow/kale/issues/965 - if you are interested please feel free to work on it!

Hariharanpugazh:
Closing this PR.

Thanks @ada333 for reviewing and trying to reproduce this. After digging in
much further, I found that the stale-metadata symptom is a timing-sensitive
race that's very hard to hit on a fast KFP backend, which is likely why it
didn't reproduce on your end.

More importantly, while investigating I found that a robust fix isn't a small
patch the left panel keeps pipeline metadata in a single shared state that is
synced to notebook files through a deferred effect, and reliably attributing an
edit to the correct notebook across tab switches needs a small state-ownership
refactor rather than the approach I took here. My attempts in this PR didn't
fully hold up under testing, so rather than merge a partial fix I'd prefer to
close this and leave #644 open for a cleaner, dedicated fix (happy to revisit it
with guidance on the intended state model).

In the meantime I've opened a separate PR for #965 (pipeline metadata not shown
when KFP is disconnected), which is reproducible and verified.

Issue #948, opened 2026-08-28
[frontend] Kale cannot be enabled when KFP status is loading


ada333:
### Environment

*  How did you deploy Kale?
On main branch - `make dev` + `make jupyter` using KinD cluster

### Steps to reproduce

1. `make dev` + `make jupyter` from main branch and have disconnected kfp
2. when kfp status is loading do not open any notebook
3. when kfp status shows disconnected open notebook and see that Kale cannot be enabled:

<img width="1960" height="1390" alt="Image" src="https://github.com/user-attachments/assets/ade8d57c-1975-4c29-bc95-bd6be4c7f6c3" />

4. also  when kfp status is loading and you open notebook Kale cannot be enabled:

<img width="1960" height="1390" alt="Image" src="https://github.com/user-attachments/assets/4ba02395-786a-4444-a62a-e02a12aa8e21" />

### Expected result
It should be possible to enable Kale regarding the kfp status situation

### Workaround

- if you open a notebook when kfp status is loading Kale can be enabled
- also if you wait for about 30s Kale can be enabled

---

Impacted by this bug? Give it a 👍.

harshhh817:
Had a look at this — I can't reproduce it locally (no JupyterLab/KinD setup here), so treat this as a code-reading lead rather than a diagnosis, but the gate looks like it's this line in `LeftPanel.tsx`:

```ts
const activeNotebook = tracker.currentWidget;
```

It's read straight from the tracker during render rather than held in React state, and the disabled toggle plus the "Open a notebook to start working with Kale" message hang off `activeNotebook ? ... : ...`. Since `currentWidget` isn't reactive, opening a notebook doesn't by itself cause the panel to re-render, so it can keep showing the "no notebook" branch.

`useNotebookLoader` does connect to `tracker.currentChanged`, but its handler calls `loadNotebookPanel` / `setIsEnabled` — it doesn't set any state that `LeftPanel` reads for `activeNotebook`, so that subscription doesn't refresh this particular branch.

That would explain the ordering in the report: when KFP status is still loading and you open a notebook, the status request resolving afterwards triggers a re-render, which re-reads `tracker.currentWidget` and the panel corrects itself — which matches the stated workaround. When the status has already settled before you open the notebook, nothing forces that re-render.

[...]

PR #962, opened 2026-09-10
fix(frontend): track the active notebook in state so the Enable toggle appears immediately

### LINKED ISSUE #647: bug(frontend): Enable toggle stays disabled until some seconds
### What happened?

In some cases, after opening a notebook, the Kale **Enable toggle stays disabled** — stuck showing the "Open a notebook" empty state even though a notebook is already open. The toggle eventually appears after a long delay, or never appears at all if any backend call fails.

https://github.com/user-attachments/assets/bd817b31-b186-47b7-b7bc-3a5fad24d8d5

### Root cause

PR #596 (`cd341b7`) added a disabled Enable toggle as an empty state when no notebook is active. The component decides which toggle to show by calling `getActiveNotebook()` during render, which reads `tracker.currentWidget` — a mutable JupyterLab object that React does not track.

When `handleNotebookChanged` fires after opening a notebook, `tracker.currentWidget` is already updated, but no `setState` is called until deep inside `setNotebookPanel`, which runs a chain of sequential `await` calls:

```
await notebook.sessionContext.ready     →  wait for kernel startup
await commands.getKfpUiHost()           →  RPC to backend
await commands.getDefaultBaseImage()    →  RPC to backend
await commands.getNamespace()           →  RPC to backend
await commands.resumeStateIfExploreNotebook()  →  RPC to backend
await commands.getBaseImage()           →  RPC to backend
await commands.getExperiments()         →  RPC to backend
```

Until **all** of these complete, no `setState` happens, so React never re-renders, and the UI stays stuck on the disabled toggle.

### Why it only reproduces on clusters

Locally (no KFP backend), these RPC calls are fast — the toggle appears within 2-5 seconds. On a ** Kubeflow cluster**, each call hits a real backend over the network. Latency stacks up across the sequential awaits, so the toggle can remain stuck for 30+ seconds. If any call hangs or times out, the toggle never appears.

This is why the bug went unnoticed — local development doesn't exercise the slow path.

### Related follow-up

`setNotebookPanel` has no try/catch around its async chain. If any RPC call throws a `KernelError` (as opposed to `RPCError`), it becomes an unhandled promise rejection, and the `globalUnhandledRejection` handler shows an "An unhandled error has been thrown" toast. This is a pre-existing issue (not introduced by this fix) and should be addressed separately. See `functional-react.md` for the broader refactor proposal.

harshhh817:
## What this does

Keeps the active notebook in React state instead of reading `tracker.currentWidget` while rendering, so the Enable toggle appears as soon as a notebook is opened.

Fixes #647.

@ederign — you asked on that issue whether it still reproduces in 2.1. #948 is that check: same bug, reported against current `main`, two months later. I reproduced it on `main` @ `1e28a4f` and it is still there.

## Why

`LeftPanel` picked between the real Enable switch and the "Open a notebook" empty state from `tracker.currentWidget`, read during render:

```ts
const activeNotebook = tracker.currentWidget;   // LeftPanel.tsx:224
```

That is a plain JupyterLab object, so opening a notebook updates it without scheduling a re-render, and the panel keeps showing whichever branch it rendered last.

It did recover, but only incidentally: `loadNotebookPanel` sets unrelated state when it finishes, and that re-render happened to re-read the tracker. That sequence awaits kernel startup and then four backend RPCs in turn, so the recovery arrives late whenever those are slow.

That is also why the KFP status matters, which is the part #948 names. Every RPC is Python executed in the notebook's kernel, and a kernel handles one request at a time. `useKfpStatus` polls `kfp.ping` on that same kernel, and `ping` calls `get_kfp_healthz()` with no timeout — so when KFP is unreachable it holds the kernel while the notebook-load calls queue behind it. That is the ~30s delay in the report, and why "wait about 30 seconds" is the stated workaround.

## The change

A small `useActiveNotebook` hook subscribes to `tracker.currentChanged` and keeps the result in state; `LeftPanel` uses that instead of reading the tracker directly. It seeds from `tracker.currentWidget` on mount so a notebook opened before the panel mounted is still picked up.

Scoped to the branch decision only — `useNotebookLoader` and its await chain are untouched. The missing try/catch around that chain, which you noted at the bottom of #647, is still separate; I saw its "An unhandled error has been thrown" toast while testing and left it alone.

## Testing

Reproduced both cases from #948 with KFP disconnected — opening a notebook after the badge settles, and while it is still checking — and confirmed the toggle appears immediately in both.

Verified as an A/B on the same machine: patch stashed, extension rebuilt, bug reproduced; patch restored, rebuilt, bug gone. Same steps each time, on a reset workspace, opening `examples/composition/main.ipynb`.

|  | `.kale-no-notebook-message` | `.kale-disabled-toggle` |
| --- | --- | --- |
| before | present | present |
| after | absent | absent |

`jlpm build:lib`, `jlpm lint:check`, `jlpm test` and `pytest kale/tests/unit_tests` (329 passed) are all clean.

**Before** — `main.ipynb` is open, but the panel still shows the disabled toggle and the empty state:

before

**After** — same steps, same disconnected KFP:

after

No unit test: the change is a signal subscription with no pure logic to extract, and the extension has no React test renderer, so the pattern used in #868 and #950 does not apply here. Happy to add one if you would rather pull in `@testing-library/react`.

### DIFF
--- a/labextension/src/widgets/LeftPanel.tsx
+++ b/labextension/src/widgets/LeftPanel.tsx
@@ -32,6 +32,7 @@ import { KaleEmptyState } from './KaleEmptyState';
 import { KFPStatusBadge } from '../components/KFPStatusBadge';
 import kaleLogo from '../../style/icons/kale.svg';
 import { useKfpStatus } from './hooks/useKfpStatus';
+import { useActiveNotebook } from './hooks/useActiveNotebook';
 import { useNotebookMetadata } from './hooks/useNotebookMetadata';
 import { useDeployment } from './hooks/useDeployment';
 import { setLeftPanelCallbacks } from '../commands/kaleToolbar';
@@ -89,6 +90,8 @@ export const KubeflowKaleLeftPanel: React.FC<IProps> = props => {
   );
 
   const kfpStatus = useKfpStatus(kernel, backend);
+
+  const activeNotebook = useActiveNotebook(tracker);
 
   const notebookMeta = useNotebookMetadata({
     tracker,
@@ -220,8 +223,6 @@ export const KubeflowKaleLeftPanel: React.FC<IProps> = props => {
       label="Enable Pipeline Caching"
     />
   );
-
-  const activeNotebook = tracker.currentWidget;
 
   return (
     <ThemeProvider theme={theme}>
--- a/labextension/src/widgets/hooks/useActiveNotebook.ts
+++ b/labextension/src/widgets/hooks/useActiveNotebook.ts
@@ -1,0 +1,58 @@
+// Copyright 2026 The Kubeflow Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { useEffect, useState } from 'react';
+import { INotebookTracker, NotebookPanel } from '@jupyterlab/notebook';
+
+/**
+ * Hook that keeps the currently active notebook in React state.
+ *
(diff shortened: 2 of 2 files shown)


ada333 (approved):
@harshhh817 thanks for working on this LGTM! I tested this and I can't see the bug reported in #948. Code looks good as well.

ada333 (approved):

jesuino (approved):

PR #966, opened 2026-09-16
fix(frontend): show notebook metadata when KFP is disconnected

### LINKED ISSUE #965: [frontend] Notebook metadata are not shown when KFP is loading or disconnected
### Environment

make dev + make jupyter from main branch

### Steps to reproduce
when KFP is loading or disconnected Pipeline metadata are not visible even though they are part of the notebook file and do not relate to KFP:

Image

### Expected result

When defined in the notebook file, the experiment name, pipeline name and pipeline description should be visible even when kfp is disconnected.

### Materials and Reference

---

Hariharanpugazh:
Screenshot 2026-09-16 112835

## What this does

Fixes #965 — the Kale left panel now shows a notebook's saved pipeline
metadata (experiment name, pipeline name, pipeline description) even when
KFP is disconnected or still loading.

## Problem

`loadNotebookPanel` fetched experiments (a KFP call) *before* the code that
applies the notebook's saved metadata to the panel, and the function had no
error handling. When KFP was unreachable, `getExperiments` rejected and the
whole load aborted — so the saved metadata, which lives in the notebook file
and is unrelated to KFP, was never rendered and the fields showed empty.

## Fix

Wrap the experiments fetch in `try/catch/finally`:
- a KFP failure degrades to an empty experiment list instead of aborting the load,
- execution always reaches the code that displays the notebook's saved metadata,
- the experiment "loading" flag is always cleared.

## Testing

- `jlpm eslint:check`, `prettier`, `tsc --noEmit` — clean
- `jest` — passing
- Manually verified with KFP disconnected: opened `examples/base/candies_sharing.ipynb`
  and the panel shows Pipeline Name `candies-sharing` and its description, where
  before the fix these were blank.


ada333 (changes requested):
hi @Hariharanpugazh thanks for sending a PR!
it does not quite solve the problem described in #965 - the functionality I was looking for was that notebook metadata in the left panel will be filled out, regardless the KFP state (loading, disconnected, connected). I think what needs to be done (in the same code section you changed) is to change the order of the calls - because now there is an await for kfp getExperiments first and that is what is blocking the metadata to be filled out.

Hariharanpugazh:
Thanks @ada333, that's exactly right - the `await getExperiments` was still
blocking the metadata from filling in while KFP loads.

I've reordered the calls as you suggested: the notebook's saved metadata
(experiment, pipeline name, description) is now applied to the panel *before*
the KFP `getExperiments` call, so the fields are populated immediately
regardless of KFP state — loading, disconnected, or connected. The experiments
fetch afterward only enriches the experiment dropdown when KFP is reachable.

Verified locally with KFP disconnected: opening `examples/base/candies_sharing.ipynb`
now shows Pipeline Name `candies-sharing` and its description right away, even
while the experiment list is still loading.


image

labextension/src/widgets/hooks/useNotebookLoader.ts
```diff
@@ -109,7 +197,13 @@ export function useNotebookLoader({
 
       const notebookMetadata = NotebookUtils.getMetaData(notebook, metadataKey);
 
-      let fetchedExperiments: IExperiment[] = [];
+      // Apply the notebook's saved metadata (pipeline name, description,
+      // experiment) to the panel FIRST, before any KFP call. This metadata is
```
ada333, line 201: not true because the `getKfpUiHost()` is called before that  (kfp call) causing delay for the notebook metadata. please move it right after ` await notebook.sessionContext.ready;`

ada333 (changes requested):
hi @Hariharanpugazh, thanks for the changes, moving in the right direction! But it is still not working quite like expected - I left 1 comment in the code and 1 suggestion here in this review-comment:

There is also a problem that the experiment name is not filled -> this is because here 
https://github.com/kubeflow/kale/blob/main/labextension/src/widgets/LeftPanel.tsx#L150
is checked if `notebookMeta.experiments.length > 0` and only if it is then the Experiment name in LeftPanel UI is populated - but the problem is that notebookMeta.experiments is only the fetched experiments from KFP which is empty if KFP is disconnected

Hariharanpugazh:
Thanks @ada333, both good catches addressed in the latest commit:

1. **Moved `applySavedMetadata` before `getKfpUiHost`.** You're right that
   `getKfpUiHost()` is a KFP call and running it first delayed the metadata. The
   notebook metadata is now applied immediately after `sessionContext.ready`,
   before any KFP call.

2. **Experiment name now shows when KFP is disconnected.** The experiment input
   in `LeftPanel.tsx` was only populated when `notebookMeta.experiments` (the
   KFP-fetched list) was non-empty. I added a fallback so that when no experiment
   list is available, the field is filled from the notebook's saved `experiment`
   / `experiment_name`.

Verified locally with KFP **disconnected** using `iris.ipynb` experiment name,
pipeline name, and description all fill in immediately:

Screenshot 2026-09-24 103225

ada333 (approved):
@Hariharanpugazh , tested this and looks very good, thank you for working on this!

PR #969, opened 2026-09-19
fix(backend): fix NamedTuple mutation crash in Step.add_artifact

### LINKED ISSUE #968: [backend] AttributeError when Step.add_artifact updates type of existing NamedTuple
### Environment
* **How did you deploy Kale?** Local Windows development environment (`make dev`)
* **Kale version:** `main` branch

### Steps to reproduce
Run the following Python snippet:

```python
from kale.step import Step, Artifact

step = Step(source=['pass'], name='my_step')
step.artifacts.append(Artifact(name='my_model', type=None, is_input=False))
step.add_artifact('my_model', 'Model', is_input=False) # Crashes here
```

### Expected result
`Step.add_artifact()` should successfully update the artifact's type field without raising an exception. 

### Materials and Reference
Python raises an `AttributeError` because `Artifact` is a `NamedTuple` (which is immutable). 

In `kale/step.py` line 109, the code attempts an in-place property assignment:

```python
existing_art.type = artifact_type  # AttributeError: can't set attribute
```

**Proposed Fix:** Replace the direct assignment with `self.artifacts[i] = existing_art._replace(type=artifact_type)`. 

I have fully reproduced this locally and would like to be assigned to this issue so I can submit a Pull Request with the fix and unit tests!

---

JaiswalShivang:
## What this does

Fixes an `AttributeError: can't set attribute` crash when updating the `type` of an existing output artifact in `Step.add_artifact`, and adds unit tests.

## Why

`Artifact` is defined as a Python `NamedTuple` in `kale/step.py` and is therefore immutable. `Step.add_artifact` attempted an in-place mutation (`existing_art.type = artifact_type`), which fails at runtime when resolving untyped output artifacts.

## The Change

- Use `existing_art._replace(type=artifact_type)` to replace the item in `self.artifacts` instead of attempting in-place assignment.
- Add unit tests in `kale/tests/unit_tests/test_step.py` covering adding artifacts, updating untyped output artifacts, and preserving existing typed artifacts.

## Testing

- Unit tests pass: `uv run pytest kale/tests/unit_tests/test_step.py -vv` (10/10 passed).
- Linting & formatting clean: `uv run ruff check kale && uv run ruff format --check kale`.


JaiswalShivang:
@ederign @StefanoFioravanzo

PR #970, opened 2026-09-19
feat: improve local KFP development workflow

### LINKED ISSUE #924: [feature] Make testing backend changes easier
### Feature Area

/area backend

I would like to have one script / make command that would spin up the wheel, set up correctly the env var, port forward and run some notebook in kfp (with the current Kale backend changes) -
f.e. `make kfp-dev candies.ipynb`

Currently you have to run 3 separate terminals to test out the backend changes:

1. `port forward`
2. `make kfp-serve`
3. `make kfp-run`

+ also in Linux you have to set the env var `KFP_HOST_ADDR`
this whole workflow takes too much time

---

Love this idea? Give it a 👍.

CodeVishal-17:
## Summary

- Add a one-command local KFP development workflow via `make kfp-dev`
- Safely manage the local KFP UI port-forward using a PID file
- Avoid PodDefaults detection outside Kubernetes pods
- Preserve PodDefaults detection inside Kubernetes
- Improve Makefile path handling by quoting the wheel directory

## Related Issue

Closes #924 

## Validation

- `uv run ruff check kale`
- `uv run ruff format --check kale`
- `uv run pytest kale/tests/unit_tests -q`
- `make kfp-compile NB=examples/base/candies_sharing.ipynb`
- `make kfp-dev NB=examples/base/candies_sharing.ipynb`
- Verified successful KFP pipeline execution
- `git diff --check`


kale/common/podutils.py
```diff
@@ -21,6 +21,11 @@
 NAMESPACE_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
 
 
+def is_running_in_pod():
```
ada333: why was this change introduced in this PR? 
if it is a separete bug, I would make new PR for this

CodeVishal-17: This change was introduced as part of the local KFP development workflow in this PR.

When running make kfp-dev locally, NotebookProcessor is executed outside a Kubernetes pod, but _configure_poddefaults() previously attempted to detect PodDefaults unconditionally. This resulted in the Kubernetes service-account namespace warning during the local workflow.

The is_running_in_pod() check makes the PodDefaults detection environment-aware: it skips the lookup when running locally, while preserving the existing behavior when running inside a Kubernetes pod.

So this isn't intended as a separate bug fix; it's to make the workflow introduced by #924 work cleanly in a local environment.

If you prefer to keep this change separate from #924, I'm happy to split it into another PR.

